PR #1, opened 2020-04-30
Performs original request when no response returned from the middleware

kettanaito:
## Changes

Whenever an outgoing request has no response returned from the middleware function, it is performed as-is, providing a transparent request behavior.

### http/https

Performs actual `http.request`/`http.get` request by calling the stored unpatched instance of that functions.

Performing an actual `https.request`/`https.get` request first restores the `http.ClientRequest` patch, and then calls the original unpatched function.


kettanaito:
## Background

Both `http` and `https` modules use `ClientRequest` to perform requests. The difference is in the `RequestOptions` they receive, which they use for constructing an `Agent` instance. This means that you can theoretically construct an Agent instance incompatible with the current request protocol and attempt to execute it, which would result into protocol violation.

## Protocol "http:" not supported. Expected "https:"

- Origin: HTTPS request issued by `XMLHttpRequest`
- Callstack:

```
XHR -> jsdom/xhr -> request (3rd) -> decides between http/https based on protocol -> [module].request
```

> It's the `request` library that augments the `RequestOptions` with the `uri` property, otherwise unknown based on NodeJS typings.

Occurs when trying to perform an HTTP request using `https` module.

However, the error itself only means that the `RequestOptions.agent.protocol` does not equal to the `defaultAgent.protocol` in [`http-client` source code](https://github.com/nodejs/node/blob/ab41b9808edadc034a4da60ab0f71e527d288892/lib/_http_client.js#L142-L156). This check wasn't met, because `options.protocol` was not set when a request is issued from `XMLHttpRequest` (I believe `fetch` as well). I assume that the `defaultAgent.protocol` is `http:`, which results into the invalid protocol exception.

### Solution

During the normalization of `ClientRequest` arguments in the `ClientRequestOverride` class I ensure that `options.protocol` is inherited from the `url`, if no protocol is set explicitly.

## Protocol "https:" not supported. Expected "http:"

- Origin: HTTPS request issued by `fetch()`

Although a mirror error to the previous one, the root cause of this is different. 

```
fetch -> node-fetch -> decides between http/https based on protocol --> [module].request()
```

PR #3, opened 2020-05-01
Adds more testing scenarios

kettanaito:
To cover the functionality of this library I'd like to add more test scenarios that target different usage of the classes it patches.

## Roadmap

- [x] http.get
- [x] http.request
- [x] https.get
- [x] https.request
- [x] XMLHttpRequest
- [x] fetch
- [x] Assert request headers in interception tests
- [x] Assert response headers in mocked response tests


kettanaito:
This setup is not perfect, as it only tests the "full" scenarios: requests with query params, body. It would be nice to have one "plain" request, and one "full" request scenarios.

PR #33, opened 2020-07-12
ClientRequest: Supports chunk/encoding in the "end" method

### LINKED ISSUE #32: ClientRequestOverride.end() does not accept chunk/encoding
> Discovered in https://github.com/mswjs/msw/issues/275

## Current behavior

`.end()` method does not accept arguments like `chunk`/`encoding` according to the [NodeJS specification](https://nodejs.org/api/http.html#http_request_end_data_encoding_callback).

Because of this, request issuing libraries that post a request body via `req.end(body)` are failing when it comes to performing an original request (request body is never written, as `req.write()` is never called).

## Expected behavior

`.end()` behaves according to the aforementioned specification, meaning:

1. It accepts an optional `chunk` argument. Appends that chunk to the existing request body.
1. It accepts an optional `encoding` argument. Propagates that encoding to the original `req.end()` call.

kettanaito:
## Changes

- Adds normalization for `ClientRequest.end` arguments.
- Appends a body chunk given to `end()` method to the existing request body.
- Propagates given body chunk to `end()` method to the original call of `req.end()` (when performing an original request).
- Merges `options.headers` and `kOutHeaders` when resolving `requestHeaders` present in the intercepted request object.
- Propagates `kOutHeaders` on the original request, if such headers were set.

## GitHub

- Fixes #32
- Fixes https://github.com/mswjs/msw/issues/275

## Roadmap

- [x] Ensure the `encoding` is propagated
- [x] Add an integration test covering the issue
- [x] Ensure the request body written through `req.write()` is also written on the original request:

https://github.com/mswjs/node-request-interceptor/blob/94969084d99969fe8bfb7b684c9b9d298c007d4b/src/http/ClientRequest/ClientRequestOverride.ts#L243-L246

> Per spec, it could be enough to just pass the entire request body as the first argument to the `req.end()` call of the original request.

- [x] Property type annotate arguments in test for `normalizeHttpRequestEndParams` function (no `any`)
- [x] Capture all headers (both initial `options.headers` and headers added via `this.setHeader()`)


kettanaito:
## Headers set via `ClientRequest.setHeader()` are not captured

This is because the `this.setHeader()` method is not patched. At the moment only the `options.headers` are captured, so any request headers set _after_ the request has been created are ignored. This is what `supertest` does when `.send({ json: true })` (sending a JSON request body).

PR #37, opened 2020-07-20
Do not load the http overwrite module in react native

Federkun:
This PR will import different modules when using this package on react-native. Specifically, we don't load `overrideHttpModule`, because the `http` core module is not available on RN.

### DIFF
--- a/src/RequestInterceptor.ts
+++ b/src/RequestInterceptor.ts
@@ -1,6 +1,5 @@
 import { RequestMiddleware, ModuleOverride } from './glossary'
-import { overrideHttpModule } from './http/override'
-import { overrideXhrModule } from './XMLHttpRequest/override'
+import { overrideModules } from './overrideModules'
 
 const debug = require('debug')('RequestInterceptor')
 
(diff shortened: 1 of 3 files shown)


kettanaito:
Hey, @Federkun, thanks for making this pull request.

It looks great. One thing I wonder: is the `*.native.js` module resolution specific to a certain React Native bundler (i.e. Metro)? Or is it a general convention in React Native?

Federkun:
It's a convention in react native itself, https://reactnative.dev/docs/platform-specific-code#platform-specific-extensions
If you need an example from another repository: https://github.com/lelandrichardson/react-primitives/tree/master/src

kettanaito:
Thanks for clarifying that!

Does the `src/overrideModules.native.ts` compiles into the build directory? Statically, it doesn't appear we are importing it in a regular JavaScript way.

Also, does this `*.native.js` resolution apply for third/fourth-party imports? `node-request-interceptor` is being imported within `msw` and not directly by your application's code. Would this still resolve the imports as expected?

```
import { setupServer } from 'msw/node'
  - import { overrideModules } from './overrideModules'
    - => "./overrideModules.native.js"
```

Federkun:
No problem.

> Does the src/overrideModules.native.ts compiles into the build directory?
It does, yes.

> Also, does this *.native.js resolution apply for third/fourth-party imports?
I can confirm that it does, yes. I had to double check, but this works for me locally

PR #42, opened 2020-08-10
Fixes default forwarded XHR.open args

Tommos0:
- `user` and `password` have a default value of `undefined` (previously `''`).
- `async` has a default value of `undefined` (`true` when forwarding) - (previously `false`).
- Removed parameter initializations; set default parameters on `open()`.
- In `getUrlByRequestOptions()`, check if `options.uri.protocol` is set before falling back to `DEFAULT_PROTOCOL`.
- minor cosmetic re-arrangements in ClientRequest interception (imports of http and https modules).

### DIFF
--- a/src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
+++ b/src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
@@ -51,10 +51,10 @@ export const createXMLHttpRequestOverride = (
     public withCredentials: boolean
     public status: number
     public statusText: string
-    public user: string
-    public password: string
+    public user?: string
+    public password?: string
     public data: string
-    public async: boolean
+    public async?: boolean
     public response: string
     public responseText: string
     public responseType: XMLHttpRequestResponseType
@@ -105,10 +105,7 @@ export const createXMLHttpRequestOverride = (
       this.withCredentials = false
       this.status = 200
       this.statusText = 'OK'
-      this.user = ''
-      this.password = ''
       this.data = ''
-      this.async = true
       this.response = ''
       this.responseType = 'text'
       this.responseText = ''
@@ -176,7 +173,7 @@ export const createXMLHttpRequestOverride = (
     public async open(
       method: string,
       url: string,
-      async?: boolean,
+      async: boolean = true,
       user?: string,
       password?: string
     ) {
@@ -192,9 +189,9 @@ export const createXMLHttpRequestOverride = (
       } else {
         this.url = url
         this.method = method
-        this.async = !!async
-        this.user = user || ''
-        this.password = password || ''
+        this.async = async
+        this.user = user
+        this.password = password
       }
     }
 
@@ -291,7 +288,7 @@ export const createXMLHttpRequestOverride = (
             originalRequest.open(
               this.method,
               this.url,
-              this.async,
+              this.async ?? true,
               this.user,
               this.password
             )
--- a/src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
+++ b/src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
@@ -248,7 +248,7 @@ export const createXMLHttpRequestOverride = (
 
           // Return a mocked response, if provided in the middleware.
           if (mockedResponse) {
-            debug('recieved mocked response', mockedResponse)
+            debug('received mocked response', mockedResponse)
 
             this.status = mockedResponse.status || 200
             this.statusText = mockedResponse.statusText || 'OK'
--- a/src/http/override.ts
+++ b/src/http/override.ts
@@ -49,13 +49,10 @@ function handleRequest(
  */
 export const overrideHttpModule: ModuleOverride = (middleware) => {
   let patchedModules: PatchedModules = {}
-  const modules: ['http', 'https'] = ['http', 'https']
+  const modules = ['http', 'https']
 
   modules.forEach((protocol) => {
-    const module = ({
-      http: require('http'),
-      https: require('https'),
-    } as Record<string, typeof http | typeof https>)[protocol]
+    const module = protocol === 'https' ? https : http
 
     const { request: originalRequest, get: originalGet } = module
 
--- a/src/utils/getUrlByRequestOptions.ts
+++ b/src/utils/getUrlByRequestOptions.ts
@@ -19,8 +19,8 @@ export function getUrlByRequestOptions(
   if (!options.protocol) {
     debug('given no protocol, resolving...')
 
-    // Assume HTTPS if using an SSL certificate.
-    options.protocol = options.cert ? 'https:' : DEFAULT_PROTOCOL
+    // Assume HTTPS if cert is set.
+    options.protocol = options.cert ? 'https:' : (options.uri?.protocol || DEFAULT_PROTOCOL)
 
     debug('resolved protocol to:', options.protocol)
   }
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 node_modules
 lib
 package-lock.json
+.idea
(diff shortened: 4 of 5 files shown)


Tommos0:
https://github.com/mswjs/msw/issues/203

src/http/override.ts
```diff
@@ -98,7 +97,7 @@ export const overrideHttpModule: ModuleOverride = (middleware) => {
     }
 
     patchedModules[protocol] = {
-      module,
+      module: module,
```
kettanaito, line 100: Why do you unfold this property assignment?

```js
{ key }
{ key: key }
```

Are identical. I think we should prefer the shorthand method.

src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
```diff
@@ -291,7 +291,7 @@ export const createXMLHttpRequestOverride = (
             originalRequest.open(
               this.method,
               this.url,
-              this.async,
+              typeof this.async === 'undefined' ? true : this.async,
```
kettanaito, line 294: Consider using null coalescing operator instead:

```js
this.async ?? true
```

This will respect `true` and `false`, and fallback to the `true` only if the value of `this.async` is `undefined`/`null`.

src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
```diff
@@ -192,9 +192,9 @@ export const createXMLHttpRequestOverride = (
       } else {
         this.url = url
         this.method = method
-        this.async = !!async
-        this.user = user || ''
-        this.password = password || ''
+        this.async = undefined
```
kettanaito, line 195: We need to be careful for the `.open()` method to have `async` as `true` by default:

https://github.com/mswjs/node-request-interceptor/blob/17453525dd439b879c428d188c05cdb37f4a1ec4/src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts#L176-L182

Source: [XMLHttpRequest/open (MDN)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open).

kettanaito:
It looks like the user you use for the commits in this pull request is not your GitHub user. I'd suggest you to double check the email attached to commits and rebase them with the updated email. This will help to associate the work you do with your GitHub account.

src/http/override.ts
Tommos0: Agreed, I ran into some weird issues while debugging at first, figured `module` might be a reserved keyword. It turns out it isn't but I forgot to change this one back.

src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
Tommos0: Ah, yes that's a nice operator :)

src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
Tommos0: Updated!

Tommos0:
> It looks like the user you use for the commits in this pull request is not your GitHub user. I'd suggest you to double check the email attached to commits and rebase them with the updated email. This will help to associate the work you do with your GitHub account.

Ah yes, thanks! Updated!

src/utils/getUrlByRequestOptions.ts
```diff
@@ -19,8 +19,8 @@ export function getUrlByRequestOptions(
   if (!options.protocol) {
     debug('given no protocol, resolving...')
 
-    // Assume HTTPS if using an SSL certificate.
-    options.protocol = options.cert ? 'https:' : DEFAULT_PROTOCOL
+    // Assume HTTPS if uri is https or if using an SSL certificate.
+    options.protocol = options.uri?.protocol || (options.cert ? 'https:' : DEFAULT_PROTOCOL)
```
kettanaito, line 23: I'm afraid this may be incorrect. If a request has been issued with a certificated attached, we must treat it as an HTTPS request, regardless of what protocol the URL may have. `options.cert` should take the higher priority. What do you think about reversing the addition:

```js
options.protocol = options.cert ? 'https:' : options.uri?.protocol || DEFAULT_PROTOCOL
```

src/http/override.ts
```diff
@@ -49,13 +49,12 @@ function handleRequest(
...
+  const modules = ['http', 'https']
 
   modules.forEach((protocol) => {
-    const module = ({
-      http: require('http'),
-      https: require('https'),
-    } as Record<string, typeof http | typeof https>)[protocol]
+    const module = protocol === 'https'
+      ? require('https') as typeof https
+      : require('http') as typeof http;
```
kettanaito, line 57: What is the practical benefit of these changes? Improved type coverage of the imported modules?

src/utils/getUrlByRequestOptions.ts
Tommos0: That's probably better. Although, this is the case where you try to do a `https` request to an URL starting with `http://`? I think Node already throws an error in that case.

src/http/override.ts
Tommos0: Just cosmetic. Actually, looking at it again i guess it could be just `const module = protocol === 'https' ? https: http`, since they are already imported in the top scope.

kettanaito:
I've rebased your feature branch against the latest `master` (includes #43). Let's see the CI status. I'll go through the changes manually and make sure nothing was missed.

kettanaito:
@Tommos0 would it be possible to add a test suite for these changes? What made the usage fail previously? Can we construct a test that would confirm we have fixed an issue during CI?

src/utils/getUrlByRequestOptions.ts
kettanaito: There are request issuing libraries that perform an HTTPS request without explicitly providing the `protocol: 'https:'` option. Instead, they have the `cert` property set. This condition is designed for that kind of libraries.

Overall, there are so many ways to perform an HTTP request in node using the `http` module it's mesmerizing. Some of them won't even compile against the type definition of the functions they call, yet they are present.

src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
```diff
@@ -105,10 +105,10 @@ export const createXMLHttpRequestOverride = (
       this.withCredentials = false
       this.status = 200
       this.statusText = 'OK'
-      this.user = ''
-      this.password = ''
+      this.user = undefined
+      this.password = undefined
       this.data = ''
-      this.async = true
+      this.async = undefined
```
kettanaito, line 111: Could you please leave an explanatory comment why the default value of `this.async` is `undefined`, while when it's used it's fallbacked to `true`? I remember you've found a reason behind this, but it needs to be noted in the code to prevent any confusion. Thanks.

Tommos0:
> @Tommos0 would it be possible to add a test suite for these changes? What made the usage fail previously? Can we construct a test that would confirm we have fixed an issue during CI?

I think the .native.js overrides makes it a bit difficult right now (it's not super straight forward to pick the native files right now). This is my similar here https://github.com/mswjs/msw/issues/203#issuecomment-674029661 .

It's not so difficult to do that, I guess - I could add it here or in another PR? Then that would make testing easier as well (instead of having to mimic a babel config that picks .native.js over .js or something).

kettanaito:
I understand. I've browsed through the changes, and it appears the changes are:

- `user` and `password` have a default value of `undefined` (previously `''`).
- `async` has a default value of `undefined` (previously `true`)
- `async` value is determined when it's being called (still unsure what's the difference as opposed to having a default value set).
- minor cosmetic re-arrangements in ClientRequest interception (imports of `http` and `https` modules).

Is this the correct changes representation? Please, could you add the list of changes to the pull request description above?

src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
Tommos0: There is no real reason - actually since it's type as optional, I suggest to remove the initialization all together. It only makes sense to have a value after open() I would say.

Tommos0:
I've removed the initialization of `user`, `password`, `async` all together, seems more readable to have a single source of defaults on the `open` method.

PR #43, opened 2020-08-11
fix: pass timeout to request

### LINKED ISSUE #7: Requests to relative URL fail
## Steps to reproduce

1. Perform a `fetch('/login')` request in Node-like environment.
1. See the following error from the `XMLHttpRequest` class:

```
    TypeError: Invalid URL: /login

      2152 |             this.readyState = this.LOADING;
      2153 |             this.data = data || '';
    > 2154 |             const url = new URL(this.url);
           |                         ^
      2155 |             const req = {
      2156 |                 url: cleanUrl(url),
      2157 |                 method: this.method,
```

marcosvega91:
## Changes

- Propagates callbacks (`ontimeout`, `onreadystatechange`, etc.) and event listeners (`timeout`, `error`, etc.) from the captured XHR instance to the original XHR instance.
- Propagates the `timeout` property to the original XHR, so that the request timeout is respected (i.e. when using `axios` with timeout).
- Adds respective tests.

## GitHub

- Originated from https://github.com/mswjs/msw/issues/284
- Fixes #7

Timeout was not passed to original request when no mock response is found.
Hope to did the right thing 😄 

Timeout should be added only if the request is async so when `this.async === true`

### DIFF
--- a/src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
+++ b/src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
@@ -320,11 +320,10 @@ export const createXMLHttpRequestOverride = (
               this.trigger('load')
             }
 
-            // Map callbacks given to the patched instance to the original request instance.
-            originalRequest.onabort = this.abort
-            originalRequest.onerror = this.onerror
-            originalRequest.ontimeout = this.ontimeout
-            originalRequest.onreadystatechange = this.onreadystatechange
+            // Assign callbacks and event listeners from the intercepted XHR instance
+            // to the original XHR instance.
+            this.propagateCallbacks(originalRequest)
+            this.propagateListeners(originalRequest)
 
             if (this.async) {
               originalRequest.timeout = this.timeout
@@ -407,5 +406,25 @@ export const createXMLHttpRequestOverride = (
     }
 
     public overrideMimeType() {}
+
+    /**
+     * Propagates captured XHR instance callbacks to the given XHR instance.
+     * @note that `onload` listener is explicitly omitted.
+     */
+    propagateCallbacks(req: XMLHttpRequest) {
+      req.onabort = this.abort
+      req.onerror = this.onerror
+      req.ontimeout = this.ontimeout
+      req.onloadstart = this.onloadstart
+      req.onloadend = this.onloadend
+      req.onprogress = this.onprogress
+      req.onreadystatechange = this.onreadystatechange
+    }
+
+    propagateListeners(req: XMLHttpRequest) {
+      this._events.forEach(({ name, listener }) => {
+        req.addEventListener(name, listener)
+      })
+    }
   }
 }
--- a/src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
+++ b/src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
@@ -326,6 +326,10 @@ export const createXMLHttpRequestOverride = (
             originalRequest.ontimeout = this.ontimeout
             originalRequest.onreadystatechange = this.onreadystatechange
 
+            if (this.async) {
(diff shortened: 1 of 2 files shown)


kettanaito:
Could you please resolve the failing tests on CI?

marcosvega91:
Yes I'm looking into it

marcosvega91:
It should be good now 🥳

kettanaito:
Looks good! The test remains. Would you be interested in writing a test case for this `timeout` change?
Basically, that NRI now respects timeout on XMLHttpRequest.

marcosvega91:
I have added the test, I think then that there is also an issue elsewhere not regarding this PR. If you set async for all requests in `xhr` helper, the error ` The "superCtor" argument must be of type function. Received undefinedTypeError [ERR_INVALID_ARG_TYPE]: The "superCtor" argument must be of type function. Received undefined` will thrown. Is the error that kent send on discord some day ago.

marcosvega91:
If is ok for I think that this can be merged and when I'll back I will try to figure out the other problem. What do you think ?

test/helpers.ts
```diff
@@ -183,7 +186,11 @@ export async function xhr(
         )
       })
     }
-
+    req.timeout = options?.timeout ?? 0
+    req.ontimeout = function (event) {
```
kettanaito, line 190: I think we shouldn't extend this utility with the timeout functionality, as it introduces too much complexity into this utility function. 

Instead, we should perform a raw `XMLHttpRequest` and assert the proper timeout handling with both callbacks and events attached to XHR instance. I'll push that test suite right now.

test/intercept/XMLHttpRequest/XMLHttpRequest.test.ts
```diff
@@ -218,3 +218,19 @@ test('intercepts an HTTP PATCH request', async () => {
...
+  try {
+    await xhr('GET', 'http://httpbin.org/get?userId=123', {
+      headers: {
+        'x-custom-header': 'yes',
+      },
+      timeout: 1,
+      onTimeout,
+      async: true,
+    })
+  } catch (error) {
```
kettanaito, line 233: By extending our `xhr` util, we also make our test behave not like an actual XHR instance would behave. XHR won't throw on timeout, and neither should our test.

src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
```diff
@@ -325,7 +324,9 @@ export const createXMLHttpRequestOverride = (
             originalRequest.onerror = this.onerror
             originalRequest.ontimeout = this.ontimeout
             originalRequest.onreadystatechange = this.onreadystatechange
-
+            if (this.async) {
+              originalRequest.timeout = this.timeout
```
kettanaito, line 328: Should we always inherit `.timeout` callback? I feel it's the end user's responsibility to align that they assign `timeout` callback only for `async` XHR.

marcosvega91: If we assign a timeout function when a request is not async an error is thrown. We can leave this error for the user

kettanaito:
Please, I will have to force push into your feature branch. Do not push anything new until that's done. Thank you!

marcosvega91:
I can't push until next week so is not a problem. Thank you @kettanaito

src/XMLHttpRequest/XMLHttpRequest/createXMLHttpRequestOverride.ts
marcosvega91: We can assign timeout function only if is passed. In this way the error will be thrown if xhr is not used in the right way

kettanaito:
I've reworked the test and ensures that callbacks _and_ event listeners were propagated between XHR instances, since previously if you listed on `timeout` via `req.addEventListener('timeout', handler)` that would not trigger the timeout event, and the test would fail.

The changes are ready to go.

PR #48, opened 2020-08-18
Accept the list of interceptors

kettanaito:
> This pull request **includes breaking changes**.

## Changes

- `RequestInterceptor` class now accepts a list of interceptors to use.

```diff
import { RequestInterceptor } from 'node-request-interceptor'
+import { interceptClientRequest } from 'node-request-interceptor'

-new RequestInterceptor()
+new RequestInterceptor([interceptClientRequest])
```

- Restructured the directories to include `interceptors` and `presets` folders.

```js
import { RequestInterceptor } from 'node-request-interceptor'
import withDefaultInterceptors from 'node-request-interceptor/presets/default'

new RequestInterceptor(withDefaultInterceptors)
```

- Adjusted test suites to reflect the usage of the interceptors list.

## Motivation

Accepting a list of interceptors allows to configure a `RequestInterceptor` instance to better suit the usage environment (NodeJS/React Native/etc.). This also adds support for custom request interceptors, if need be.

## GitHub

- Originated from https://github.com/mswjs/msw/issues/203


kettanaito:
Hey, @Tommos0 👋 Following [our discussion](https://github.com/mswjs/msw/issues/203#issuecomment-675288788), I've refactored the library to support granular interceptors. May I please ask for your review on this?

One point that troubles me, is that constructing a `new RequestInterceptor` returns an `interceptor`, so it feels wrong to call `ClientRequest` and `XMLHttpRequest` "interceptors". Feels like a poor naming choice on my side. I can't say I like "overrides" more, as it's too generic and doesn't describe what the modules in that group represent. What do you think?

timdeschryver:
I'm not comfortable with this codebase, but after giving this PR a quick look I think it's a good refactor.
Just one personal nit, is that I don't like default exports, I mostly use named exports.

About the naming, I've seen some debug statements with `patch` in those files, could we rename the request to something with `Patch` e.g. `ClientRequestPatch(er)` ?

kettanaito:
@timdeschryver thank you for looking into these changes! Could you please reference to which `debug` statements you refer?

timdeschryver:
Sure, there are even comments! 

- https://github.com/mswjs/node-request-interceptor/blob/master/src/RequestInterceptor.ts#L20
- https://github.com/mswjs/node-request-interceptor/blob/master/src/http/ClientRequest/ClientRequestOverride.ts#L257
- https://github.com/mswjs/node-request-interceptor/blob/master/src/http/ClientRequest/ClientRequestOverride.ts#L268
- https://github.com/mswjs/node-request-interceptor/blob/master/src/XMLHttpRequest/override.ts#L18

kettanaito:
Thanks! Those `debug` calls already are bound to a certain context:

https://github.com/mswjs/node-request-interceptor/blob/5da41a9b435f9da0b88aa93a870c7aaf6577923b/src/http/ClientRequest/ClientRequestOverride.ts#L32

For example, when called such `debug` function it would produce an output that includes a specific context:

```
http GET /example reverting patches...
```

Do you still find such format confusing?

In case of [this](https://github.com/mswjs/node-request-interceptor/blob/master/src/RequestInterceptor.ts#L20), there is no way to specify which module patches will get restored, because these changes introduce a parametric nature of modules: you can provide whichever interceptors you want to `RequestInterceptor`. Its `restore` method will revert the patches those interceptors make to various modules.

README.md
```diff
@@ -80,6 +105,8 @@ interceptor.use((req) => {
 
 ##### Response mocking
 
+When a request middleware returns a [`MockedResponse`](#MockedResponse) object, it's being used as a mocked response to the intercepted request. This library aligns an abstract mocked response interface to the actual response in correspondence to the module that has issued the request.
```
msutkowski, line 108:
```suggestion
When a request interceptor middleware returns a [`MockedResponse`](#MockedResponse) object, it will then be returned as the mocked response to the intercepted request. This library maps an abstract mocked response interface to the expected response of the module that issued the request.
```

README.md
```diff
@@ -100,7 +127,7 @@ interceptor.use((req) => {
 
 #### `.restore(): void`
 
-Restores all patched modules and stops the interception.
+Restores all patched modules and stops the requests interception.
```
msutkowski, line 130:
```suggestion
Restores all patched modules and stops the interception of any future requests.
```

README.md
msutkowski: The second sentence might not be needed here as it's a little confusing to read, no matter which way it's written.

kettanaito: The point of the second sentence is to explain that NRI uses an abstract response interface and then automatically creates a proper response object from it, depending on what module has issued the request (http/xhr). That's useful to highlight, because different request issuing modules create and expect a different response shape.

msutkowski: That makes sense, thanks for the additional clarification! I don't know why I had so much trouble reading that line.

timdeschryver:
@kettanaito no that makes sense. As I said it's because of my little knowledge of the codebase. 
Thanks for the clarification!

Tommos0:
Looks good, nice work! :+1:

PR #50, opened 2020-08-24
fix: null value of http.ClientRequest 

marcosvega91:
The PR will fix https://github.com/mswjs/node-request-interceptor/issues/49

The problem was that on next when you make an hard refresh the old instance of `RequestInterceptor` was not cleared.
This behavior could cause an issue because the `http.ClientRequest` sometime was null.

I don't know if we should allow the user to create multiple `RequestInterceptor` on the same interceptors 🤔

### DIFF
--- a/src/interceptors/ClientRequest/ClientRequestOverride.ts
+++ b/src/interceptors/ClientRequest/ClientRequestOverride.ts
@@ -249,25 +249,18 @@ export function createClientRequestOverrideClass(
       let req: ClientRequest
       debug('using', performOriginalRequest)
 
-      const { ClientRequest } = http
-
       // Decide whether to use HTTPS based on the URL protocol.
       // XHR can trigger http.request for HTTPS URL.
       if (url.protocol === 'https:') {
         debug('reverting patches...')
+        const { ClientRequest } = http
 
         http.ClientRequest = originalClientRequest
-
-        // Override the global pointer to the original client request.
-        // This way whenever a bypass call bubbles to `handleRequest`
-        // it always performs respecting this `ClientRequest` restoration.
-        originalClientRequest = null as any
 
         req = performOriginalRequest(options)
 
         debug('re-applying patches...')
         http.ClientRequest = ClientRequest
-        originalClientRequest = ClientRequest
       } else {
         req = performOriginalRequest(options)
       }
--- a/src/interceptors/ClientRequest/index.ts
+++ b/src/interceptors/ClientRequest/index.ts
@@ -106,11 +106,6 @@ export const interceptClientRequest: Interceptor = (middleware) => {
   return () => {
     debug('restoring patches...')
 
-    //restore the original ClientReqest if it was overridden
-    if (originalClientRequest) {
-      http.ClientRequest = originalClientRequest
-      originalClientRequest = null as any
-    }
     Object.values(patchedModules).forEach(({ module, request, get }) => {
       module.request = request
       module.get = get
--- a/src/interceptors/ClientRequest/index.ts
+++ b/src/interceptors/ClientRequest/index.ts
@@ -24,18 +24,21 @@ function handleRequest(
   middleware: RequestMiddleware,
   args: any[]
 ): ClientRequest {
+  //The first time we execute this, I'll save the original ClientRequest.
+  //This because is used to restore the dafault one later
   if (!originalClientRequest) {
-    const ClientRequestOverride = createClientRequestOverrideClass(
-      middleware,
-      originalMethod,
-      originalClientRequest
-    )
-
-    debug('patching native http.ClientRequest...')
-
-    // @ts-ignore
-    http.ClientRequest = ClientRequestOverride
+    originalClientRequest = http.ClientRequest
   }
+
+  const ClientRequestOverride = createClientRequestOverrideClass(
+    middleware,
+    originalMethod,
+    originalClientRequest
+  )
+  debug('patching native http.ClientRequest...')
+  //Only http.ClientRequest is overridden because https uses http
+  //@ts-ignore
+  http.ClientRequest = ClientRequestOverride
 
   debug('new http.ClientRequest (origin: %s)', protocol)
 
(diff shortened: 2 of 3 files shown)


src/interceptors/ClientRequest/ClientRequestOverride.ts
```diff
@@ -249,25 +249,18 @@ export function createClientRequestOverrideClass(
       let req: ClientRequest
       debug('using', performOriginalRequest)
 
-      const { ClientRequest } = http
-
       // Decide whether to use HTTPS based on the URL protocol.
       // XHR can trigger http.request for HTTPS URL.
       if (url.protocol === 'https:') {
         debug('reverting patches...')
+        const { ClientRequest } = http
```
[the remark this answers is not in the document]
marcosvega91: yes as `nock` does when the http.ClientRequest is restored to the original value (before patch) before the request and then after that patch it again

src/interceptors/ClientRequest/ClientRequestOverride.ts
```diff
@@ -249,25 +249,18 @@ export function createClientRequestOverrideClass(
...
 
-        // Override the global pointer to the original client request.
-        // This way whenever a bypass call bubbles to `handleRequest`
-        // it always performs respecting this `ClientRequest` restoration.
-        originalClientRequest = null as any
-
         req = performOriginalRequest(options)
 
         debug('re-applying patches...')
         http.ClientRequest = ClientRequest
```
kettanaito, line 263: Should we restore this reference to the patched `ClientRequest` class when the `req` is actually finished (either successfully or unsuccessfully)?

I'm thinking that `performOriginalRequest` creates a request, but may provision references to `ClientRequest` between the moment the request instance is created and the moment the request is actually finished. On this line we restore the `http.ClientRequest` to the patched version, so if an underlying logic attempts to reference `ClientRequest` during the _original_ request, it would still point to the patched instance. 

That being said, how likely it is to reference `ClientRequest`? I don't think that's a pattern to create a request instance within a request instance. Especially, since that is how NodeJS operates, so you don't normally have access to this. My concern is valid only in corner cases like:

```js
// usage.js
// This request will be constructed using the original unpatched ClientRequest class.
const req = http.get('https://api.com/user', (res) => {
  // This request, however, will be constructed using the patched class, as we restored the ref.
  const anotherReq = http.get('https://another.domain')
})
```

In a way this is actually intentional, as that second request in a callback function should also be intercepted and cam potentially have a mocked response. Thank you for bearing with me through this comment 🦆

src/interceptors/ClientRequest/index.ts
```diff
@@ -103,6 +106,11 @@ export const interceptClientRequest: Interceptor = (middleware) => {
   return () => {
     debug('restoring patches...')
 
+    //restore the original ClientReqest if it was overridden
+    if (originalClientRequest) {
```
kettanaito, line 110: I think this condition is redundant. This is the restore function that a module interceptor returns, and it must always restore any patched references. 

Do you see any harm in dropping the condition? Can it be that we restore the `http.ClientRequest` to a patched version due to some concurrency or multiple instances of `RequestInterceptor`?

src/interceptors/ClientRequest/index.ts
```diff
@@ -24,19 +24,22 @@ function handleRequest(
...
 
+  const ClientRequestOverride = createClientRequestOverrideClass(
+    middleware,
+    originalMethod,
+    originalClientRequest
+  )
+  debug('patching native http.ClientRequest...')
+  //Only http.ClientRequest is overridden because https uses http
+  //@ts-ignore
+  http.ClientRequest = ClientRequestOverride
```
kettanaito, line 41: When NRI will construct an actual request (non-mocked), it will also bubble to this `handleRequest` function. In that case the flow will be:

```
handleRequest
has original client request? yes
create an override class
patch native "http.ClientRequest" with the override class
create an instance of an override class
```

I think that's not correct. In case of original request this `handleRequest` function must construct an instance of _unpatched_ `http.ClientRequest` class. That's why there's this original client request reference is introduced.

kettanaito (changes requested):
Thank you so much for your work on this, @marcosvega91!
I've left a few comments, mainly clarification and a few observations. Your knowledge on this after debugging would be invaluable to this discussion!

src/interceptors/ClientRequest/index.ts
marcosvega91: Yes you are right we can drop

src/interceptors/ClientRequest/ClientRequestOverride.ts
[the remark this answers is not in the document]
kettanaito: I see. Still a mention of caution to treat `nock` as an example, but not a guideline, as NRI operates substantially different than `nock`. `nock` has a request matching logic, so they can determine whether to mock (patch) a request before it happens. NRI, on the other hand, is agnostic of your decision to patch a request, so we always create a request instance and decide whether to perform _another_ (actual) request if there's no mocks for it.

src/interceptors/ClientRequest/index.ts
kettanaito: I believe even if there's a multiple instances issue, calling a restore function must still restore what it patched. Then the issue would be in multiple instances, while the behavior of restore would remain as expected.

src/interceptors/ClientRequest/index.ts
marcosvega91: When the request is not mocked it is handled by the patched `ClientRequest` but `performOriginalRequest` will be called. `performOriginalRequest` function will call the original http/https.request so our custom instance will only proxy to the normal flow

src/interceptors/ClientRequest/ClientRequestOverride.ts
marcosvega91: http.get and http.request will create an instance of the ClientRequest when one of them is called so after performOriginalRequest we can restore the ClientRequest

src/interceptors/ClientRequest/index.ts
marcosvega91: Maybe I don't understand your point. If I restore an instance then when a new request is intercepted `ClientRequest` will be patch again

kettanaito (approved):
The changes look solid! Thank you, @marcosvega91.

I would treat this as a breaking change (minor version bump).

PR #57, opened 2020-09-08
Uses local HTTP/HTTPS server for integration tests

kettanaito:
## Changes

- Removes dependency on the external servers (`httpbin`) to increase reliability of integration tests.
- Adjusts test suits to use the `createServer` utility and make requests to the spawned server URLs.
- Adjusts `https` requests to use a custom `httpsAgent` that allows requests to self-signed HTTPS servers.
- Adjusts `XMLHttpRequest` (in `jsdom`) to set `strictSSL: false` to allow requests to self-signed HTTPS servers.


kettanaito:
All tests are passing offline.

test/intercept/https/https.request.test.ts
```diff
@@ -1,15 +1,31 @@
...
 import { InterceptedRequest } from '../../../src/glossary'
 import { prepare, httpsRequest } from '../../helpers'
+import { ServerAPI, createServer, httpsAgent } from '../../utils/createServer'
 
 let requestInterceptor: RequestInterceptor
 let pool: InterceptedRequest[] = []
+let server: ServerAPI
+
+beforeAll(async () => {
+  server = await createServer((app) => {
```
marcosvega91, line 16: it's possible to `createServer` in a general `beforeAll` valid for all tests?
Then in very test where you need the server you will add only the routes

test/intercept/https/https.request.test.ts
```diff
@@ -20,37 +36,40 @@ afterEach(() => {
...
-afterAll(() => {
+afterAll(async () => {
   requestInterceptor.restore()
+  await server.close()
 })
 
 test('intercepts an HTTPS GET request', async () => {
   const request = await prepare(
-    httpsRequest('https://httpbin.org/get?userId=123', {
+    httpsRequest(server.makeHttpsUrl('/user?id=123'), {
```
marcosvega91, line 46: I think is better to create a request like `/user/123` instead of `/user?id=123`

marcosvega91 (changes requested):
Hi @kettanaito everything very good and cool. I have only a question and a little change 🚀

marcosvega91:
few late hahahaha

test/intercept/https/https.request.test.ts
kettanaito: If I recall correctly, the query parameter is there to test if the request URL gets parsed correctly into query parameters. In case of `/user/123` it becomes a part of a regular URL, and it drops the assertion of query parameters that way.

test/intercept/https/https.request.test.ts
kettanaito: I thought about that as well, but decided to spawn a server in each test explicitly. First of all in order to control when a server is needed, as not all tests may require a local server. Secondly, it's always good to start with a non-harmful repetition and either wait until it becomes harmful (i.e. degrades performance), or leave it forever.

kettanaito:
@marcosvega91, thank you for the review! Sorry I merged it prematurely :)

test/intercept/https/https.request.test.ts
marcosvega91: Ah yes i didn't think about it. I thought only the form. Ignore it 🤣

test/intercept/https/https.request.test.ts
marcosvega91: It can make sense 👍

PR #62, opened 2020-10-22
Resolve protocol from agent, if set

sean-hernon:
In `normalizeHttpRequestParams`, the protocol is taken from the url, if there is none set in the options https://github.com/mswjs/node-request-interceptor/blob/09ec3011638131f9b06b63d106c762ab0d59d9f9/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts#L81

As the comment above the code states, this is to avoid an error of mismatching protocols, from this section of the node code https://github.com/nodejs/node/blob/d84f1312915fe45fe0febe888db692c74894c382/lib/_http_client.js#L142-L145

In the same code, we can see that the expected protocol comes from `this.agent`, if set and lines 125 and 140 show us that this will be the agent that is set in the options, if any exists.

Therefore in order to avoid the error when this is set, we should set the url protocol to match this one (from the url returned here https://github.com/mswjs/node-request-interceptor/blob/09ec3011638131f9b06b63d106c762ab0d59d9f9/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts#L67)

As we default to http when there is no protocol or certificate, if we have an https agent set, but no protocol set on the options, we will have `options.agent.protocol=https:` and `url.protocol=http:`, then when we set options protocol from the url one (https://github.com/mswjs/node-request-interceptor/blob/09ec3011638131f9b06b63d106c762ab0d59d9f9/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts#L81) we will also have `options.protocol=http:`, which gives us the mismatching protocol error in the node js code.

### DIFF
--- a/src/utils/getUrlByRequestOptions.test.ts
+++ b/src/utils/getUrlByRequestOptions.test.ts
@@ -1,5 +1,5 @@
 import { getUrlByRequestOptions } from './getUrlByRequestOptions'
-import { RequestOptions } from 'https'
+import { RequestOptions, Agent } from 'https'
 
 test('returns a URL based on the basic RequestOptions', () => {
   const options: RequestOptions = {
@@ -12,6 +12,20 @@ test('returns a URL based on the basic RequestOptions', () => {
   expect(url).toBeInstanceOf(URL)
   expect(url).toHaveProperty('port', '')
   expect(url).toHaveProperty('href', 'https://127.0.0.1/resource')
+})
+
+test('resolves protocol from agent, if one exists', () => {
+  const options: RequestOptions = {
+    host: '127.0.0.1',
+    path: '/',
+    agent: new Agent()
+  }
+  const url = getUrlByRequestOptions(options)
+
+  expect(url).toBeInstanceOf(URL)
+  expect(url).toHaveProperty('protocol', 'https:')
+  expect(url).toHaveProperty('port', '')
+  expect(url).toHaveProperty('href', 'https://127.0.0.1/')
 })
 
 test('resolves protocol to "http" given no explicit protocol and no certificate', () => {
--- a/src/utils/getUrlByRequestOptions.test.ts
+++ b/src/utils/getUrlByRequestOptions.test.ts
@@ -1,5 +1,6 @@
+import { Agent as HttpAgent } from 'http'
+import { RequestOptions, Agent as HttpsAgent } from 'https'
 import { getUrlByRequestOptions } from './getUrlByRequestOptions'
-import { RequestOptions, Agent } from 'https'
 
 test('returns a URL based on the basic RequestOptions', () => {
   const options: RequestOptions = {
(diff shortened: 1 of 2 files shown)


src/utils/getUrlByRequestOptions.ts
```diff
@@ -16,6 +16,8 @@ export function getUrlByRequestOptions(
 
   debug('creating URL from options:', options)
 
+  options.protocol = options.protocol || (options.agent && (options.agent as RequestOptions).protocol) || undefined
```
kettanaito, line 19: Can we use optional chaining for a shorter reference?

```js
options.protocol = options.protocol || options.agent?.protocol || undefined
```

Does this raise a type violation?

src/utils/getUrlByRequestOptions.test.ts
```diff
@@ -14,6 +14,20 @@ test('returns a URL based on the basic RequestOptions', () => {
...
+test('resolves protocol from agent, if one exists', () => {
+  const options: RequestOptions = {
+    host: '127.0.0.1',
+    path: '/',
+    agent: new Agent()
+  }
+  const url = getUrlByRequestOptions(options)
+
+  expect(url).toBeInstanceOf(URL)
+  expect(url).toHaveProperty('protocol', 'https:')
```
kettanaito, line 26: This used to be `http:` in the past, correct?

Since it doesn't have an explicit `protocol` set, we defaulted to `http:`, while this particular case used `https.Agent`.

kettanaito:
I'm blown away by the quality and profoundness of your contributions, @sean-hernon.

Please see some of my questions above. Overall this looks fantastic. If you have nothing to add we can merge this.

kettanaito (approved):

src/utils/getUrlByRequestOptions.ts
marcosvega91: I see that something similar is also for port. Maybe we should do the same ?

https://github.com/nodejs/node/blob/d84f1312915fe45fe0febe888db692c74894c382/lib/_http_client.js#L158-L161

src/utils/getUrlByRequestOptions.ts
```diff
@@ -16,11 +17,18 @@ export function getUrlByRequestOptions(
 
   debug('creating URL from options:', options)
 
+  // Inherit the protocol from the Agent, if present.
+  if (options.agent instanceof Agent) {
```
kettanaito, line 21: `agent: Agent | boolean`. To prevent false positive with `agent && agent.protocol` (where `agent` may be `true` and `true.protocol` throws`) I'm explicitly checking if it's an instance of `http.Agent`. `https.Agent` extends `http.Agent`, so the check would pass for it as well.

kettanaito (approved):
This is a great improvement, @sean-hernon. 
Thank you and welcome to contributors! 🎉

src/utils/getUrlByRequestOptions.ts
kettanaito: Great point, @marcosvega91. Added this in e6e4edf7588297133666ab6f4c300cdac4f85d53

PR #63, opened 2020-10-23
Support legacy Url objects

sean-hernon:
In addition to the WHATWG `URL` object, node js also has a legacy `Url` object, created through `url.parse`, which is accepted when creating http requests. Although this type is deprecated as of node 11, there are many packages which are using it.

This change is to allow interoperability with a wider range of packages by normalising the legacy Url to a WHATWG URL.

### DIFF
--- a/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
@@ -1,3 +1,5 @@
+import { parse } from 'url';
+import { getUrlByRequestOptions } from '../../../utils/getUrlByRequestOptions';
 import { normalizeHttpRequestParams } from './normalizeHttpRequestParams'
 
 test('handles [string, callback] input', () => {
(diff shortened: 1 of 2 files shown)


src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
```diff
@@ -1,4 +1,5 @@
 import { normalizeHttpRequestParams } from './normalizeHttpRequestParams'
+import { parse } from 'url';
```
kettanaito, line 2: Nitpick: could you re-order external imports to the top?

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
```diff
@@ -42,6 +43,10 @@ export function normalizeHttpRequestParams(
 
   debug('arguments', args)
 
+  if (args[0].constructor.name === 'Url') {
+    args[0] = new URL((args[0] as Url).href)
```
kettanaito, line 47: So the intention is to coerce the legacy URL into the modern URL instance and then proceed with the request params retrieval? If so, maybe we should put this coercion elsewhere to preserve the purpose of this normalization function?

sean-hernon: Is the suggestion to move this section into a separate function, or to have the coercion occur before arriving at this function? If the latter, where would be the best place, in your opinion?

kettanaito: I confused this function with `resolveRequestOptions`. I think this is the right place to perform this kind of coercion, as it's a part of the request parameters normalization (handling a vast nature of NodeJS `ClientRequest` arguments).

How would you improve this coercion in regards to the `if` statements below?

sean-hernon: By improve the coercion, do you mean in terms of style or that there may be some potential semantic improvement?

kettanaito: Yes. When we fall into this closure we know that `args[0]` is an instance of a `URL`. I'm thinking if we can utilize that knowledge in the following `args[0]` check (if it's a string, if it has `origin`, etc.). For instance, if we know it's a `URL` instance, the first `if` (string check) becomes redundant and we know that ahead of time. 

I wonder what are your thoughts on this.

sean-hernon: We do some common work at the end of the function, which we could move into a separate function and then return early in the main function, instead of having an if-else chain.

```
export function normalizeHttpRequestParams(
  ...args: HttpRequestArgs
): [URL, RequestOptions & RequestSelf, HttpRequestCallback?] {
  let url: URL
  let options: RequestOptions & RequestSelf
  let callback: HttpRequestCallback | undefined = resolveCallback(args)

  debug('arguments', args)

  if (args[0].constructor.name === 'Url') {
    return normalizeUrlHttpRequestParams(new URL((args[0] as Url).href), options, callback)
  }

  if (typeof args[0] === 'string') {
    debug('given a location string:', args[0])

    url = new URL(args[0])
    debug('created a URL:', url)

    debug('created request options:', options)

    return finaliseHttpRequestParams(url, options, callback)
  } 

  if ('origin' in args[0]) {
    return normalizeUrlHttpRequestParams(args[0], options, callback)
  }

  if ('method' in args[0]) {
    options = args[0]
    debug('given request options:', options)

    url = getUrlByRequestOptions(options)
    debug('created a URL:', url)

    return finaliseHttpRequestParams(url, options, callback)
  }

  throw new Error(
      `Failed to construct ClientRequest with these parameters: ${args}`
   )
}

function normalizeUrlHttpRequestParams(url: URL, options: RequestOptions & RequestSelf, callback: HttpRequestCallback?) {
    debug('given a URL:', url)

    debug('created request options', options)

    return finaliseHttpRequestParams(url, options, callback)
}

function finaliseHttpRequestParams(url: URL, options: RequestOptions & RequestSelf, callback: HttpRequestCallback?) {
   // Enforce protocol on `RequestOptions` so when `ClientRequest` compares
  // the agent protocol to the request options protocol they match.
  // @see https://github.com/nodejs/node/blob/d84f1312915fe45fe0febe888db692c74894c382/lib/_http_client.js#L142-L145
  // This prevents `Protocol "http:" not supported. Expected "https:"` exception for `https.request` calls.
  options.protocol = options.protocol || url.protocol

  debug('resolved URL:', url)
  debug('resolved options:', options)

  return [url, options, callback]
}
```

Apologies for any slight syntax errors

sean-hernon: Alternatively, we can return early with a recursive call. We will still incur the extra check for string, but it will read quite nicely.

i.e.

```
if (args[0].constructor.name === 'Url') {
    return normalizeHttpRequestParams(new URL((args[0] as Url).href), options, callback)
}
```

Then in this second call, we will hit the `URL` handling block.

sean-hernon: So, I did the (directly) above in the end, because there is also the case of _relative_ legacy urls to consider. I added some extra tests and and extra condition to cover this case.

When we have a relative legacy url, the only interesting thing inside is the path (pathname and query string), so it doesn't make sense to use it to build a URL object, since all of the other parts would be provided by any options set. So instead, in this case, I am copying the path to any existing options object and calling the method again. Any existing path in the options object is preferred to the one in the relative url, to be consistent with current node behaviour. https://github.com/nodejs/node/blob/d84f1312915fe45fe0febe888db692c74894c382/lib/_http_client.js#L122

When the legacy url is absolute, then it's easier to simply create a WHATWG URL and follow the existing flow for URL.

I also added a small function to check if an argument is a `RequestOptions`, which we could re-use in the other PR (instead of checking for the `method` property).

sean-hernon: I was thinking about what you said on the other PR and I've moved the block down into the other if-else structure, so that typescript can infer the `Url` type. I had to check that `method` was not on the object, because it was breaking some of the fetch tests, without that.

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
```diff
@@ -67,6 +72,30 @@ export function normalizeHttpRequestParams(
...
+        We are dealing with a relative url, so use the path as an "option" and
+        merge-in any existing options, giving priority to exising options -- i.e. a path in any
+        existing options will take precedence over the one contained in the url. This is consistent
+        with the behaviour in ClientRequest.
+        
+        https://github.com/nodejs/node/blob/d84f1312915fe45fe0febe888db692c74894c382/lib/_http_client.js#L122
+      */
+      debug('given a relative legacy url:', args[0])
+
+      return isRequestOptions(args[1])
```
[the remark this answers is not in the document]
kettanaito: If we already have `isObject` and it satisfies the logic constrains, then please use it instead.

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
```diff
@@ -67,6 +72,30 @@ export function normalizeHttpRequestParams(
...
+        ? normalizeHttpRequestParams({path: args[0].path, ...args[1]}, args[2])
+        : normalizeHttpRequestParams({path: args[0].path}, args[1] as HttpRequestCallback)
+    }
+
+    debug('given an absolute legacy url:', args[0])
+
+    //We are dealing with an absolute url, so convert to WHATWG and try again
+    return normalizeHttpRequestParams(...[new URL(args[0].href), ...args.slice(1)] as HttpRequestArgs)
```
kettanaito, line 97: This argument is quite hard to read through. Do you think we could break it down into pieces?

kettanaito (changes requested):

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
sean-hernon: I've added some checks here to allow the types to be inferred by typescript, instead.

kettanaito (approved):
This is a great improvement over request params support.
Thank you, @sean-hernon!

PR #64, opened 2020-10-23
Allow request options to be empty and default hostname to 'localhost'

sean-hernon:
The `RequestOptions` object does not have any required properties, therefore we should not rely on `method` for its identification. Any object should be accepted as request options.

Since the hostname and host are also optional, we should not rely on either of them being set when constructing the URL. If node cannot find any set in hostname or host, then it defaults to `localhost` (https://github.com/nodejs/node/blob/d84f1312915fe45fe0febe888db692c74894c382/lib/_http_client.js#L163)

In order to retain parity, the default hostname should be this when creating the URL.

### DIFF
--- a/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
@@ -116,6 +116,16 @@ test('handles [RequestOptions, callback] input', () => {
   expect(callback).toHaveProperty('name', 'cb')
 })
 
+test('handles [Empty RequestOptions, callback] input', () => {
+  const [_, __, callback] = normalizeHttpRequestParams(
+    {},
+    function cb() {}
+  )
+
+  // Callback must be preserved
+  expect(callback).toHaveProperty('name', 'cb')
+})
+
 /**
  * @see https://github.com/mswjs/node-request-interceptor/issues/19
  */
--- a/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
@@ -60,7 +60,7 @@ export function normalizeHttpRequestParams(
     debug('created request options', options)
 
     callback = resolveCallback(args)
-  } else if ('method' in args[0]) {
+  } else if (Object.prototype.toString.call(args[0]) === '[object Object]') {
     options = args[0]
     debug('given request options:', options)
 
--- a/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
@@ -2,6 +2,7 @@ import { RequestOptions } from 'https'
 import { HttpRequestCallback, RequestSelf } from '../../../glossary'
 import { getRequestOptionsByUrl } from '../../../utils/getRequestOptionsByUrl'
 import { getUrlByRequestOptions } from '../../../utils/getUrlByRequestOptions'
+import { isObject } from '../../../utils/isObject'
 
 const debug = require('debug')('http normalizeHttpRequestParams')
 
@@ -42,6 +43,8 @@ export function normalizeHttpRequestParams(
 
   debug('arguments', args)
 
+  // Convert a url string into a URL instance
+  // and derive request options from it.
   if (typeof args[0] === 'string') {
     debug('given a location string:', args[0])
 
@@ -52,7 +55,10 @@ export function normalizeHttpRequestParams(
     debug('created request options:', options)
 
     callback = resolveCallback(args)
-  } else if ('origin' in args[0]) {
+  }
+  // Handle a given URL instance as-is
+  // and derive request options from it.
+  else if ('origin' in args[0]) {
     url = args[0]
     debug('given a URL:', url)
 
@@ -60,7 +66,10 @@ export function normalizeHttpRequestParams(
     debug('created request options', options)
 
     callback = resolveCallback(args)
-  } else if (Object.prototype.toString.call(args[0]) === '[object Object]') {
+  }
+  // Handle a given request options object as-is
+  // and derive the URL instance from it.
+  else if (isObject(args[0])) {
     options = args[0]
     debug('given request options:', options)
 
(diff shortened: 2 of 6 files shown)


src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
```diff
@@ -60,7 +60,7 @@ export function normalizeHttpRequestParams(
     debug('created request options', options)
 
     callback = resolveCallback(args)
-  } else if ('method' in args[0]) {
+  } else if (Object.prototype.toString.call(args[0]) === '[object Object]') {
```
kettanaito, line 63: Does this result in correct typing in this if statement's body? I don't think that TypeScript would like this comparison.

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
```diff
@@ -116,6 +116,16 @@ test('handles [RequestOptions, callback] input', () => {
   expect(callback).toHaveProperty('name', 'cb')
 })
 
+test('handles [Empty RequestOptions, callback] input', () => {
+  const [_, __, callback] = normalizeHttpRequestParams(
```
kettanaito, line 120: Are you sure you can dispatch `new ClientRequest` without any options? That doesn't seem to be a realistic scenario. When would such thing happen in practice?

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
sean-hernon: Typescript does not complain.

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
sean-hernon: Yes, since each of the properties on `RequestOptions` is optional.

In practice, it's difficult to say whether you would see it used much without _any_ of the options being set, but the point is that since none of them are mandatory, there isn't a single one that can be relied upon, so a sane output must be produced in the case where each option is missing. The easiest way to test this is but setting no options.

I did not test the output, since the URL construction is delegated to another function, so here the test will pass if we do not get an exception. The test for `getUrlByRequestOptions` asserts the correct default URL is produced.

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
sean-hernon: I've added a `isRequestOptions` function in PR #63. Do you think that the approach there is better, or do you feel that the check would be too lax, for this use case?

https://github.com/mswjs/node-request-interceptor/pull/63/files#diff-dd33cf0eafb9cac75538d7fd8fb5be0147f2224d1e471400203a8a28ce5947fdR34

kettanaito: > Typescript does not complain.

I know, I'm more interested if it can correctly determine the `args[0]` and other types within this closure. Can it?

sean-hernon: It infers that the type is `RequestOptions`, if I change the `else if` to an `else`, because the first two blocks are matching the other two types. If I keep the third `else if` as an `else if`, but change the condition to something which matches no three of the types, like `else if(typeof args[0] === 'number')`, then it resolves the type of args[0] as `never`. 

If change the condition to something like `Object.prototype.toString.call(args[0]) === 'blah'`, then it still infers `args[0]` as `RequestOptions`. So it seems that it's deciding that it is `RequestOptions` because we are in an `else` and the other two possibilities have been caught by the other blocks, and that it is not using _this_ style of condition to infer the type at all.

In short, it has decided that we have a `RequestOptions` object _unless_ the condition explicitly makes this impossible.

sean-hernon: I guess since the previous condition was assuming we had an object at that stage, it could be argued that since RequestOptions has no mandatory properties, perhaps the original `else-if` should be swapped for an `else` and the exception should be removed?

kettanaito: @sean-hernon so you think it's safe to remove the last `else` clause with an exception, because now this check handles any given object?

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
```diff
@@ -52,15 +55,21 @@ export function normalizeHttpRequestParams(
...
 
     options = resolveRequestOptions(args, url)
     debug('created request options', options)
 
     callback = resolveCallback(args)
-  } else if ('method' in args[0]) {
+  }
+  // Handle a given request options object as-is
+  // and derive the URL instance from it.
+  else if (isObject(args[0])) {
```
kettanaito, line 72: I've isolated this predicate into its own utility, making this code easier to read.
I've also added a brief explanatory comment above each `if` statement, so it could be quicker to understand what this case handles.

kettanaito (approved):
I find this changes awesome ✨ Approved.
Let me know if you think we should provide any additional adjustments (like removing the `else` statement with the exception) before merging.

Outstanding work, @sean-hernon.

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
sean-hernon: I would say it's safe to remove, because since we already use the `in` operator on the argument at this point, we are dealing with something object-like, so even if someone passed a function or an array, then it would simply result in lots of `undefined` in ClientRequest and the defaults would be used. Checking that we have an actual plain `Object object` is explicitly checking that we're likely to have something more sane, though.

sean-hernon:
> I find this changes awesome ✨ Approved.
> Let me know if you think we should provide any additional adjustments (like removing the `else` statement with the exception) before merging.
> 
> Outstanding work, @sean-hernon.

I would say let's keep the `else`, in case someone passes something that we don't expect. That way we will fail instead of silently treating the parameter as `RequestOptions`. This way we can easily see any cases that we've missed, that result from a different type of object being passed.

PR #74, opened 2020-12-02
determine secure protocol by 443 port

hamishhossack:
Relates to issue on [msw #486](https://github.com/mswjs/msw/issues/486).

When port is 443 use the secure https protocol.

### DIFF
--- a/src/utils/getUrlByRequestOptions.test.ts
+++ b/src/utils/getUrlByRequestOptions.test.ts
@@ -82,7 +82,7 @@ test('resolves protocol to "https" given no explicit protocol, but port is 443',
 
   expect(url).toBeInstanceOf(URL)
   expect(url).toHaveProperty('port', '')
-  expect(url).toHaveProperty('href', 'https://127.0.0.1:443/resource')
+  expect(url).toHaveProperty('href', 'https://127.0.0.1/resource')
 })
 
 test('inherits "port" if given', () => {
--- a/src/utils/getUrlByRequestOptions.test.ts
+++ b/src/utils/getUrlByRequestOptions.test.ts
@@ -85,19 +85,6 @@ test('resolves protocol to "https" given no explicit protocol, but port is 443',
   expect(url).toHaveProperty('href', 'https://127.0.0.1/resource')
 })
 
-test('resolves protocol to "https" given no explicit protocol, but port is 443', () => {
-  const options: RequestOptions = {
-    host: '127.0.0.1',
-    port: 443,
-    path: '/resource',
-  }
-  const url = getUrlByRequestOptions(options)
-
-  expect(url).toBeInstanceOf(URL)
-  expect(url).toHaveProperty('port', '')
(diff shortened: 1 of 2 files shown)


src/utils/getUrlByRequestOptions.test.ts
```diff
@@ -72,6 +72,19 @@ test('resolves protocol to "https" given no explicit protocol, but certificate',
...
+test('resolves protocol to "https" given no explicit protocol, but port is 443', () => {
+  const options: RequestOptions = {
+    host: '127.0.0.1',
+    port: 443,
+    path: '/resource',
+  }
+  const url = getUrlByRequestOptions(options)
+
+  expect(url).toBeInstanceOf(URL)
+  expect(url).toHaveProperty('port', '443')
```
hamishhossack, line 84: in fact this was failing locally due to `new URL('/resource', 'https://127.0.0.1')` not returning a valid url.

Created the PR to see if the same is happening on ci.

src/utils/getUrlByRequestOptions.ts
```diff
@@ -31,7 +32,9 @@ function getProtocolByRequestOptions(
     return agentProtocol
   }
 
-  return options.cert ? 'https:' : options.uri?.protocol || DEFAULT_PROTOCOL
+  const isSecureRequest = options.cert || options.port === SSL_PORT
+
+  return isSecureRequest ? 'https:' : options.uri?.protocol || DEFAULT_PROTOCOL
```
marcosvega91, line 37: I have only one doubt here. The port could be not the one in the `port` property of `options`.
I you have a look at `getPortByRequestOptions` the port can be also taken from the `agent`. 
I think that we should pass the port to the function.

marcosvega91 (changes requested):
Hi @hamishhossack thanks for your awesome work, I have only a doubt. Thanks a lot 😄

src/utils/getUrlByRequestOptions.ts
hamishhossack: I opted to use the port util in the protocol function as it seemed cleaner than spreading the port with the options?
Obviously not ideal since that util would be used twice, however I found it safer in terms of debugging and with minimal cost associated with it.

```
/**
 * Creates a `URL` instance from a given `RequestOptions` object.
 */
export function getUrlByRequestOptions(options: IsomorphicRequestOptions): URL {
  debug('request options', options)

  const port = getPortByRequestOptions(options)
  // This could mask where the protocol originates making it harder to debug
  const protocol = getProtocolByRequestOptions({ ...options, port })
...
```
What do you think?

marcosvega91 (approved):
For me is great :) @kettanaito any thoughts?

kettanaito (approved):
This looks great. Thank you for this improvement, @hamishhossack!

kettanaito:
Welcome to contributors, @hamishhossack! 🎉

PR #75, opened 2020-12-03
XMLHttpRequest: Handles "progress" event on omni-value response body

kettanaito:
## GitHub

- Fixes https://github.com/mswjs/msw/issues/487

## Changes

- The `progress` event of `XMLHttpRequest` now fires only when the mocked response body was set (there's a body to load), and the coerced response body according to `responseType` is present (coercion was successful, the mocked response body is valid against the specified `responseType`). 
- The `progress` event now converts the raw mocked response body (string) into a `Buffer`, to handle various body types (string, JSON, ArrayBuffer, Blob) and determine their length.
- Adds a test to assert missing/invalid response body handling of an `XMLHttpRequest` with `responseType = 'json'`. Actual XHR behavior was taken for the reference.

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -275,11 +275,17 @@ export const createXMLHttpRequestOverride = (
 
             debug('assigned response body', this.response)
 
-            // Trigger a progress event based on the mocked response body.
-            this.trigger('progress', {
-              loaded: this.response.length,
-              total: this.response.length,
-            })
+            if (mockedResponse.body && this.response) {
+              // Presense of the mocked response implies a response body (not null).
+              // Presece of the coerced `this.response` implies the mocked body is valid.
+              const bodyBuffer = Buffer.from(mockedResponse.body)
+
+              // Trigger a progress event based on the mocked response body.
+              this.trigger('progress', {
+                loaded: bodyBuffer.length,
+                total: bodyBuffer.length,
+              })
+            }
 
             // Explicitly mark the request as done, so its response never hangs.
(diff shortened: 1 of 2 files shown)


src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -275,11 +275,17 @@ export const createXMLHttpRequestOverride = (
...
 
-            // Trigger a progress event based on the mocked response body.
-            this.trigger('progress', {
-              loaded: this.response.length,
-              total: this.response.length,
-            })
+            if (mockedResponse.body && this.response) {
+              // Presense of the mocked response implies a response body (not null).
+              // Presece of the coerced `this.response` implies the mocked body is valid.
+              const bodyBuffer = Buffer.from(mockedResponse.body)
```
kettanaito, line 281: At this point, we are always creating a `Buffer` from a _string_ body. 

I don't think converting `Blob`/`ArrayBuffer` into `Buffer` once more is a good idea. Buffer representation of a string body should give a correct length.

marcosvega91: fine, I'm not sure about only one thing the order of the triggered events. Following the specs it seams that `loadstart` should be triggered before `onpregress`

https://xhr.spec.whatwg.org/#event-xhr-loadstart

kettanaito: You're absolutely right. We can re-order the events to match the spec in a separate pull request if you don't mind.

kettanaito:
XMLHttpRequest events re-ordering issue created: https://github.com/mswjs/node-request-interceptor/issues/76

marcosvega91 (approved):
this is good for me 😄  🥳

PR #77, opened 2020-12-08
Fix two memory leaks

alexk111:
1st leak comes from Debug module, when debug instances are created within a function. For more info: https://github.com/visionmedia/debug/issues/678

2nd leak is caused by `createClientRequestOverrideClass`, when `handleRequest` is called multiple times (check screenshot below). It inherits `ClientRequestOverride` from a previous `ClientRequestOverride` instance:
- `http.ClientRequest = ClientRequestOverride`
https://github.com/mswjs/node-request-interceptor/blob/728822439966cd06e67948e3bea370377d098177/src/interceptors/ClientRequest/index.ts#L33-L41
- next calls wrap `ClientRequestOverride` which wraps `ClientRequestOverride` which wraps `ClientRequestOverride` etc.
https://github.com/mswjs/node-request-interceptor/blob/728822439966cd06e67948e3bea370377d098177/src/interceptors/ClientRequest/ClientRequestOverride.ts#L332

Maybe I'm missing something and you intentionally inherited it from `http.ClientRequest` instead of `originalClientRequest`?

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   },
   "dependencies": {
     "@open-draft/until": "^1.0.3",
-    "debug": "^4.1.1",
+    "debug": "^4.3.0",
     "headers-utils": "^1.2.0"
   },
   "keywords": [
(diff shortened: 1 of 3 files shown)


shanecurran (approved):
Good catch 👍

kettanaito (approved):
This is a great improvement! Thank you, @alexk111, for investigating this and shipping a fix. You're awesome!

PR #81, opened 2020-12-18
add static copies of readyState values

### LINKED ISSUE #78: XMLHttpRequest.DONE is undefined after setupServer
In my Jest test:

```js
import {setupServer} from 'msw/node';

console.log(XMLHttpRequest.DONE);
// logs: 4

setupServer();

console.log(XMLHttpRequest.DONE);
// logs: undefined
```

I am using:

- node-request-interceptor@0.5.9
- msw@0.24.2
- jest@26.6.3
- jsdom@16.4.0

cainlevy:
TypeScript's `interface XMLHttpRequest` in lib.dom.d.ts requires these properties to exist as static and instance properties. I've added the static properties here, but I'm unsure how to enforce them against the type.

https://github.com/microsoft/TypeScript/blob/2428ade1a91248e847f3e1561e31a9426650efee/lib/lib.dom.d.ts#L18806-L18810

https://github.com/microsoft/TypeScript/blob/2428ade1a91248e847f3e1561e31a9426650efee/lib/lib.dom.d.ts#L18817-L18825

Fixes #78

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -56,6 +56,11 @@ export const createXMLHttpRequestOverride = (
     _events: XMLHttpRequestEvent<XMLHttpRequestEventTargetEventMap>[] = []
 
     /* Request state */
+    public static readonly UNSENT = 0
+    public static readonly OPENED = 1
+    public static readonly HEADERS_RECEIVED = 2
+    public static readonly LOADING = 3
+    public static readonly DONE = 4
     public readonly UNSENT = 0
     public readonly OPENED = 1
     public readonly HEADERS_RECEIVED = 2
--- a/test/compliance/XMLHttpRequest.test.ts
+++ b/test/compliance/XMLHttpRequest.test.ts
@@ -1,0 +1,27 @@
+import { RequestInterceptor } from '../../src'
+import { interceptXMLHttpRequest } from '../../src/interceptors/XMLHttpRequest'
+
+let interceptor: RequestInterceptor
+
+beforeAll(() => {
+  interceptor = new RequestInterceptor([interceptXMLHttpRequest])
+})
+
+afterAll(() => {
+  interceptor.restore()
+})
+
+test('exposes ready state enums both as static and public properties', () => {
+  expect(XMLHttpRequest.UNSENT).toBe(0)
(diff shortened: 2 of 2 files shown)


src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -38,6 +38,11 @@ export const createXMLHttpRequestOverride = (
...
 
     /* Request state */
+    public static readonly UNSENT = 0
+    public static readonly OPENED = 1
+    public static readonly HEADERS_RECEIVED = 2
+    public static readonly LOADING = 3
+    public static readonly DONE = 4
     public readonly UNSENT = 0
     public readonly OPENED = 1
     public readonly HEADERS_RECEIVED = 2
```
kettanaito, line 48: We cannot mark the existing properties as static, because states should be available on both class instance and as a class static property, right?

```js
XMLHttpRequest.DONE
req.DONE
```

kettanaito: Yeap, that seems to be the case. Awesome!

test/intercept/XMLHttpRequest/XMLHttpRequest.test.ts
```diff
@@ -281,3 +281,18 @@ test('intercepts an HTTPS HEAD request', async () => {
     'x-custom-header': 'yes',
   })
 })
+
+test('ready state names', () => {
```
kettanaito, line 285: Do you mind moving this test suite into a new folder called `test/compliance`? This one doesn't directly concern request interception. Thanks.

kettanaito (changes requested):
This looks great! I only had one nitpick about placing the new test into a new test directory

```
test/
  compliance/
    XMLHttpRequest.test.ts
```

cainlevy:
Sure thing!

I've force-pushed my commit to keep the change history simple.

kettanaito:
Hey, @cainlevy. I've rebased your feature branch and added the request interceptor to the test. Without it, we are testing the `XMLHttpRequest` instance that `jsdom` exposes, which is promised to be compatible. With `new RequestInterceptor()` in place, that instance gets monkey patched, and it's the test's point to assert that patch's compatibility.

kettanaito (approved):
Thank you for spotting and fixing this, @cainlevy 🎉 Welcome to contributors!

PR #83, opened 2020-12-18
fix: propagate headers when making original XMLHttpRequest

marcosvega91:
## Changes

- When the request is not mocked, the original request is made. This PR will fix an issue where client headers were not forwarded to the server

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -380,6 +380,7 @@ export const createXMLHttpRequestOverride = (
             // to the original XHR instance.
             this.propagateCallbacks(originalRequest)
             this.propagateListeners(originalRequest)
+            this.propagateHeaders(originalRequest, requestHeaders)
 
             if (this.async) {
               originalRequest.timeout = this.timeout
@@ -530,6 +531,18 @@ export const createXMLHttpRequestOverride = (
     propagateListeners(req: XMLHttpRequest) {
       this._events.forEach(({ name, listener }) => {
         req.addEventListener(name, listener)
+      })
+    }
+
+    propagateHeaders(
+      req: XMLHttpRequest,
+      headers: Record<string, string | string[]>
+    ) {
+      Object.entries(headers).forEach(([key, value]) => {
+        req.setRequestHeader(
+          key,
+          Array.isArray(value) ? value.join(',') : value
+        )
       })
     }
   }
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -358,6 +358,9 @@ export const createXMLHttpRequestOverride = (
               this.trigger('load')
 
               const responseHeaders = originalRequest.getAllResponseHeaders()
+              this.responseHeaders = flattenHeadersObject(
+                headersToObject(stringToHeaders(responseHeaders))
+              )
               debug('original response headers', responseHeaders)
 
               const normalizedResponseHeaders = headersToObject(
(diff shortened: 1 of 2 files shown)


test/regressions/xhr-propagate-headers.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  interceptor.restore()
+})
+
+test('forward the request headers to the server', async () => {
+  await createXMLHttpRequest((req) => {
+    req.open('GET', server.makeHttpUrl('/account'))
+    req.setRequestHeader('x-client-header', 'yes')
+    req.setRequestHeader('x-multi-values', 'value1, value2')
+    req.send()
+    req.addEventListener('loadend', function () {
```
kettanaito, line 39: The `loadend` event is what the `createXMLHttpRequest` Promise is based on. Consider this instead:

```js
const req = await createXMLHttpRequest((req) => {
  // Create a request
  // Append headers
  // Can omit `req.send()`, it's done by default
})

const headers = req.getAllResponseHeaders()
// Continue the assertions
```

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -532,5 +533,17 @@ export const createXMLHttpRequestOverride = (
...
     }
+
+    propagateHeaders(
+      req: XMLHttpRequest,
+      headers: Record<string, string | string[]>
+    ) {
+      Object.entries(headers).forEach(([key, value]) => {
+        req.setRequestHeader(
+          key,
+          Array.isArray(value) ? value.join(',') : value
```
kettanaito, line 544: Could you please use `flattenHeadersObject` from [`headers-utils`](https://github.com/mswjs/headers-utils) for this?

```js
const flatHeaders = flattenHeadersObject(headers)
Object.entries(flatHeaders).forEach(...)
```

kettanaito (changes requested):
This is great! Please, could you look at a few comments I've left? Thanks.

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -358,6 +358,9 @@ export const createXMLHttpRequestOverride = (
               this.trigger('load')
 
               const responseHeaders = originalRequest.getAllResponseHeaders()
+              this.responseHeaders = flattenHeadersObject(
+                headersToObject(stringToHeaders(responseHeaders))
+              )
```
marcosvega91, line 363: I see that we have forgotten to return back the headers

kettanaito: Is it necessary? We keep an internal `responseHeaders` property to get/set response headers. That's also the purpose of the same internal property on the original XHR (the one I used to get via a Symbol).

If you take a look at `[XMLHttpRequest properties](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#Properties) you can see that `responseHeaders` isn't one of them. Developers should retrieve response headers using the `getAllResponseHeaders` or `getResponseHeader` methods of the XHR instance.

marcosvega91: Yes , but how the user will retrieve headers if the requests are different? 

The user creates an instance of XMLHttpRequest that is our custom class. Our class then call the real XMLHttpRequest. The real XMLHttpRequest has headers but our instance not

kettanaito: 🤔 That makes sense. Sorry I've overlooked that.

kettanaito: Did your new test fail without this assignment? I wonder if we're covering this behavior.

marcosvega91: Yes, accessing the header from `req` instead of the event function will cover that case

kettanaito (approved):
This is magnificent. Thank you for discovering and fixing this, @marcosvega91 🎉

PR #87, opened 2021-01-13
ClientRequest: Clones request options to prevent mutations

### LINKED ISSUE #86: Getting "TypeError [ERR_INVALID_ARG_TYPE]: The "timeout" argument must be of type number. Received an instance of Object" when using "got"
When using msw on node for requests created by [got](https://www.npmjs.com/package/got) I'll get the following error.

```xxx:error TypeError [ERR_INVALID_ARG_TYPE]: The "timeout" argument must be of type number. Received an instance of Object
    at validateNumber (internal/validators.js:129:11)
    at getTimerDuration (internal/timers.js:381:3)
    at new ClientRequest (_http_client.js:170:20)
    at request (http.js:50:10)
    at Object.proxiedOriginalRequest (./node_modules/node-request-interceptor/src/interceptors/ClientRequest/index.ts:67:29)
    at ClientRequestOverride.<anonymous> (./node_modules/node-request-interceptor/src/interceptors/ClientRequest/ClientRequestOverride.ts:268:15)
    at step (./node_modules/node-request-interceptor/lib/interceptors/ClientRequest/ClientRequestOverride.js:33:23)
    at Object.next (./node_modules/node-request-interceptor/lib/interceptors/ClientRequest/ClientRequestOverride.js:14:53)
    at fulfilled (./node_modules/node-request-interceptor/lib/interceptors/ClientRequest/ClientRequestOverride.js:5:58)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```
The initial requested created by me doesn't even contain a timeout value.
I've created a minimal repo to showcase the error: https://github.com/KnisterPeter/node-request-interceptor-with-got-issue

kettanaito:
- Fixes #86


KnisterPeter:
@kettanaito Do you need something from me to get this merged?

kettanaito:
Hey, @KnisterPeter. Unfortunately, even with this change, the issue is not resolved.

This pull request does fix the mutation of `ClientRequest` options and the got's `this.timeout` doesn't leak to NRI. But there's still an exception because at some point `got` tries to perform a request to a `:80` port of the host and not the specified port. I've described it [here](https://github.com/sindresorhus/got/issues/1579#issuecomment-759301241), the issue has been re-opened. 

While that's probably an issue on the `got` side, there's always a chance that NRI can ship an improvement to prevent that behavior from happening.

KnisterPeter:
Thanks for explaining. Would be nice to at least see this issue fixed.

kettanaito:
I don't mind merging this, but the issue won't be solved by this pull request alone. That's why I'd prefer not to merge half-fixes. I hope for your understanding on that.

KnisterPeter:
Sure, do as you like. Thanks anyway :smiley:

szmarczak:
I put `console.log(url)` after https://github.com/mswjs/node-request-interceptor/blob/64c42229cf4d68fe58809e4679bce3592939d9c7/src/interceptors/ClientRequest/ClientRequestOverride.ts#L35 and the URL is `http://127.0.0.1:[random_port]/user` so it must be something wrong on your side.

szmarczak:
I found the culprit: https://github.com/mswjs/node-request-interceptor/blob/64c42229cf4d68fe58809e4679bce3592939d9c7/src/interceptors/ClientRequest/ClientRequestOverride.ts#L282

where is the `url` passed? The native `http` doesn't support `options.url`.

szmarczak:
`req = performOriginalRequest((options as any).url, options)` and the test passes

kettanaito:
Thank you so much for the investigation, @szmarczak ❤️ You're right, we didn't pass the correct url to the original `http` request.

I've set it to propagate the `url` parsed from the ClientRequest parameters:

```ts
req = performOriginalRequest(url.toString(), options)
```

That way we don't have to typecast the url to `any`. Outstanding investigation powers.

msutkowski (approved):
Looks good!

szmarczak:
I'm glad I could help ♥️

kettanaito:
You are incredibly helpful, @szmarczak. Thank you!

PR #89, opened 2021-02-04
XMLHttpRequest `getResponseHeader` case sensitivity

kalopilato:
I came across a problem when using MSW in testing: the component under test has a dependency that makes API calls (authentication) and inspects the "Content-Type" response header (note the casing), but the value of this header was always null. I traced this to the XMLHttpRequestOverride's `getResponseHeader` function which evaluated "Content-Type" (the passed in header name) against "content-type" (the actual header name in the response), thereby failing and returning null.

This PR updates the XHR `getResponseHeader` function to normalise the passed in header name and the response headers to lowercase when evaluating equality. This looks to be similar behaviour to [jsdom's implementation](https://github.com/jsdom/jsdom/blob/5279cfda5fe4d52f04b2eb6a801c98d81f9b55da/lib/jsdom/living/xhr/xhr-utils.js#L32) and aligns with HTTP spec.

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -422,13 +422,7 @@ export const createXMLHttpRequestOverride = (
       const headerValue = Object.entries(this.responseHeaders).reduce<
         string | null
       >((_, [headerName, headerValue]) => {
-        // Ignore header name casing while still allowing to set response headers
-        // with an arbitrary casing (no normalization).
-        if ([headerName, headerName.toLowerCase()].includes(name)) {
-          return headerValue
-        }
-
-        return null
+        return (headerName.toLowerCase() === name.toLowerCase()) ? headerValue : null
       }, null)
 
       debug('resolved response header', name, headerValue, this.responseHeaders)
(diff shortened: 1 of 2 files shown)


src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -422,13 +422,7 @@ export const createXMLHttpRequestOverride = (
       const headerValue = Object.entries(this.responseHeaders).reduce<
         string | null
       >((_, [headerName, headerValue]) => {
-        // Ignore header name casing while still allowing to set response headers
-        // with an arbitrary casing (no normalization).
```
kalopilato, line 426: This gives me pause - it looks like it was a conscious decision _not_ to normalize case so I might be missing something here?

kettanaito: I think you're correct: header names must be case-insensitive when retrieving a header. We forgot to normalize the input `name` to provision that.

kettanaito (approved):
Thank you for spotting and fixing this! Tremendous appreciation for adding a test covering the regression ❤️ 

Welcome to contributors, @kalopilato! 🎉

kettanaito:
@kalopilato could you please set the right email for your commits? They don't seem to be recognized as yours on GitHub. This often happens if your email in Git doesn't match the email on GitHub. You deserve your contribution to be associated with you.

kalopilato:
Thanks @kettanaito!! And thanks for the heads up about my git config, have updated and amended those commits. Pleased to have finally made my first open source contribution 😄

kettanaito:
We are happy to see you contributing, especially with this superb finding. Thank you!

PR #91, opened 2021-02-04
correct the lib file resolve path in README

aprilandjan:
This PR corrects the lib file resolve path in README, to avoid confusions.

see https://github.com/mswjs/node-request-interceptor/issues/68#issuecomment-772552694


kettanaito (approved):
Thank you for fixing this, @aprilandjan! Welcome to contributors 🎉

PR #93, opened 2021-03-05
Adds "fetch" browser interceptor

kettanaito:
## Changes

- Sets the CircleCI Docker image to `circleci/node:12.18.0-browsers` to include Chrome used for automated browser tests with `page-with`. 
- The library now supports a client-side interceptor for `window.fetch`.

```js
import { createInterceptor } from 'node-request-interceptor'
import { interceptFetch } from 'node-request-interceptor/lib/interceptors/fetch'

createInterceptor({
  modules: [interceptFetch],
  resolver(request) {}
})
```

- Uses `page-with` for in-browser testing.
- Adds dedicated `yarn test:integration:browser` command for running in-browser tests.
- Adds dedicated `yarn test:integration:node` command for running Node.js-oriented tests.

## Motivation

This change extends the responsibilities of NRI to also patch browser-specific modules. It's the first step towards NRI being a generic request interception library for any environment. 

Fundamentally, this is a pre-requisite for the "Fallback mode" in MSW: an interception for environments when Service Worker can't otherwise function (insecture localhost, statically built app). 

## Roadmap

- [x] Provide the request `ref` to the resolver function for `fetch` intercepted request. 
- [x] Handle non-string request bodies: Blob, ArrayBuffer, ReadableStream.


kettanaito:
Needs a new release of `page-with` to bundle `webpack` as a dependency. Including `webpack` as a peer dependency causes all sorts of type violations even when webpack isn't even imported in the library's code.

src/interceptors/fetch/index.ts
```diff
@@ -0,0 +1,90 @@
...
+    const url = typeof input === 'string' ? input : input.url
+    const method = init?.method || 'GET'
+
+    debug('[%s] %s', method, url)
+
+    const request: IsomoprhicRequest = {
+      url: new URL(url, location.origin),
+      method: method,
+      headers: init?.headers ? headersToObject(new Headers(init.headers)) : {},
+      body: await ref.text(),
```
kettanaito, line 32: This part is extremely frustrating to test as non-string values must be serialized when passing from Playwrigth (context where the request has been intercepted) and the test runner (Jest, in Node.js). 

@marcosvega91, do you think this kind of body conversion will work for request bodies provided as `Blob`/`ArrayBuffer`/`FormData`/`ReadableStream`/etc.?

marcosvega91: `ReadableStream` is available only in the response. I think that in this case there will be no problem

src/interceptors/fetch/index.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      observer.emit('response', request, isomorphicResponse)
+
+      return new Response(response.body, {
+        ...isomorphicResponse,
+        // `Response.headers` cannot be instantiated with the `Headers` polyfill.
+        // Apparently, it halts if the `Headers` class contains unknown properties
+        // (i.e. the internal `Headers.map`).
+        headers: flattenHeadersObject(response.headers || {}),
+      })
```
marcosvega91, line 53: Do you think that here we can add a try catch? `new Response` could throw some exceptions for example using a body != null with a status of 204. Maybe we can add a graceful message in this case? 🤔

test/intercept/fetch/fetch.body.browser.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+test('handles requests with no body', async () => {
+  const context = await prepareRuntime()
+  await context.page.evaluate(() => window.requestWithEmptyBody())
+})
+
+test('handles requests with a Blob body', async () => {
+  const context = await prepareRuntime()
+  await context.page.evaluate(() => window.requestWithBlob())
+})
```
marcosvega91, line 28: As said on discord this part is not really clear and seems that test does nothing :(. Do you think there is a way to move the code from the mock to the test file ?

marcosvega91 (changes requested):

test/intercept/fetch/fetch.body.browser.test.ts
kettanaito: Thank you for the suggestion! Assertions have been moved to the test file.

src/interceptors/fetch/index.ts
kettanaito: If a new Response instance is given an impossible input, then the exception should surface. That means that the user has provided that impossible input in the mock:

```js
createInterceptor({
  modules: [interceptFetch],
  resolver() {
    return {
      status: 204,
      body: 'hey, i should not be here!'
    }
  }
})
```

marcosvega91 (approved):
LGTM 💪

PR #94, opened 2021-03-08
ClientRequest: Infers protocol from the request module (defaultProtocol) if not provided

### LINKED ISSUE #67: Http protocol resolved for https module when no protocol information is provided.
A request in the style of `require('https').request({host: 'www.banana.com', path: '/cherry'})` should be normalised to an https request, not http. I have created a pull request with a failing test and more information, so that the discussion may be continue there.

https://github.com/mswjs/node-request-interceptor/pull/66

kettanaito:
## GitHub

- Fixes https://github.com/mswjs/msw/issues/632
- Fixes #67 
- Closes #66

## Changes

- `normalizeHttpRequestParams` function now accepts the `defaultProtocol: string` argument that represents the default protocol associated with the request module (`http`/`https`). 
- **Only if the `protocol` isn't otherwise provided** (i.e. in `options.protocol`, or `URL.protocol`), request params normalization infers the `defaultProtocol`.

### DIFF
--- a/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
@@ -36,6 +36,7 @@ function resolveCallback(
  * so it always has a `URL` and `RequestOptions`.
  */
 export function normalizeHttpRequestParams(
+  defaultProtocol: string,
   ...args: HttpRequestArgs
 ): [URL, RequestOptions & RequestSelf, HttpRequestCallback?] {
   let url: URL
@@ -43,6 +44,7 @@ export function normalizeHttpRequestParams(
   let callback: HttpRequestCallback | undefined
 
   debug('arguments', args)
+  debug('default protocol', defaultProtocol)
 
   // Convert a url string into a URL instance
   // and derive request options from it.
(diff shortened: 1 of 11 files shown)


src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
```diff
@@ -98,17 +102,27 @@ export function normalizeHttpRequestParams(
...
   }
   // Handle a given RequestOptions object as-is
   // and derive the URL instance from it.
   else if (isObject(args[0])) {
     options = args[0]
     debug('given request options:', options)
 
+    // When handling a `RequestOptions` object without an explicit "protocol",
+    // infer the protocol from the request issuing module (http/https).
+    options.protocol = options.protocol || defaultProtocol
```
kettanaito, line 123: Protocol coercion is applicable only when passing `RequestOptions` object as the argument to `http.ClientRequest`. In that case, fallback to the default protocol inferred from the request module if no explicit protocol was provided.

> Some libraries (i.e. `superagent`) do not provide any protocol explicitly. They do issue a request using the right request module, however (http/https).

marcosvega91 (approved):
It's super nice ! :)

timdeschryver (approved):

sean-hernon:
I fixed this 5 months ago in another PR. 😂 

Now I can finally go back to using the upstream version, instead of keeping my custom one up-to-date. 👍

kettanaito:
@sean-hernon, thank you for your fix as well, it was useful as a reference. The final one is different a little, you can compare the diff if you're interested in the details. 

Not sure when this will be published, as we also need for #93 to land.

PR #95, opened 2021-03-08
fix: set readyState to 4 (DONE) on bypassed response

marcosvega91:
As for the mocked request also bypass request should have the readyState set to 4 (DONE) when a response is received

Fixes https://github.com/mswjs/msw/issues/636


marcosvega91:
it should be ok now

PR #100, opened 2021-03-16
Supports XMLHttpRequest interception in a browser

kettanaito:
## GitHub

- Follow-up to #93 

## Changes

- Adds in-browser tests for XMLHttpRequest interception and response mocking.
- Uses the env-agnostic `bufferFrom` utility instead of `Buffer.from()` that's available only in Node.js.
- Adds tests for the `createEvent` function.

We don't need to use the native `window.Event` because it forbids setting the `target`, which in the case of `XMLHttpRequestOverride` should be the overridden class. That's one of the reasons we have the `EventPolyfill` because it allows setting a custom event target in the class constructor.

### DIFF
--- a/src/interceptors/XMLHttpRequest/utils/bufferFrom.ts
+++ b/src/interceptors/XMLHttpRequest/utils/bufferFrom.ts
@@ -1,0 +1,16 @@
+/**
+ * Convert a given string into a `Uint8Array`.
+ * We don't use `TextEncoder` because it's unavailable in some environments.
+ */
+export function bufferFrom(init: string): Uint8Array {
+  const encodedString = encodeURIComponent(init)
+  const binaryString = encodedString.replace(/%([0-9A-F]{2})/g, (_, char) => {
+    return String.fromCharCode(('0x' + char) as any)
+  })
+  const buffer = new Uint8Array(binaryString.length)
+  Array.prototype.forEach.call(binaryString, (char, index) => {
+    buffer[index] = char.charCodeAt(0)
+  })
+
+  return buffer
+}
--- a/test/response/xhr.browser.test.ts
+++ b/test/response/xhr.browser.test.ts
@@ -1,0 +1,104 @@
+import * as path from 'path'
+import { pageWith } from 'page-with'
+import { createServer, ServerApi } from '@open-draft/test-server'
+import { createBrowserXMLHttpRequest } from '../helpers'
+import { InterceptorApi } from '../../src'
+
+declare namespace window {
+  export const interceptor: InterceptorApi
+  export let serverHttpUrl: string
+  export let serverHttpsUrl: string
+}
+
+let server: ServerApi
+
+async function prepareRuntime() {
+  const scenario = await pageWith({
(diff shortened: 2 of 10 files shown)


src/interceptors/XMLHttpRequest/utils/bufferFrom.ts
```diff
@@ -0,0 +1,12 @@
...
+  const binstr = escstr.replace(/%([0-9A-F]{2})/g, function (_, char) {
+    return String.fromCharCode(('0x' + char) as any)
+  })
+  const buffer = new Uint8Array(binstr.length)
+  Array.prototype.forEach.call(binstr, (char, index) => {
+    buffer[index] = char.charCodeAt(0)
+  })
+
+  return buffer
+}
```
msutkowski, line 12: This is probably a naive question - but does this need to cover any additional cases or be more defensive? I was researching `Buffer.from` ponyfills, and they all handle things quite a bit differently. For example: is it possible for a response body to be a number such as `13` and have a response header type that is erroneously set to `arraybuffer`? Would that break here? I get that it's most likely not the case and is probably an error on the developer's side, but I was curious as to if that should be a concern or not? It seems like we might want to throw TypeErrors or something somewhere if it's misconfigured, but not 100% sure. 

Example: https://github.com/feross/buffer/blob/master/index.js#L166

test/response/xhr.browser.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  const scenario = await prepareRuntime()
+  const request = createBrowserXMLHttpRequest(scenario)
+  const response = await request('GET', server.http.makeUrl('/get'))
+
+  expect(response).toHaveProperty('status', 200)
+  expect(response).toHaveProperty('statusText', 'OK')
+  expect(response).toHaveProperty('body', JSON.stringify({ route: '/get' }))
+})
+
+test('bypasses any request when the interceptor is resotred', async () => {
```
msutkowski, line 84:
```suggestion
test('bypasses any request when the interceptor is restored', async () => {
```

msutkowski (commented):
I think this looks great overall! I just have a few clarifying questions.

src/interceptors/XMLHttpRequest/utils/bufferFrom.ts
kettanaito: This is a good point. The response body is always retrieved as a string since it's always set as a string (even if you pass a number). I don't think that would cause a problem when transforming `"13"` into a buffer. 

This is a rather dark area, and I think our users will tell us if this implementation works as expected. Note that it's also not specific to this change: we converted the response body to `Buffer` previously, here I'm replacing the native `Buffer` class with this utility for in-browser usage. As always with NRI, I expect regressions and we should cover them with tests.

kettanaito:
@msutkowski, thank you for the review, Matt! I've addressed your comments and elaborated on the Buffer changes made here. Hope it's better now.

msutkowski (approved):
Thanks for the clarification!

PR #101, opened 2021-03-18
XMLHttpRequest: Uses "Headers" polyfill for internal request/response headers

kettanaito:
## GitHub

Preparing for the next release.

## Changes

- Updates dependencies (`headers-utils@3.0.0`).
- XHR: Uses `Headers` polyfill class to store internal `requestHeaders`/`responseHeaders`.
  - Fewer transformations needed to work with headers (set request headers, retrieve response headers). According to the spec, header names must be case-sensitive, so this shouldn't be an issue. 
- `createEvent` test: no longer relies on the internal `XMLHttpRequest` `flag` property stored behind a Symbol. That property is removed in the latest version of Node/Jest.
- Fix: calling `req.getResponseHeader(name)` resulted into `null` if there were multiple matching headers. Lookup in the reduce didn't handle the case when the header value has been found in the previous iteration.


kettanaito:
Hey, @marcosvega91 👋 Please, could you help me reviewing this pull request? I wonder what's your take on that headers change is going to be. Thank you!

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -422,15 +414,14 @@ export const createXMLHttpRequestOverride = (
         return null
       }
 
-      const headerValue = Object.entries(this.responseHeaders).reduce<
-        string | null
-      >((_, [headerName, headerValue]) => {
-        return headerName.toLowerCase() === name.toLowerCase()
```
kettanaito, line 428: This actually fixes an issue: if there are multiple headers and the first one is the one you're getting, the next header in the iteration will reset the accumulated value to `null`.

marcosvega91 (approved):
this is nice, and as I see headers are a super pain hahahah

kettanaito:
Thank you for the review, @marcosvega91! I'll add a new compliance test for XHR request headers to make sure the casing change above doesn't break the compatibility. Looks like we'd have to maintain the original casing of the header while allow retrieving the headers in a case-insensitive manner. That'd be the `headers-utils` change.

PR #102, opened 2021-03-18
fix(XMLHttpRequest): reorder trigger "loadstart" upon mocked response…

### LINKED ISSUE #76: XMLHttpRequest: Ensure events order according to the specification
- [Specification](https://xhr.spec.whatwg.org/#event-xhr-loadstart)

## Current behavior

Some of the events are grouped together, not representing the valid flow of an XHR request:

https://github.com/mswjs/node-request-interceptor/blob/50e5aa72ee102a8bec8e2f38e3efc9efa26d5e74/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L288-L290

## Expected behavior

The `XMLHttpRequest` events must be dispatched in the order described in the specification to guarantee compatibility for the users.

alfdocimo:
## GitHub

As stated in #76, some events are grouped together _(loadstart, load, loadend)_ which is not a valid flow of an XHR request. This small PR reorders the events to the specification to guarantee compatibility for the users as @kettanaito states.

## Changes: 

- Moves up the loadstart event so that it's dispatched as soon it handles a mocked response.
Based on the [Events summary](https://xhr.spec.whatwg.org/#event-xhr-loadstart)
- Fixes #76

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -263,6 +263,9 @@ export const createXMLHttpRequestOverride = (
           if (mockedResponse) {
             debug('received mocked response', mockedResponse)
 
+            // Trigger a loadstart event to indicate the initialization of the fetch
+            this.trigger('loadstart')
+            
             this.status = mockedResponse.status || 200
             this.statusText = mockedResponse.statusText || 'OK'
             this._responseHeaders = mockedResponse.headers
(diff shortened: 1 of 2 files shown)


kettanaito:
Hey, @alfdocimo. Thank you for the pull request! 

I've rebased it against the latest `master` and added a few changes:

- Integration (compliance) test that the XMLHttpRequest events are executed in the correct order and with the correct `readyState` value depending on the event phase.
- Abstracted the `readyState` value changes into the internal `this.setReadtState()` function.
- Removed the `readystatechange` event trigger from the `this.trigger()` method. The `readystatechange` event is triggered _only when the value of `this.readyState` changes_ (see [ref](https://xhr.spec.whatwg.org/#event-xhr-readystatechange)).
- Added a bunch of logs to the related events so they can be inspected via `DEBUG=XHR*`.

kettanaito (approved):
Superb work! Thank you and welcome to contributors, @alfdocimo 🎉

PR #104, opened 2021-03-22
Renames the package to "@mswjs/interceptors"

kettanaito:
## Motivation

This package is no longer Node.js-specific after the introduction of the `fetch` interception (#93). The "node" part in the name may be confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,31 +1,27 @@
-[![Latest version](https://img.shields.io/npm/v/node-request-interceptor.svg)](https://www.npmjs.com/package/node-request-interceptor)
-[![Build status](https://img.shields.io/circleci/project/github/mswjs/node-request-interceptor/master.svg)](https://app.circleci.com/pipelines/github/mswjs/node-request-interceptor)
+[![Latest version](https://img.shields.io/npm/v/@mswjs/interceptors.svg)](https://www.npmjs.com/package/@mswjs/interceptors)
+[![Build status](https://img.shields.io/circleci/project/github/mswjs/interceptors/master.svg)](https://app.circleci.com/pipelines/github/mswjs/interceptors)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,31 +1,27 @@
...
 
 ### How is this library different?
 
-As interception is often combined with request route matching, some libraries can determine whether a request should be mocked _before_ it actually happens. This approach is not suitable for this library, as it rather _intercepts all requests_ and then let's you decide which ones should be mocked. This affects the level at which interception happens, and also the way mocked/original response is constructed, in comparison to other solutions.
-
-### Why XMLHttpRequest?
-
-Although NodeJS has no `XMLHttpRequest` implementation, this library covers it for the sake of processes that still run in NodeJS, but emulate a browser-like environment (i.e. `jsdom` when running tests in Jest).
+As interception is often combined with request route matching, some libraries can determine whether a request should be mocked _before_ it actually happens. This approach is not suitable for this library, as it rather _intercepts all requests_ and then let's you decide which ones should be mocked. This affects the level at which interception happens, and also the way mocked/original responses is constructed, in comparison to other solutions.
```
msutkowski, line 24:
```suggestion
As interception is often combined with request route matching, some libraries can determine whether a request should be mocked _before_ it actually happens. This approach is not suitable for this library, as it rather _intercepts all requests_ and then let's you decide which ones should be mocked. This affects the level at which interception happens, and also the way mocked/original responses are constructed, in comparison to other solutions.
```

msutkowski (commented):
Looks great! Just a couple of grammar suggestions.

kettanaito:
Thank you for the suggestions, @msutkowski! Addressed.

msutkowski (approved):
LGTM :+1: :trophy:

PR #105, opened 2021-03-23
Manages request headers as the HeadersPolyfill instance

kettanaito:
## Change

- `request.headers` (isomorphic request headers) is now managed as the `HeadersPolyfill` instance.

## Motivation

`Headers` class is much more versatile than managing headers as an object. I see no reason why we shouldn't consistently manage request/response headers using the polyfill class.


src/interceptors/ClientRequest/createClientRequestOverride.ts
```diff
@@ -156,28 +152,31 @@ export function createClientRequestOverride(
...
-            {}
-          )
-        : {}
+      const requesHeadersObject = Object.entries(
+        resolvedRequestHeaders
+      ).reduce<HeadersObject>((headersObject, [name, value]) => {
+        if (value) {
+          const corcedValue =
+            typeof value === 'number' ? value.toString() : value
+          headersObject[name.toLowerCase()] = corcedValue
```
kettanaito, line 168: Without this header name normalization it creates an object containing multiple same headers with different name casing:

```js
{
  accept: '*/*',
  Accept: '*/*'
}
```

When constructing the `Headers` instance based on such input, it results in headers values' being merged despite them being the same. Raised in https://github.com/mswjs/headers-utils/issues/13. Not an issue, given we preserve the header name normalization here.

msutkowski: Nice. Perhaps in a future we just renamed `corcedValue` to `coercedValue`, but not a big deal :)

msutkowski (commented):
Nice readability improvements :trophy:

kettanaito:
@msutkowski well, I just keep on giving you some proofreading exercises 🤦 Sorry.

PR #107, opened 2021-03-27
Update aliases from nri -> @mswjs/interceptors

msutkowski:
Updates aliases and GitHub issues to be representative of the new package name

### DIFF
--- a/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
@@ -26,7 +26,7 @@ function resolveRequestOptions(
   /**
    * Clone the request options to lock their state
    * at the moment they are provided to `ClientRequest.
-   * @see https://github.com/mswjs/node-request-interceptor/issues/86
+   * @see https://github.com/mswjs/interceptors/issues/86
    */
   return args[1] ? cloneObject(args[1]) : ({} as RequestOptions)
 }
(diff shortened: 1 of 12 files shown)


src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
```diff
@@ -26,7 +26,7 @@ function resolveRequestOptions(
   /**
    * Clone the request options to lock their state
    * at the moment they are provided to `ClientRequest.
-   * @see https://github.com/mswjs/node-request-interceptor/issues/86
+   * @see https://github.com/@mswjs/interceptors/issues/86
```
kettanaito, line 29: You don't have to include `@` in the organization name. This should be `mswjs/interceptors` instead.

kettanaito (changes requested):
Thanks for addressing this, @msutkowski! 
If I may, just one point regarding the GitHub links.

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
msutkowski: Bad find and replace on my part 🤦‍♂️. Will update.

msutkowski:
> Thanks for addressing this, @msutkowski!
> If I may, just one point regarding the GitHub links.

All set, and confirmed the links work. Sorry about that 🤦 .

kettanaito (approved):
Gorgeous ✨ Thank you, Matt!

Issue #109, opened 2021-04-02
node interceptors corrupt incoming binary bodies


FauxFaux:
https://github.com/mswjs/interceptors/blob/c6f440a2b1abb3b230094b8c794a0dc35d2f1981/src/interceptors/ClientRequest/utils/getIncomingMessageBody.ts#L8

`chunk` here is normally(?) a `Buffer`, and `+=` is silently calling `.toString(encoding='utf-8')` on it, which is lossy. This prevents the interception of transfer encoded bodies, as this corrupts `gzip` data.

thomas-p-wilson:
Can confirm that

```
export function getIncomingMessageBody(res: IncomingMessage): Promise<string> {
  let responseBody = Buffer.from('');

  return new Promise((resolve, reject) => {
    res.once('error', reject)
    res.on('data', (chunk) => { responseBody = Buffer.concat([responseBody, chunk]); })
    res.once('end', () => {
      resolve(responseBody)
    })
  })
}
```

resolves my particular problem with gzipped response contents. I expect this would also resolve OP's problem. I think it's safe to say that the above does not represent an acceptable solution due to the likely overwhelming reliance on the response body being a string.

Some general thoughts:
- `let responseBody = Buffer.from('');` should be provided a fixed size based on the `Content-Length` header if at all possible, whereby we could avoid concatenating buffers needlessly. This falls over if Content-Length is unavailable or incorrect.
- Could we default to returning a string, but provide the developer an option to receive the buffer instead? Not sure if there's the possibility of supplying such an option.

Alternately, we could handle the gzip compression ourselves. To me this seems to get into the weeds a bit too much, and risks broadening the scope of the library to handle other forms of compression, inflate/deflate, and other such things. But for the sake of the conversation, the following would potentially be an acceptable solution to this particular problem as it continues to return a string as the response body, as would be expected by existing users of this library:

```
import { createGunzip } from 'zlib';

export function getIncomingMessageBody(res: IncomingMessage): Promise<string> {
  let responseBody = '';
  let stream = res;
  if (res.headers['content-encoding'] === 'gzip') {
    const gzip = createGunzip();
    stream = res.pipe(gzip);
  }

  return new Promise((resolve, reject) => {
    stream.once('error', reject)
    stream.on('data', (chunk) => { responseBody += chunk); })
    stream.once('end', () => {
      resolve(responseBody)
    })
  })
}
```

[...]

PR #110, opened 2021-04-02
fix: assume node returns buffers, not strings

FauxFaux:
Sub-optimal fix for #109; just assume everything is a buffer, and `toString` it in other places.

### DIFF
--- a/src/interceptors/ClientRequest/utils/getIncomingMessageBody.ts
+++ b/src/interceptors/ClientRequest/utils/getIncomingMessageBody.ts
@@ -1,13 +1,13 @@
 import { IncomingMessage } from 'http'
 
-export function getIncomingMessageBody(res: IncomingMessage): Promise<string> {
-  let responseBody = ''
+export function getIncomingMessageBody(res: IncomingMessage): Promise<Buffer> {
+  const bufs: Buffer[] = [];
 
   return new Promise((resolve, reject) => {
     res.once('error', reject)
-    res.on('data', (chunk) => (responseBody += chunk))
+    res.on('data', (chunk) => bufs.push(chunk))
     res.once('end', () => {
-      resolve(responseBody)
+      resolve(Buffer.concat(bufs))
     })
   })
 }
--- a/test/regressions/http-compression.test.ts
+++ b/test/regressions/http-compression.test.ts
@@ -1,0 +1,49 @@
+/**
+ * @jest-environment node
+ * @see https://github.com/mswjs/interceptors/issues/109
+ */
+import got from 'got';
+import compression from 'compression';
+import * as zlib from 'zlib';
+
+import { createServer, ServerApi } from '@open-draft/test-server'
+import { createInterceptor, IsomorphicResponse } from '../../src'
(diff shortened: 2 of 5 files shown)


src/interceptors/ClientRequest/utils/getIncomingMessageBody.ts
```diff
@@ -1,13 +1,13 @@
 import { IncomingMessage } from 'http'
 
-export function getIncomingMessageBody(res: IncomingMessage): Promise<string> {
-  let responseBody = ''
+export function getIncomingMessageBody(res: IncomingMessage): Promise<Buffer> {
+  const bufs: Buffer[] = [];
 
   return new Promise((resolve, reject) => {
     res.once('error', reject)
-    res.on('data', (chunk) => (responseBody += chunk))
```
kettanaito, line 8: Are there any implications if we handle this `responseBody` and `chunk` handling as Buffer here instead of adjusting all the other places?

kettanaito:
Hey, @FauxFaux. Thank you for the fix! 
I've got one question if we can approach it vice versa. Please, could you submit a test for this issue? We can start by adding a unit test to `getIncomingMessageBody` function at `src/interceptors/ClientRequest/utils/getIncomingMessageBody.test.ts`. If you encounter any struggle with the test, just let me know here about the use case so we could replicate it reliably in a test together. Thanks.

FauxFaux:
I've attached an integration test for the issue, against npm:compression and the already included `got`, which does compression by default.

I notice my change has broken other stuff on CI, apparently I didn't run the right tests locally, because the browsers scared me off.

test/regressions/http-compression.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  await server.close()
+})
+
+const GZIP_MAGIC_HEX = '1f8b08';
+
+test('returns a valid body', async () => {
+  const intercepted = new Promise<IsomorphicResponse>(resolve =>
+    interceptor.on('response', (req, resp) =>
+      resolve(resp)));
+  expect(await got(server.http.makeUrl('/')).json()).toEqual({hello: 'world'});
```
kettanaito, line 42: I like how this test showcases real usage, but I find it having too many moving parts to be straightforward. I think we should get rid of as many dependencies as we can and narrow it down to the barebone case. For instance:

1. Replace `got` with `http`. 
2. Not sure why `zlib` is here. Looks like it handles the compressed buffer, but doesn't that introduce a middle-man in the logic? Unless the issue is zlib-specific.

FauxFaux: `zlib` is here to validate the returned response is valid, and not otherwise corrupted by the middleware. Unfortunately zlib throws an absolutely useless error ("invalid!") when it sees invalid data.

I don't plan to further contribute to this issue at this time, as the library has other bugs which mean that it is not useful to me, and which I can't easily reproduce outside of our application. e.g. it causes supertest to "hang" (never resolve) during some error situations.

kettanaito: Thank you for your effort anyway, @FauxFaux. I'll see if this can be turned into a properly isolated test suite.

Issue #111, opened 2021-04-07
Node request interceptor is getting skipped for the response interceptor


nravic:
When attaching a debugger to the following code (with breakpoints set at the request and response interceptors):

```typescript
export class NodeHttpClient implements Provider<AxiosLike> {
	value() {
		const interceptor = createInterceptor({
			modules: nodeInterceptors,
			resolver(request) {},
		});

		interceptor.on('request', (request) => {
			console.log(request);
			return request;
		});

		interceptor.on('response', (request, response) => {
                        console.log(request, response);
                        return response;
		});

		const instance = axios.create();
		instance.defaults.adapter = require('axios/lib/adapters/http');
		interceptor.apply();
		return instance;
	}
}
```
it appears as if the request interceptor is skipped and execution is started with the response interceptor when a request is initiated. I've tried this with other interceptor modules and they all behave the same way. 

I'd appreciate any help looking into this or a direction to be pointed towards to resolve this, especially if it's something minor I'm overlooking. 

Thank you!

kettanaito:
Hey, @nravic. Thanks for reporting this.

It looks like the current [ClientRequest interceptor](https://github.com/mswjs/interceptors/blob/ae27f7466b66340ab5bea1175b3c152c235b08d0/src/interceptors/ClientRequest/createClientRequestOverride.ts) doesn't emit the "request" event at all. The same can be said about the XMLHttpRequest interceptor. It may be a typo (response -> request), or simply a bug. 

Please, would you be interested in opening a pull request to add `observer.emit('request', isoRequest)` calls after these lines?

https://github.com/mswjs/interceptors/blob/ae27f7466b66340ab5bea1175b3c152c235b08d0/src/interceptors/ClientRequest/createClientRequestOverride.ts#L179-L184

https://github.com/mswjs/interceptors/blob/ae27f7466b66340ab5bea1175b3c152c235b08d0/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L243-L248

PR #112, opened 2021-04-08
Emits the "request" event in the ClientRequest and XMLHttpRequest overrides

nravic:
Fix for https://github.com/mswjs/interceptors/issues/111.


chaoyangnz:
nice to be fixed.

chaoyangnz:
question? I don't think `request` event is really useful, when it has `Resolver` which even can get the raw http.Request

kettanaito:
@chaoyangnz, it's a good point. I suppose the difference is mainly semantic: `resolver` is meant to return a mocked response (even if conditionally). The "request" event is a ready-only way to find out that a request has happened. Anyway, this is a good point to refine at v1.0

kettanaito (approved):
Thank you for fixing this, @nravic! I've added a test suite that asserts the "request" event being properly emitted. 

Welcome to contributors!

nravic:
Sounds good, thank you!

nravic:
@kettanaito do you have an estimate of when you'll publish the npm package with the changes so I can use it? 

Thank you again!

kettanaito:
@nravic, I'll publish it on the weekend. Meanwhile, you can still react to requests in the `resolver` argument of `createInterceptors()` function.

PR #114, opened 2021-04-21
Fix condition to assume an `URL` object

lauraseidler:
Fixes https://github.com/mswjs/msw/issues/705

### DIFF
--- a/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
@@ -207,6 +207,7 @@ test('handles [RequestOptions, callback] input', () => {
     method: 'POST',
     protocol: 'https:',
     host: 'mswjs.io',
+    origin: 'https://mswjs.io', // see: https://github.com/mswjs/msw/issues/705
     path: '/resource',
     headers: {
       'Content-Type': 'text/plain',
--- a/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
@@ -67,7 +67,7 @@ export function normalizeHttpRequestParams(
   }
   // Handle a given URL instance as-is
   // and derive request options from it.
-  else if ('origin' in args[0]) {
+  else if (args[0] instanceof URL) {
     url = args[0]
     debug('given a URL:', url)
 


kettanaito:
Leaving here the variable input types for posterity.

## `string`

`string -> URL -> RequestOptions`

## `URL`

`URL -> RequestOptions`

## `RequestOptions` (`ClientRequestArgs`)

`RequestOptions -> URL`

This must _not_ be treated as a URL. URL must be derived from these options.

```ts
interface ClientRequestArgs {
  protocol?: string | null;
  host?: string | null;
  hostname?: string | null;
  family?: number;
  port?: number | string | null;
  defaultPort?: number | string;
  localAddress?: string;
  socketPath?: string;
  /**
    * @default 8192
    */
  maxHeaderSize?: number;
  method?: string;
  path?: string | null;
  headers?: OutgoingHttpHeaders;
  auth?: string | null;
  agent?: Agent | boolean;
  _defaultAgent?: Agent;
  timeout?: number;
  setHost?: boolean;
  // https://github.com/nodejs/node/blob/master/lib/_http_client.js#L278
  createConnection?: (options: ClientRequestArgs, oncreate: (err: Error, socket: Socket) => void) => Socket;
}
```

> —[@types/node](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f7ec78508c6797e42f87a4390735bc2c650a1bfd/types/node/http.d.ts#L83-L106)

## Legacy URL

`LegacyUrl -> URL -> RequestOptions`

This must _not_ be treated as a URL. URL must be derived from these options.

```ts
interface Url {
  auth: string | null;
  hash: string | null;
  host: string | null;
  hostname: string | null;
  href: string;
  path: string | null;
  pathname: string | null;
  protocol: string | null;
  search: string | null;
  slashes: boolean | null;
  port: string | null;
  query: string | null | ParsedUrlQuery;
}
```

> — [@types/node](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f7ec78508c6797e42f87a4390735bc2c650a1bfd/types/node/url.d.ts#L23-L37)

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
```diff
@@ -207,6 +207,7 @@ test('handles [RequestOptions, callback] input', () => {
     method: 'POST',
     protocol: 'https:',
     host: 'mswjs.io',
+    origin: 'https://mswjs.io', // see: https://github.com/mswjs/msw/issues/705
```
kettanaito, line 210: Strangely, `RequestOptions` doesn't have the `origin` property ([type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f7ec78508c6797e42f87a4390735bc2c650a1bfd/types/node/http.d.ts#L83-L106)). It once again looks like some arbitrary object is constructed by a third-party library that is not suitable as an input to `ClientRequest`. Unknown properties don't do harm in JavaScript, but they are a culprit when determining what was intended as an input. 

You can verify this yourself:

```ts
// Removed "string" and "URL" as irrelevant inputs.
function test(obj: RequestOptions | LegacyURL): void {}

test({
  origin: '.' // can only specify known properties
})
```

I don't think we should cover each individual (arguably improper) usage of request options. Instead, let's craft a condition for the `URL` case using _other_ properties that URL has (i.e. `hash`). 

```ts
if ('origin' in args[0] && 'hash' in args[0]) {}
```

I think that checking for `instanceof URL` is still a good call, and we should leave it:

```ts
if (args[0] instanceof URL || ('origin' in args[0] && 'hash' in args[0])) {}
```

The bigger this condition grows, the more it asks to be its own function, isolated and unit tested. 

Please, would you be interested in creating this function (i.e. `src/interceptors/ClientRequest/utils/isUrl.ts`)?

lauraseidler: Where would you see the issues with using `instanceof`? My thinking here is, if you want to guarantee that a `URL` (including all properties and methods like `toString` and `toJSON`) is returned, as defined here:

https://github.com/mswjs/interceptors/blob/16fd35622efe6328497ade15e21a84f6afda0a93/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts#L47

the only check for that really _is_ `instanceof` (or testing for every single property and method), otherwise you won't ever really be sure. Node itself (no Typescript), also only [specifies `Object`](https://nodejs.org/api/http.html#http_http_request_options_callback), so while other options won't be useful, they're also not explicitly forbidde.

IMO, a good way to go would be to use `instanceof` for the `URL` condition, and then tweak the condition for the legacy urls to make sure this really only fits legacy urls. Everything else by my understanding should then be picked up by this case:

https://github.com/mswjs/interceptors/blob/16fd35622efe6328497ade15e21a84f6afda0a93/src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts#L123

which then basically does just about what you suggested:

https://github.com/mswjs/interceptors/blob/16fd35622efe6328497ade15e21a84f6afda0a93/src/utils/getUrlByRequestOptions.ts#L91

kettanaito: Sounds reasonable. Let's go with the `instanceof` alone and release this is a breaking change. I do hope this eliminates options-related issues in the future. Thanks!

kettanaito (approved):

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.test.ts
lauraseidler: Sounds good - thanks for the quick responses!

kettanaito:
Thank you for the amazing work on investigating and fixing this, @lauraseidler. Welcome to contributors! 🎉

lauraseidler:
@kettanaito Cheers! Feel free to ping me if this actually breaks anything :)

Issue #115, opened 2021-04-26
ClientRequest patch is never restored


kettanaito:
The following patch of the `ClientRequest` class doesn't seem to get restored anywhere:

https://github.com/mswjs/interceptors/blob/16fd35622efe6328497ade15e21a84f6afda0a93/src/interceptors/ClientRequest/index.ts#L47

## Solution

Restore the `http.ClientRequest` to the `pureClientRequest` (stored original class) in the cleanup function of `interceptClientRequest`:

https://github.com/mswjs/interceptors/blob/16fd35622efe6328497ade15e21a84f6afda0a93/src/interceptors/ClientRequest/index.ts#L113-L122

## GitHub

- Originated from https://github.com/mswjs/msw/issues/710

mathanpec:
Hi @kettanaito,

I see a small problem with just restoring `http.ClientRequest` with `pureClientRequest` since the `pureClientRequest` is getting initialised itself only when the actual request gets fired inside `requestOverride`.  But in large code based with multiple test files `server.listen()` and `server.close()` will happen for every test file irrespective of whether the request is going to be fired inside a test file or not. So in that case `pureClientRequest` will not be initialised and when we do `server.close()` after the test file execution we will end up setting `undefined` to `http.ClientRequest`. We can have null check there but that kinda feels hacky and it will be difficult to recall later as to why that null condition was added.

So on further looking into the code, I feel we can completely remove the monkey patching of `http.ClientRequest` itself and directly return object created out of `ClientRequestOverride` itself from function `handleClientRequest`. Since `ClientRequestOverride` is already inheriting from 'http.ClientRequest', it will have access to all the proto functions of `http.ClientRequest` also.

After removing the monkey patching completely, I ran the test cases and they all look green. Also heap memory also stays constant between the test files after this.

mathanpec:
@kettanaito ,
Since PR: https://github.com/mswjs/interceptors/pull/116 is merged, can we close this also?

kettanaito:
Yes, @mathanpec, I'll close this.

PR #116, opened 2021-04-27
Removes ClientRequest monkey patching

mathanpec:
- Fixes https://github.com/mswjs/msw/issues/710 (memory leak in MSW)

Discussion on this can be found at issue https://github.com/mswjs/interceptors/issues/115

### DIFF
--- a/src/interceptors/ClientRequest/createClientRequestOverride.ts
+++ b/src/interceptors/ClientRequest/createClientRequestOverride.ts
@@ -284,17 +284,7 @@ export function createClientRequestOverride(
       // Decide whether to use HTTPS based on the URL protocol.
       // XHR can trigger http.request for HTTPS URL.
       if (url.protocol === 'https:') {
-        debug('reverting patches...')
-        const { ClientRequest } = http
-
-        http.ClientRequest = pureClientRequest
-
         request = pureMethod(options)
-
-        debug('re-applying patches...')
-
-        // @ts-ignore
-        http.ClientRequest = ClientRequest
       } else {
         request = pureMethod(url.toString(), options)
       }
--- a/src/interceptors/ClientRequest/index.ts
+++ b/src/interceptors/ClientRequest/index.ts
@@ -39,17 +39,11 @@ function handleClientRequest(
     observer,
     resolver,
   })
-
-  debug('patching native http.ClientRequest...')
-
-  // Only http.ClientRequest is overridden because https uses http
-  // @ts-ignore
-  http.ClientRequest = ClientRequestOverride
-
-  debug('new http.ClientRequest (origin: %s)', protocol)
+  
+  debug('new ClientRequestOverride (origin: %s)', protocol)
 
   // @ts-expect-error Variable call signature.
-  return new http.ClientRequest(...args)
+  return new ClientRequestOverride(...args)
 }
 
 /**


src/interceptors/ClientRequest/createClientRequestOverride.ts
```diff
@@ -284,17 +284,7 @@ export function createClientRequestOverride(
       // Decide whether to use HTTPS based on the URL protocol.
       // XHR can trigger http.request for HTTPS URL.
       if (url.protocol === 'https:') {
-        debug('reverting patches...')
```
mathanpec, line 287: Since we are removing monkey patch, we can get away with this also

kettanaito:
Hey, @mathanpec. Thank you for analyzing the issue and providing this fix! Please give me some time to carefully read through this and test it locally, as changes like these are quite fundamental and their consequences aren't always evident.

kettanaito:
I've tested these changes locally as well as in the MSW build linked to this version of `interceptors`. Didn't encounter any issues.

I think the previous logic was added since the original implementation of this library is heavily inspired by Nock:

https://github.com/nock/nock/blob/3efd7382011a8d1ca084bfc4043a332a796a0618/lib/intercept.js#L409-L416

Looks like we don't need to restore/patch again the `ClientRequest` due to the difference in how Nock and Interceptors work:

- Nock _patches_ `ClientRequst` conditionally, because it has a built-in assertion and knows if a request has a corresponding handler.
- Interceptors doesn't have a built-in assertion, which means we _always_ patch `ClientRequest`, and only need to ensure that we use the unpatched version when performing a bypassed (as-is) request (when the mocked response wasn't returned as a part of the ClientRequest's operations).

kettanaito (approved):

src/interceptors/ClientRequest/index.ts
```diff
@@ -39,17 +39,11 @@ function handleClientRequest(
     observer,
     resolver,
   })
-
-  debug('patching native http.ClientRequest...')
-
-  // Only http.ClientRequest is overridden because https uses http
-  // @ts-ignore
-  http.ClientRequest = ClientRequestOverride
```
kettanaito, line 47: It's worth mentioning that by removing the `ClientRequest` patching we can no longer spy on the requests that are made by creating a direct `http.ClientRequest` instance. All the use-cases pass at the moment because we assert against `http.get`/`http.request` Node.js API that encapsulates `ClientRequest`. We stub those `get`/`request` functions, and that's how we are able to spy on any made request. 

I'd like to research if the direct usage of `ClientRequest` is a thing, and whether we should cover for it. 

My main concern is that there may be third-party libraries that use `ClientRequest` directly.

kettanaito: > This object is created _internally_ and returned from `http.request()`
> —[Node.js documentation of `http.ClientRequest`](https://nodejs.org/api/http.html#http_class_http_clientrequest)

I think there is no use case to create the `http.ClientRequest` instance directly. Looks like we may omit to patch the `ClientRequest` directly.

mathanpec: Yes, that might be a valid use case. But if thats needed then patching should be done during the initialisation (`server.listen()`) time itself alongside where we patch `http.get / http.request` I think..

PR #117, opened 2021-04-29
fix: set responseXML when resolver returns xml

rrogowski:
Prior to this PR, `responseXML` was always set to `null`, even if the `Content-Type: */xml` header was present. This PR updates the XMLHttpRequest interceptor to return `responseXML` when XML is returned by the resolver.

I originally opened https://github.com/mswjs/msw/issues/715 for this issue before I understood that this repo was a better location for the fix.

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -9,6 +9,7 @@ import {
   objectToHeaders,
   headersToString,
 } from 'headers-utils'
+import { DOMParser } from 'xmldom'
 import { IsomorphicRequest, Observer, Resolver } from '../../createInterceptor'
 import { parseJson } from '../../utils/parseJson'
 import { bufferFrom } from './utils/bufferFrom'
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -292,6 +292,7 @@ export const createXMLHttpRequestOverride = (
             debug('response type', this.responseType)
             this.response = this.getResponseBody(mockedResponse.body)
             this.responseText = mockedResponse.body || ''
+            this.responseXML = this.getResponseXML()
 
             debug('set response body', this.response)
 
@@ -513,6 +514,14 @@ export const createXMLHttpRequestOverride = (
         default:
           return textBody
       }
+    }
+
+    getResponseXML() {
+      const contentType = this.getResponseHeader('Content-Type')
+      if (contentType === 'application/xml' || contentType === 'text/xml') {
+        return new DOMParser().parseFromString(this.responseText, contentType)
+      }
+      return null
     }
 
     /**
--- a/test/regressions/xhr-response-xml-body.test.ts
+++ b/test/regressions/xhr-response-xml-body.test.ts
@@ -1,0 +1,67 @@
+import { createInterceptor } from '../../src';
+import { interceptXMLHttpRequest } from '../../src/interceptors/XMLHttpRequest';
+import { createXMLHttpRequest } from '../helpers';
+
+const XML_STRING = '<node key="value">Content</node>'
+
+describe('Content-Type: application/xml', () => {
+  const interceptor = createInterceptor({
+    modules: [interceptXMLHttpRequest],
+    resolver() {
+      return {
+        headers: { 'Content-Type': 'application/xml' },
+        status: 200,
+        body: XML_STRING,
+      }
+    },
+  })
+
+  beforeAll(() => {
+    interceptor.apply()
+  })
+
+  afterAll(() => {
+    interceptor.restore()
+  })
+
+  test('supports XHR mocked response with an XML response body', async () => {
+    const req = await createXMLHttpRequest((req) => {
+      req.open('GET', '/arbitrary-url')
+    })
+
+    expect(req.responseXML).toStrictEqual(
+      new DOMParser().parseFromString(XML_STRING, 'application/xml')
+    )
+  })
+})
+
+describe('Content-Type: text/xml', () => {
+  const interceptor = createInterceptor({
+    modules: [interceptXMLHttpRequest],
+    resolver() {
+      return {
+        headers: { 'Content-Type': 'text/xml' },
+        status: 200,
+        body: XML_STRING,
+      }
+    },
+  })
+
+  beforeAll(() => {
+    interceptor.apply()
+  })
+
+  afterAll(() => {
+    interceptor.restore()
+  })
+
+  test('supports XHR mocked response with an XML response body', async () => {
+    const req = await createXMLHttpRequest((req) => {
+      req.open('GET', '/arbitrary-url')
+    })
+
+    expect(req.responseXML).toStrictEqual(
+      new DOMParser().parseFromString(XML_STRING, 'text/xml')
+    )
+  })
+})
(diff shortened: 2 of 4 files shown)


test/regressions/xhr-response-xml-body.test.ts
```diff
@@ -0,0 +1,67 @@
+import { createInterceptor } from '../../src';
+import { interceptXMLHttpRequest } from '../../src/interceptors/XMLHttpRequest';
+import { createXMLHttpRequest } from '../helpers';
+
+const XML_STRING = '<node key="value">Content</node>'
+
+describe('Content-Type: application/xml', () => {
```
rrogowski, line 7: I wonder if there is a better way to consolidate these tests while still checking that both `Content-Type` headers will yield a value for `responseXML`. 🤔 As far as I can tell, `text/xml` and `application/xml` are used fairly interchangeably, but [RFC (3023)](https://tools.ietf.org/html/rfc3023) specifies the following:
> If an XML document -- that is, the unprocessed, source XML document -- is readable by casual users, text/xml is preferable to application/xml.  MIME user agents (and web user agents) that do not have explicit support for text/xml will treat it as text/plain, for example, by displaying the XML MIME entity as plain text. Application/xml is preferable when the XML MIME entity is unreadable by casual users.

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -515,6 +516,14 @@ export const createXMLHttpRequestOverride = (
       }
     }
 
+    getResponseXML() {
+      const contentType = this.getResponseHeader('Content-Type')
+      if (contentType === 'application/xml' || contentType === 'text/xml') {
+        return new DOMParser().parseFromString(this.responseText, contentType)
```
rrogowski, line 522: FYI, in the event of an invalid XML string, `parseFromString` will return a `Document` representing an error state. See https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString#error_handling.

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -515,6 +516,14 @@ export const createXMLHttpRequestOverride = (
       }
     }
 
+    getResponseXML() {
+      const contentType = this.getResponseHeader('Content-Type')
+      if (contentType === 'application/xml' || contentType === 'text/xml') {
```
kettanaito, line 521: Do you think it's worth to have a more generic content type check like `contentType.endsWith('xml')`, or would it be too insecure on the library's side?

kettanaito: There's not big conceptual reason, but I generally prefer checking multiple value equality as `Array.prototype.includes`:

```js
if (['text/xml', 'application/xml'].includes(contentType)) {}
```

rrogowski: Hi @kettanaito! Good point. I chose this particular solution because `parseFromString` only accepts a `DOMParserSupportedType` for the second argument. By checking directly against strings, I'm taking advantage of type inference in order to satisfy the TypeScript compiler.

We could still use `.endsWith` or `.includes` and do an explicit type cast like so:
```typescript
const contentType = this.getResponseHeader('Content-Type') ?? ''
if (['application/xml', 'text/xml'].includes(contentType)) {
  return new DOMParser().parseFromString(
    this.responseText,
    contentType as DOMParserSupportedType
  )
}
```

But I thought this was a little less clean. Let me know which approach you think is better, or if you have any other ideas!

rrogowski: > Do you think it's worth to have a more generic content type check like `contentType.endsWith('xml')`, or would it be too insecure on the library's side?

I'm not sure -- I don't think I have enough experience to know if this would pose a security issue.

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
kettanaito, line 522: It appears that `DOMParser` is a browser-specific API (the name implies, but I've still double-checked):

- [MDN](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser)
- [Why can't I use DOMParser in Node.js? (StackOverflow)](https://stackoverflow.com/a/55668667/2754939)

May I please ask you to verify this by adding a Node.js-oriented test, just as you've done with the JSDOM one?
You can achieve that by copying your existing test and appending the `@jest-environment node` pragma comment to make Jest run that test file in Node.js:

```js
// test/regressions/xhr-response-xml-body.node.test.ts
/**
 * @jest-environment node
 */
```

If my findings are correct, this new test would fail. Let's discuss that if that happens. Thank you!

kettanaito (changes requested):
Thank you once more for tackling this, I think it's a great addition!
I have one concern about DOMParser being a browser-specific API, which would error when users attempted to use this library in Node.js (which is a huge segment it's designed for).

Let's give it a try in a new Node.js-oriented test and see how it behaves. Perhaps we'd consider using a third-party package that provides a similar functionality to `DOMParser` to ensure operability across both browser and Node.js

rrogowski:
@kettanaito Fair point about `DOMParser` being browser-specific. I can see if any good third-party packages exist to provide that functionality.

One observation -- the code in question is specifically for intercepted XMLHttpRequests, the concept of which only exists in the browser (unless using a polyfill from node). In that case, do we need to write a test for the node environment that instantiates an XMLHttpRequest using an existing polyfill library?

rrogowski:
Another observation. These lines lead me to believe that XMLHttpRequests are not intercepted if `window` is not defined: https://github.com/mswjs/interceptors/blob/master/src/interceptors/XMLHttpRequest/index.ts#L6-L25

`window` would only be defined in a browser-like environment (i.e. JSDOM), so it appears to me as though this library does not support intercepting XMLHttpRequests from a pure node environment. Am I understanding that correctly? If so, I would posit the following:
- Use a third-party library that is node- and browser-compatible, as you suggested, to future-proof the code
- Skip the pure node environment test, as this would be involve supporting XMLHttpRequest interception in node, which is outside the original scope of this PR

kettanaito:
Yes, we still need a Node-specific test for XMLHttpRequest to target JSDOM—one of the most used polyfills for Node that ships with `XMLHttpRequest`. 

> so it appears to me as though this library does not support intercepting XMLHttpRequests from a pure node environment

Correct, there's no such thing as `XMLHttpRequest` in a pure Node.js environment, as you've rightfully pointed out. However, this library supports browser-like environments (i.e. JSDOM), where XHR is defined while `DOMParser` _may be not_. To support both actual browser and browser-like environments it makes sense to use a DOMParser polyfill on the library's side to guarantee a seamless experience. 

> Skip the pure node environment test, as this would be involve supporting XMLHttpRequest interception in node without the presence of JSDOM or some similar library, which is outside the original scope of this PR

I think you are right on this. I might have been wrong in pursuing Node tests on this. What I meant is to ensure:

- The feature works in an actual browser.
- The feature works in browser-like environments (i.e. JSDOM). For this part, we'd have to use a polyfill.

rrogowski:
Hi @kettanaito, thanks for the great discussion and feedback so far! I've just amended the PR to use a polyfill for `DOMParser`. I'm using `xmldom`, which supports both the node and browser environments. This should satisfy the requirements you mentioned above:
> - The feature works in an actual browser.
> - The feature works in browser-like environments (i.e. JSDOM). For this part, we'd have to use a polyfill.

Hopefully I understood all of your feedback so far and didn't miss anything you asked for, but if I did please let me know! 😄

kettanaito (approved):
This is a superb addition to how we handle the XML response body. Thank you, @rrogowski! Your work on this is outstanding. Welcome to contributors! 🎉 

I'd be happy to hear more from you, if you find any interesting tasks in this library or throughout the MSW ecosystem. Thank you for being awesome.

PR #118, opened 2021-05-07
Uses "jest-setup-glob" to simplify testing

kettanaito:
With the addition of `jest-setup-glob` dependency we can remove the following commands:

- `yarn test:internal`
- `yarn test:integration`
- `yarn test:integration:node`
- `yarn test:integration:browser`

And use a single `yarn test` command that will use a proper environment setup based on the given test filename.

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -1,0 +1,5 @@
+const setup = require('jest-setup-glob')
+
+setup({
+  '**.browser.test.ts': './test/jest.browser.setup.ts',
+})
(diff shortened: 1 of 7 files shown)


jest.setup.js
```diff
@@ -0,0 +1,5 @@
+const setup = require('jest-setup-glob')
+
+setup({
+  '**.browser.test.ts': './test/jest.browser.setup.ts',
```
kettanaito, line 4: Whenever a test file name matches this glob, the associated Jest setup will be used. Otherwise, the default setup is used.

kettanaito:
Something is terribly wrong with this CI build: random dependencies start to fail as if not being installed.

kettanaito:
We're not using Jest anymore and it's fairly easy to do glob-based environment matching with Vitest.

PR #120, opened 2021-05-24
Unref SocketPolyfill#setTimeout()

esamattis:
Noticed that my tests hanged in Jest with

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

when I has using the Elasticsearch client with msw passthrough. It used 30 second timeout. 

After comparing to the real socket in node.js core I noticed it uses unreffed timers so I guess it should be used here too. At least this fixes my hanging test issue. Although I'm not sure why the SocketPolyfill is in play with the passthrough requests.

Node.js source links:
https://github.com/nodejs/node/blob/7766b3853a5a521fd4e8994fab5e9d4b6c83ccba/lib/net.js#L458
https://github.com/nodejs/node/blob/7766b3853a5a521fd4e8994fab5e9d4b6c83ccba/lib/internal/stream_base_commons.js#L264

### DIFF
--- a/src/interceptors/ClientRequest/polyfills/SocketPolyfill.ts
+++ b/src/interceptors/ClientRequest/polyfills/SocketPolyfill.ts
@@ -94,10 +94,18 @@ export class SocketPolyfill extends EventEmitter {
   }
 
   setTimeout(timeout: number, callback?: () => void): SocketPolyfill {
-    setTimeout(() => {
+    const timer = setTimeout(() => {
       callback?.()
       this.emit('timeout')
     }, timeout)
+
+    /**
+     * Unref the timer in node.js so the process won't hang on exit if long
+     * timeouts were used
+     */
+    if (typeof timer.unref === 'function') {
+      timer.unref()
+    }
 
     return this
   }
--- a/src/interceptors/ClientRequest/polyfills/SocketPolyfill.ts
+++ b/src/interceptors/ClientRequest/polyfills/SocketPolyfill.ts
@@ -99,10 +99,8 @@ export class SocketPolyfill extends EventEmitter {
       this.emit('timeout')
     }, timeout)
 
-    /**
-     * Unref the timer in node.js so the process won't hang on exit if long
-     * timeouts were used
-     */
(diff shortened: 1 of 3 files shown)


kettanaito:
Hey, @esamattis. Thank you for fixing this!

Please, could you share how you're making a request that results in a timeout? It'd be great to add an integration test that ensures this issue doesn't happen again.

esamattis:
Here's a reproduce without any testing framework

```js
const { rest } = require("msw");
const { setupServer } = require("msw/node");
const http = require("http");

const server = setupServer(
    rest.get("http://test.invalid/page", (req, res, ctx) => {
        return res(ctx.json({ message: "Hello World" }));
    }),
);

server.listen();

const req = http.request("http://test.invalid/page", (res) => {
    console.log("RES", res.statusCode);

    server.close();
});

req.setTimeout(5000);

req.end();
```

Just run it `node example.js` and it will hang after the response status is logged. If you remove the `req.setTimeout(5000);` line it will exit immediately.

kettanaito:
Thank you for the reproduction example. I've tried putting it in a test, but haven't found a way to assert that the process doesn't hang (although I can validate that manually and confirm your change fixing the issue). `process.on('exit')` never emits, and I cannot get any other means to reject the test if the process keeps hanging due to an unterminated socket timeout. 

What's interesting, is that socket's timeout is not cleared, but only un-referenced. If we add `clearTimeout(timer)` after `timer.unref()`, then we can assert that the socket timeout is not called. Yet that's now how Node.js behaves, so it'd be a deviation, and I'm not comfortable with assessing its implications.

kettanaito:
I came to the conclusion that to reliably reflect the expectations in the test, it must spawn _another_ test in a child process. That's because the expectations here are laid against the Jest test runner's behavior (that it doesn't hang with an unterminated process).

kettanaito (approved):
Thank you for providing the fix, @esamattis! Welcome to contributors 🎉

PR #124, opened 2021-05-29
fix: add support for Jest 27

### LINKED ISSUE #123: Does not work well with jest 27
[Jest 27 removes `setImmediate`](https://github.com/facebook/jest/blob/master/CHANGELOG.md#fixes-2), which is used in [createClientRequestOverride](https://github.com/mswjs/interceptors/blob/f95c5c2d6028e22bd0681695bc4887c89cac81bc/src/interceptors/ClientRequest/createClientRequestOverride.ts#L124). 

Currently, I get the following in my tests:
```
ReferenceError: setImmediate is not defined
          at ClientRequestOverride.write (~/node_modules/@mswjs/interceptors/src/interceptors/ClientRequest/createClientRequestOverride.ts:124:7)
          at writeToStream (~/node_modules/node-fetch/lib/index.js:653:8)
          at ~/node_modules/node-fetch/lib/index.js:1627:3

```

weyert:
Jest 27 removes the `setImmediate`-global so replaced it with setTimeout with a 0 delay

fixes #123

### DIFF
--- a/src/interceptors/ClientRequest/createClientRequestOverride.ts
+++ b/src/interceptors/ClientRequest/createClientRequestOverride.ts
@@ -121,9 +121,9 @@ export function createClientRequestOverride(
         }
       }
 
-      setImmediate(() => {
+      setTimeout(() => {
         this.emit('drain')
-      })
+      }, 0)
 
       return false
     }


src/interceptors/ClientRequest/createClientRequestOverride.ts
```diff
@@ -121,9 +121,9 @@ export function createClientRequestOverride(
         }
       }
 
-      setImmediate(() => {
+      setTimeout(() => {
         this.emit('drain')
-      })
+      }, 0)
```
jonkoops, line 126: `0` is the default value according the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout#parameters):

> If this parameter is omitted, a value of 0 is used, meaning execute "immediately", or more accurately, the next event cycle. 


This means you can omit the argument for the delay.
```suggestion
      })
```

jonkoops:
Can confirm this is causing issues in our application as well, I'd love to see this merged as well.

src/interceptors/ClientRequest/createClientRequestOverride.ts
weyert: Yeah, I thought using `0` as timeout made it more clear (without reading the dosc) that it will immediately run

christemple: Not sure which I prefer, personally I lean towards being explicit about the intention i.e. what weyert is suggesting, that way the intention is self documented as opposed to leaving it open to the possibility of someone assuming `0` was forgotten to be passed.

Either way I don't mind, just thought I'd come by and thank you guys for doing a fix, been bugging me all morning :)

kettanaito: Explicit `0` is perfectly fine, you can leave it like it is.

christemple (approved):

src/interceptors/ClientRequest/createClientRequestOverride.ts
weyert: Sweet :)

weyert:
@kettanaito  Did you had a chance to have a look?

msutkowski (approved):

kettanaito (approved):
Thank you for tackling this, @weyert.

PR #127, opened 2021-06-05
Adds jest@27 compatibility

### LINKED ISSUE #126: Jest@27 compatibility
- [jest@27.0 Release notes](https://github.com/facebook/jest/releases/tag/v27.0.0)

## Issues

- [x] JSDOM is no longer the default test environment.
- [x] `setImmediate` is not defined (#123)
- [x] `TypeError [ERR_INVALID_ARG_TYPE]: The "options.agent" property must be one of Agent-like Object, undefined, or false. Received an instance of Object`
- [x] Test functions cannot both take a 'done' callback and return something. Either use a 'done' callback, or return a promise. (previously less strict). 
- [x] `ReferenceError: browserType.launch: setImmediate is not defined` when running in-browser tests.

kettanaito:
- Fixes #126
- Follow-up to #124 

## Changes

- Updates to `jest@27`.
- Sets explicit `@jest-environment` to JSDOM, as `node` is the default test environment since jest 27.
- `cloneObject` now copies class prototype to the cloned object (fixes `options.agent` issue).
- `cloneObject` now skips cloning any non-plain object instances to prevent recursive issues when cloning circular references in classes.


kettanaito:
## `setImmediate` and Playwright

It's still unclear as to how `setImmediate` is intended to be polyfilled when it's used by third-party libraries. I've raised the question in Reactiflux. It's confusing that Jest sets a browser and Node.js global to `undefined`.

src/interceptors/ClientRequest/utils/normalizeHttpRequestParams.ts
```diff
@@ -17,15 +17,15 @@ function resolveRequestOptions(
   args: HttpRequestArgs,
   url: URL
 ): RequestOptions {
-  // Calling `fetch` provides only URL to ClientRequest,
-  // without RequestOptions or callback.
-  if (['function', 'undefined'].includes(typeof args[1])) {
+  // Calling `fetch` provides only URL to `ClientRequest`
+  // without any `RequestOptions` or callback.
+  if (typeof args[1] === 'undefined' || typeof args[1] === 'function') {
```
kettanaito, line 22: Getting properly narrowed type this way.

PR #129, opened 2021-06-29
Supports request interception in a ChildProcess

kettanaito:
## Motivation

Since this library patches the request-issuing modules in the _current process_, there is no way to enable request interception in a _child_ process (i.e. when testing a running application). An alternative has been suggested previously to enable the interception on the application's side, for example:

```js
// src/index.js
createInterceptors({ ... }).apply()
```

However, this implies that:

- Response resolution logic needs to be described in the app (possibly on runtime via environmental variables).
- A single test cannot affect the response resolution logic, as the application and a test live in separate universes.

This change aims to support request interception in a child process by using the IPC between the parent (test) and the child (spawned application) processes and signaling requests/responses between them. 

## Changes

- Isomorphic requests now have a required `id` property—a UUID representing a request and assigned by the library.
- The library now exports two new functions: `createRemoteInterceptor` and `createRemoteResolver`, to be used in child and parent processes respectively. 

## Usage example

```js
// parent.js
import { spawn } from 'child_process'
import { createRemoteResolver } from '@mswjs/interceptors/remote'

const serverProcess = spawn('node', ['child.js'], {
  // It's required to spawn the child process with the "ipc" stdio
  // in order to have the IPC working in Node.js.
  stdio: ['inherit', 'inherit', 'inherit', 'ipc']
})

// Receives the request events from the given child "process"
// and uses the "resolver" function to produce a mocked response.
// Sends the mocked response to the child process to be used.
createRemoteResolver({
  process: serverProcess,
  resolver(request) {
    return { status: 301 }
  }
})
```

```js
// child.js
import { interceptClientRequest } from '@mswjs/interceptors/lib/interceptors/ClientRequest'
import { createRemoteInterceptor } from '@mswjs/interceptors/remote'

// Intercepts requests according to the given modules
// and signals the parent process to send back the mocked response.
const interceptor = createRemoteInterceptor({
  modules: [interceptClientRequest]
})
```

## Limitations & caveats

- The child process must be spawned with the "ipc" `stdio` value provided. This is required for the IPC (inter-process communication) to work in Node.js (for `process.send` to be defined).
- The request reference (the second `resolver` argument) cannot be reliably serialized and passed via IPC (serializing classes isn't a good idea), so it's `undefined` when using the remote interception. 

## Roadmap

- [ ] Try constructing a request reference coerced from the known information (second argument for the `resolver` function).

### DIFF
--- a/src/remote.ts
+++ b/src/remote.ts
@@ -1,13 +1,15 @@
-import { ChildProcess } from 'child_process'
+import { ChildProcess, Serializable } from 'child_process'
 import { Headers } from 'headers-utils'
-import { Serializable } from 'node:child_process'
+import { StrictEventEmitter } from 'strict-event-emitter'
 import {
   createInterceptor,
   InterceptorApi,
+  InterceptorEventsMap,
   InterceptorOptions,
   IsomorphicRequest,
   Resolver,
 } from './createInterceptor'
+import { toIsoResponse } from './utils/toIsoResponse'
 
 type ProcessEventListener = (message: Serializable, ...args: any[]) => void
 
@@ -15,6 +17,8 @@ export type CreateRemoteInterceptorOptions = Omit<
   InterceptorOptions,
   'resolver'
 >
+
+export type RemoteResolverApi = Pick<InterceptorApi, 'on'>
 
 export interface CreateRemoteResolverOptions {
   process: ChildProcess
@@ -107,7 +111,9 @@ spawn('node', ['module.js'], { stdio: ['ipc'] })\
  */
 export function createRemoteResolver(
   options: CreateRemoteResolverOptions
-): void {
+): RemoteResolverApi {
+  const observer = new StrictEventEmitter<InterceptorEventsMap>()
+
   const handleChildMessage: ProcessEventListener = async (message) => {
     if (typeof message !== 'string') {
       return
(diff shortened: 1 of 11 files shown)


src/remote.ts
```diff
@@ -0,0 +1,145 @@
...
+    )
+  }
+
+  let handleParentMessage: ProcessEventListener
+
+  const interceptor = createInterceptor({
+    ...options,
+    resolver(request) {
+      const serializedRequest = JSON.stringify(request)
+      process.send?.(`request:${serializedRequest}`)
```
JacobParis, line 68: Should be able to drop this condition now that the error message is in place

```suggestion
      process.send(`request:${serializedRequest}`)
```

src/utils/uuid.ts
```diff
@@ -0,0 +1,7 @@
+export function uuidv4(): string {
```
JacobParis, line 1: I'd recommend just installing the `uuid` package on npm, this script has been [complained about on Twitter](https://twitter.com/stijnveilig/status/1409120079648808962?s=20) by infosec folks for reasons I don't fully understand

src/remote.ts
kettanaito: Logically—yes, but TypeScript still treats this `send` as potentially undefined. I suppose TS treats this closure as a runtime expression that doesn't guarantee that `process.send` has been checked before (or it can be a TS bug). 

So we can either use optional chaining or add another `typeof process.send` check here (which is redundant). Do you know about another option we could try?

src/utils/uuid.ts
kettanaito: That's an interesting take. I've asked the author to elaborate on the precise reasons the snippet is unrecommended. I don't mind installing an extra dependency, but I like understanding things first.

PR #136, opened 2021-07-03
Supports gzip content-encoding when retrieving the IncomingMessage body

### LINKED ISSUE #109: node interceptors corrupt incoming binary bodies

thomas-p-wilson:
## GitHub

- Fixes #109 

## Changes

- `getIncomingMessageBody` will now handle the gzip compression to prevent the corruption of compressed response bodies. 
- Adds previously missing unit tests for the `getIncomingMessageBody` function.

### DIFF
--- a/src/interceptors/ClientRequest/utils/getIncomingMessageBody.ts
+++ b/src/interceptors/ClientRequest/utils/getIncomingMessageBody.ts
@@ -1,12 +1,19 @@
 import { IncomingMessage } from 'http'
+import { Stream } from 'stream'
+import * as zlib from 'zlib'
 
 export function getIncomingMessageBody(res: IncomingMessage): Promise<string> {
   let responseBody = ''
+  let stream: Stream = res
+
+  if (res.headers['content-encoding'] === 'gzip') {
+    stream = res.pipe(zlib.createGunzip())
+  }
 
   return new Promise((resolve, reject) => {
-    res.once('error', reject)
-    res.on('data', (chunk) => (responseBody += chunk))
-    res.once('end', () => {
+    stream.once('error', reject)
+    stream.on('data', (chunk) => (responseBody += chunk))
+    stream.once('end', () => {
       resolve(responseBody)
     })
   })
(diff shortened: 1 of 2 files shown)


src/interceptors/ClientRequest/utils/getIncomingMessageBody.ts
```diff
@@ -1,12 +1,20 @@
+import * as zlib from 'zlib'
 import { IncomingMessage } from 'http'
+import { Stream } from 'stream'
 
 export function getIncomingMessageBody(res: IncomingMessage): Promise<string> {
   let responseBody = ''
+  let stream: Stream = res
+
+  if (res.headers['content-encoding'] === 'gzip') {
+    const gzip = zlib.createGunzip()
```
kettanaito, line 10: What implications do you see if we adopt your previous suggestion:

```js
let responseBody = Buffer.from('')
```

I'm hesitant to bring gzip handling to this method, it may be a stretch for its responsibilities.

thomas-p-wilson:
Using a buffer alone will not solve the problem, as at the end we'd have to
convert it back to a string anyway, and it's that string conversion that is
lossy, and prevents us from unzipping later on.

String concatenation and buffer concatenation will both yield the same
result.

kettanaito:
Thank you for the explanation, @thomas-p-wilson!

kettanaito (approved):
I've only added some cosmetic changes like test case names, otherwise, the change is great.

Thank you, @thomas-p-wilson, for addressing this in a timely manner. 

We'd love to see more contributions from you, so let us know if there are any open issues you'd be interested in tackling.

thomas-p-wilson:
Hey Artem!

Happy to help. I'll troll through the issue list in the coming days. I
really appreciate how this library has simplified my integration testing.
Happy to continue contributing!

Issue #138, opened 2021-07-09
Intercepting all requests


JohnPhamous:
Hello,

I'm trying to intercept all network requests on a web app. I'd like to record everything that shows up in the Networks tab in the browser developer tools (except for web sockets).

I've added `interceptors` to my web app:

```tsx
const interceptor = createInterceptor({
    modules: browserInterceptors,
    resolver(req, ref) {
        console.log(req.url);

        console.log(ref);
    },
});

interceptor.apply();
```

I have 2 buttons that make requests:

```tsx
<button
    onClick={async () => {
        const data = await fetch(
            'https://pokeapi.co/api/v2/pokemon/1'
        );
        const json = await data.json();
        console.info({ json });
    }}
>
    Fetch Request
</button>
<button
    onClick={async () => {
        const oReq = new XMLHttpRequest();
        oReq.open('GET', 'https://pokeapi.co/api/v2/pokemon/1');
        oReq.send();
    }}
>
    XHR
</button>
```

## Problems

1. The XHR request interceptor is working. The `fetch` isn't. I'm getting this error for `fetch`: `TypeError: Failed to execute 'json' on 'Response': body stream already read`
   - Looks like #113 is related?
2. Other network requests that are not using XHR are not being intercepted (should they?).

## Things I've checked

1. Using the `browserInteceptors` module, `fetch` and `XMLHttpRequest` are being shimmed. I checked this by running `fetch` and `XMLHttpRequest` in the console and the output was not `XMLHttpRequest() { [native code] }`.

kettanaito:
Hey, @JohnPhamous. 

I highly recommend using [MSW](https://github.com/mswjs/msw) for request interception in both browser and Node.js. MSW uses Service Worker API that intercepts requests regardless of the request client in the browser, so that would be an ideal solution for your monitoring tool. 

The "interceptors" library is a backbone for the Node.js support in MSW and although it can be used standalone, MSW provides a much better experience in a browser due to the usage of the Service Worker.

**You can follow this [step-by-step tutorial](https://mswjs.io/docs/getting-started/install) to set up the requests interception in the browser**. 

---

That being said, the "body stream already read" issue does seem related to #113. It looks like we need to clone the response before reading its body. Please, would you be interested in providing a pull request to fix that?

> Other network requests that are not using XHR are not being intercepted (should they?).

What kind of network requests are you referring to? Can you give an example? 

[...]

PR #150, opened 2021-08-29
Fixes Fetch bug on response not being cloned

### LINKED ISSUE #113: response (IncomingMessaging) has been read by interceptor and make downstream library unreadable
I am using Axios to get the raw stream, but Interceptors read the stream in `utils/getIncomingMessageBody`, so that the downstream http client cannot read the stream anymore. This can be a side effect and ideally that stream should be piped and not touch the original stream.

wsantos-conekta:
Hey, I was getting an error on fetch interceptor, especially when webpack makes the hot reload. This is the error:

`[HMR] Update failed: TypeError: Failed to execute 'json' on 'Response': body stream already read`

I noticed that the fetch response is being read inside `normalizeFetchResponse` function, so... It's required to clone the response before passing it to the function, That wont touch the real response.

Here is a Stackoverflow thread: https://stackoverflow.com/questions/40497859/reread-a-response-body-from-javascripts-fetch

## GitHub

- May fix #113 and #138

### DIFF
--- a/src/interceptors/fetch/index.ts
+++ b/src/interceptors/fetch/index.ts
@@ -58,12 +58,13 @@ export const interceptFetch: Interceptor = (observer, resolver) => {
     debug('no mocked response found, bypassing...')
 
     return pureFetch(input, init).then(async (response) => {
-      debug('original fetch performed', response)
+      const clonedResponse = response.clone()
+      debug('original fetch performed', clonedResponse)
 
       observer.emit(
         'response',
         isoRequest,
-        await normalizeFetchResponse(response)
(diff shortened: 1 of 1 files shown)


kettanaito:
Hey, @wsantos-conekta. Thank you for working on a fix for this!

Your changes look great. One thing we'd need is to add a test to catch this. Please, would you be interested in doing so?

kettanaito:
Thank you for tackling this, @wsantos-conekta! I've merged a similar fix with the tests in https://github.com/mswjs/interceptors/pull/152.

PR #151, opened 2021-08-30
response clone

### LINKED ISSUE #113: response (IncomingMessaging) has been read by interceptor and make downstream library unreadable

SCWR:
Copy response passed to trigger event
may be fix #113 #138

### DIFF
--- a/src/interceptors/fetch/index.ts
+++ b/src/interceptors/fetch/index.ts
@@ -63,7 +63,7 @@ export const interceptFetch: Interceptor = (observer, resolver) => {
       observer.emit(
         'response',
         isoRequest,
-        await normalizeFetchResponse(response)
+        await normalizeFetchResponse(response.clone())
       )
       return response
     })


kettanaito:
Hey, @SCWR. Thank you for providing a fix for that!

This appears to be a duplicate of #150. If you don't mind, I'd close this pull request and continue the discussion in the said one. The changes look great, the one thing to add would be a test to catch this issue (and any future regressions). If writing such a test sounds like something you'd be interested in doing, we'd highly appreciate your help. You can see the list of existing tests in the repo and collaborate with @wsantos-conekta (the author of the linked pull request) to come up with the test.

SCWR:
It has already been there, it is a bit awkward. This reminds me that the next time I need to look at whether the PR is repeated. Your suggestion I need to understand the test. Of course I am very interested.

kettanaito:
A good test always resembles the usage. In the case of this issue, the problem occurs when trying to access the response body, as I understand. I believe we can model the test to do just that:

1. Configure interceptors.
2. Make a request for a matching URL.
3. Attempt to do `await res.text()` in the test.
4. Assert that nothing is thrown and the response text is extracted by the "user" (our test).

PR #165, opened 2021-10-29
fix(NodeClientRequest): clone "response" to allow reading its body twice

### LINKED ISSUE #161: Prevent blocking the IncomingMessage stream in "getIncomingMessageBody"
The current implementation of `getIncomingMessageBody` blocks the given `res` so it can't be read afterward:

https://github.com/mswjs/interceptors/blob/adc15848cb43e691dacf052792cb386bcad881c3/src/interceptors/ClientRequest/utils/getIncomingMessageBody.ts#L5-L26

This results in any additional `res.on('data')` never being invoked because the stream is paused. We need to resume the stream (ideally, clone beforehand) to allow any consequent listeners to work as expected.

kettanaito:
- Fixes https://github.com/mswjs/msw/issues/941
- Fixes #161

## Changes

- Uses an internal `cloneIncomingMessage` function that pipes a given IncomingMessage to a `PassThrough` stream. It also copies all the message's properties and symbols to ensure an identical copy for whichever code may rely on it (piped stream does not have response properties like `statusCode` or `rawHeaders`).
- Fixes `getIncomingMessageBody` test so they won't silently swallow assertion failures.
- Adds a regression test to capture the Stripe use-case (see https://github.com/mswjs/msw/issues/941)

### DIFF
--- a/src/interceptors/ClientRequest/NodeClientRequest.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.ts
@@ -23,6 +23,7 @@ import {
   ClientRequestWriteArgs,
   normalizeClientRequestWriteArgs,
 } from './utils/normalizeClientRequestWriteArgs'
+import { cloneIncomingMessage } from './utils/cloneIncomingMessage'
 
 export type Protocol = 'http' | 'https'
 
@@ -142,7 +143,7 @@ export class NodeClientRequest extends ClientRequest {
       this.log('original request aborted!')
     })
 
-    this.once('response', async (response) => {
+    this.once('response-internal', async (response: IncomingMessage) => {
       const responseBody = await getIncomingMessageBody(response)
       this.log(response.statusCode, response.statusMessage, responseBody)
       this.log('original response headers:', response.headers)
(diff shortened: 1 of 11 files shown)


src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -163,6 +164,38 @@ export class NodeClientRequest extends ClientRequest {
...
+         * Clone the response object when emitting the "response" event.
+         * This prevents the response body stream from locking
+         * and allows reading it twice:
+         * 1. Internal "response" event from the observer.
+         * 2. Any external response body listeners.
+         * @see https://github.com/mswjs/interceptors/issues/161
+         */
+        const response = data[0] as IncomingMessage
+        const firstClone = cloneIncomingMessage(response)
+        const secondClone = cloneIncomingMessage(response)
```
kettanaito, line 184: Clone the original `IncomingMessage` (response) into two `PassThrough` streams so they can be read independently by:

- Internal `response` event emitted by the `observer` (contains `response.body` from `getIncomingMessageBody`);
- Any external `response` event listeners on `ClientRequest`.

src/interceptors/ClientRequest/utils/cloneIncomingMessage.ts
```diff
@@ -0,0 +1,37 @@
...
+      propertyName
+    )
+
+    Object.defineProperty(stream, propertyName, {
+      ...propertyDescriptor,
+      value: message[propertyName],
+    })
+  }
+
+  Object.defineProperty(stream, IS_CLONE, {
```
kettanaito, line 31: Mark the cloned `IncomingMessage` for introspection purposes.

kettanaito:
I've resolved the `test/events/request.test.ts` pending issue by using `http.request` directly. Getting rid of `httpRequest` utility meant that it no longer reads the response body, which caused the test to hang. That is an issue, I suppose, but it's not caught by any response reading integration tests currently present. I conclude it to be a test setup issue.

PR #170, opened 2021-11-06
fix(IsomorphicRequest): add "credentials" property

### LINKED ISSUE #18: Handle request "credentials" for browser-like environments
## GitHub

- Related to https://github.com/mswjs/msw/issues/189

kettanaito:
## GitHub

- Closes #18 
- Fixes https://github.com/mswjs/msw/issues/967

## Changes

- Adds a new "credentials" property on the `IsomorphicRequest` object.
- XMLHttpRequest: Coerces `this.withCredentials` option to "include" (`true`) or "omit" (`false`).
- ClientRequest: Always sets `request.credentials` to "omit" (no way to specify the behavior).
- Improves in-browser request utilities for `fetch` and `XMLHttpRequest`. **Removes direct assertions from the `*.runtime.js` files.** Moves assertions to the test's context. 

## Roadmap

- [x] Add integration tests.
- [x] Update the README to list the newly added "credentials" property on the `IsomorphicRequest`. Explain its values.


src/createInterceptor.ts
```diff
@@ -19,6 +19,7 @@ export interface IsomorphicRequest {
   url: URL
   method: string
   headers: Headers
+  credentials: 'omit' | 'include' | 'same-origin'
```
kettanaito, line 22: I'm preserving the "same-origin" option because the `fetch` interceptor is designed for the browser where such an option is legitimate and possible. It's a no-op in Node.js.

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -292,6 +292,7 @@ export class NodeClientRequest extends ClientRequest {
       id: uuidv4(),
       url: this.url,
       method: this.options.method || 'GET',
+      credentials: 'omit',
```
kettanaito, line 295: No way to control the credentials behavior in `ClientRequest`.

kettanaito:
"credentials" is generally a concern of `window.fetch` and any browser-like environments that polyfill it (via `XMLHttpRequest`). This change primarily affects those two interceptors.

src/createInterceptor.ts
```diff
@@ -21,6 +21,11 @@ export interface IsomorphicRequest {
   url: URL
   method: string
   headers: Headers
+  /**
+   * The value of the request client's "credentials" option
+   * or a compatible alternative (i.e. `withCredentials` for `XMLHttpRequest`).
+   * Always equals to "omit" in Node.js.
+   */
   credentials: RequestCredentials
```
kettanaito, line 29: We do not document `IsomorphicRequest` in README. I'm adding this short description directly on the property.

crutch12:
Are you going to release it?

kettanaito:
@crutch12, yes, we will. You can use local resolution for `0.13.0` for now, there are no breaking changes between 0.12 and 0.13.

PR #175, opened 2021-11-09
chore: migrate to github actions

kettanaito:
Uses GitHub Actions and workflows instead of CircleCI.


kettanaito:
The repository is already using GitHub actions.

PR #178, opened 2021-11-09
fix(NodeClientRequest): suppress connection errors for mocked requests

### LINKED ISSUE #168: ClientRequest: Prevent establishing the connection until in bypass context
Right now the `NodeClientRequest` instance always establishes the connection to the given host. If there's any sort of delay during the mocked response processing, that established connection fails with:

```sh
Error: connect ECONNREFUSED 127.0.0.1:80
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1159:16) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 80
}
```

Can be reproduced by putting a dummy timeout promise anywhere _after_ the `resolver` function call in the `end` method:

```ts
await new Promise((resolve) => setTimeout(resolve, 1500))
```

## Partial confirmation

If the `emit` method is adjusted to never propagate the "error" event, the issue is resolved. That's not the solution, of course, request errors must be propagated. It only shows that the unreliable response processing is caused by the request instance throwing the `ECONNREFUSED` error during the connection. 

## Solution

1. Analyze the `_http_client.js` module from Node.js.
2. Find the moment where it establishes the connection (maybe in the `OutgoingMessage`).
3. Stub that moment, allowing connections only for bypassed requests.

## Socket journey in `http.ClientRequest`

1. The `Agent` is [assigned in the ClientRequest constructor](https://github.com/nodejs/node/blob/9c405f2591f5833d0247ed0fafdcd68c5b14ce7a/lib/_http_client.js#L129-L144).
2. The [connection is initiated](https://github.com/nodejs/node/blob/9c405f2591f5833d0247ed0fafdcd68c5b14ce7a/lib/_http_client.js#L303-L332) in the ClientRequest constructor.
3. If the `this.agent ` is set, it [adds a request to the agent](https://github.com/nodejs/node/blob/9c405f2591f5833d0247ed0fafdcd68c5b14ce7a/lib/_http_agent.js#L242).
4. If `this.agent` is not set, ClientRequest attempts to [create a connection from options](https://github.com/nodejs/node/blob/9c405f2591f5833d0247ed0fafdcd68c5b14ce7a/lib/_http_client.js#L320-L324).
5. Eventually, ClientRequest calls [onSocket method](https://github.com/nodejs/node/blob/9c405f2591f5833d0247ed0fafdcd68c5b14ce7a/lib/_http_client.js#L783) when the socket connection is established.
6. Then, it [propagates](https://github.com/nodejs/node/blob/9c405f2591f5833d0247ed0fafdcd68c5b14ce7a/lib/_http_client.js#L783) to the [`tickOnSocket`](https://github.com/nodejs/node/blob/9c405f2591f5833d0247ed0fafdcd68c5b14ce7a/lib/_http_client.js#L722), where ClientRequest sets a bunch of properties related to the successful socket connection.

kettanaito:
## GitHub

- Fixes #168 

## Changes

- Captures and suppresses certain ClientRequest errors (i.e. `ECONNREFUSED`). This prevents the request instance from being destroyed. 
- Replays the first captured error while constructing the ClientRequest once using the "bypass" response source.

## Motivation

I've concluded my initial attempt to provide a custom `Agent.createConnection` to be unnecessary. While it's possible to prevent socket connection that way, it's problematic to let the ClientRequest believe the connection was made (otherwise it won't function properly). Preventing the socket connection altogether is also undesired as it effectively means that we'd have to _recreate_ that connection from scratch once using the "bypass" response source. This is precisely what makes Node.js request interception slow when using MSW (the deferred declaration of the "pure" ClientRequest instance).

With this implementation, we allow the ClientRequest instance to establish (lookup/validate) the connection as usual, but only _suppress any errors it may have_. We ignore those errors when the request has a mocked response (note that IP validation and all underlying error handling is still preserved, you must perform valid requests). If there's no mocked response, the first captured error is replayed (re-emitted). 

> I'm using the first captured error because in the normal ClientRequest flow the first error emitted will be the first error destroying the request instance, preventing any further errors from being emitted.

### DIFF
--- a/src/interceptors/ClientRequest/NodeClientRequest.test.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.test.ts
@@ -1,4 +1,8 @@
+/**
+ * @jest-environment node
+ */
 import { EventEmitter } from 'events'
+import * as express from 'express'
 import { ServerApi, createServer } from '@open-draft/test-server'
 import { NodeClientRequest } from './NodeClientRequest'
 import { getIncomingMessageBody } from './utils/getIncomingMessageBody'
@@ -11,10 +15,20 @@ interface ErrorConnectionRefused extends NodeJS.ErrnoException {
 
 let httpServer: ServerApi
 
+function waitFor(duration: number): Promise<void> {
+  return new Promise((resolve) => {
+    setTimeout(resolve, duration)
+  })
+}
+
 beforeAll(async () => {
   httpServer = await createServer((app) => {
     app.post('/comment', (req, res) => {
       res.status(200).send('original-response')
+    })
+
+    app.post('/write', express.text(), (req, res) => {
+      res.status(200).send(req.body)
     })
   })
 })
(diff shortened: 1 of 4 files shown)


kettanaito:
## Noteworthy mentions

The request will keep pending indefinitely if you don't call the `.end()` method, when applicable. Since the library can know if there's any mocked response only in the `.end()` method (when the request's body is complete), we suppress any errors until then. 

That shouldn't be an issue, as not calling `.end()` means creating a malformed request. 

I'd like to check how the native `ClientRequest` behaves if:

- Connecting to a non-existing hostname;
- Attaching the "error" listener;
- But not calling `.end()`.

We should have identical behavior.

kettanaito:
While this does suppress the errors, there's still an issue that the socket connection is, indeed, established. In particular, I have a suspicion that the request data is being written to the active socket connection, sending it to the server as it comes in. This may be the reason why one of the response patching tests fails in MSW using this version of the library.

kettanaito:
The changes are now 100% compliant with MSW integration tests 🎉

Issue #159, opened 2021-10-17
Support Undici


tompeace:
[undici](https://undici.nodejs.org/) is a node.js library written by some core node.js contributors. It offers many great features for making fast http request that largely outperform the native `http` & `https` modules by using a customer http parser called `llhttp`.

MSW does not work with this library as it does not  mock any of the modules covered in `./src/interceptors/ClientRequest`.

undici does offer its own mocking mechanism, but I love MSW's api and am wondering if there are any plans to add an interceptor for this in the future?

For context my use case is mocking graphql `apollo-datasource-http` requests.

Thanks!

chentsulin:
Hi @kettanaito @tompeace,
I created a PR #180 to include the undici tests for [undici.request](https://undici.nodejs.org/#/?id=undicirequesturl-options-promise) and [undici.fetch](https://undici.nodejs.org/#/?id=undicifetchinput-init-promise).  

I'm not sure I'm doing everything right, but I might be a good starting point.

kettanaito:
Hey, @chentsulin! Thank you for your work on that. I'll take a look at the pull request as soon as I have time. I'm sure it'll act as a great foundation to introduce the Undici support itself.

SupremeTechnopriest:
Has anyone made any headway on this interceptor?

kettanaito:
Hey, @SupremeTechnopriest. No, there hasn't been any progress on this. We've started by adding some failing tests in https://github.com/mswjs/interceptors/pull/180 but the implementation is still missing. 

Would you be interested in working on this? I'd be happy to support you with code reviews and discussions about bringing Undici support to this library.

SupremeTechnopriest:
Thinking I will take a swing at it.  Has the tests been merged?

SupremeTechnopriest:
No worries, I just pulled the tests into my fork.  I Imagine you want to try to intercept as close to the core of undici as possible.  Maybe the connection from the `tls` and `net` module.  I will start there and work my way up stack.

kettanaito:
@SupremeTechnopriest sounds like a good starting point. I've noticed that Undici uses a mocked `Agent` for their recommended mocking setup. That's a nice surface indeed, but I'm afraid it lacks some vital info about the request. 

[...]

PR #180, opened 2021-11-10
test(undici): add interceptor tests for undici

chentsulin:
See #159.

### DIFF
--- a/test/intercept/undici/undici.fetch.test.ts
+++ b/test/intercept/undici/undici.fetch.test.ts
@@ -1,0 +1,190 @@
+/**
+ * @jest-environment node
+ */
+import { fetch as undiciFetch } from 'undici'
+import { RequestHandler } from 'express'
+import { ServerApi, createServer } from '@open-draft/test-server'
+import { createInterceptor } from '../../../src'
+import { IsomorphicRequest } from '../../../src/createInterceptor'
+import { interceptClientRequest } from '../../../src/interceptors/ClientRequest'
+
+let requests: IsomorphicRequest[] = []
+let httpServer: ServerApi
+
+const interceptor = createInterceptor({
+  modules: [interceptClientRequest],
+  resolver(request) {
+    requests.push(request)
+  },
+})
+
+beforeAll(async () => {
+  httpServer = await createServer((app) => {
+    const handleUserRequest: RequestHandler = (req, res) => {
+      res.status(200).send('user-body').end()
+    }
+
+    app.get('/user', handleUserRequest)
+    app.post('/user', handleUserRequest)
+    app.put('/user', handleUserRequest)
+    app.delete('/user', handleUserRequest)
+    app.patch('/user', handleUserRequest)
+    app.head('/user', handleUserRequest)
+  })
+
+  interceptor.apply()
+})
+
+afterEach(() => {
+  requests = []
+})
+
+afterAll(async () => {
+  interceptor.restore()
+  await httpServer.close()
+})
+
+test('intercepts a HEAD request', async () => {
+  const url = httpServer.https.makeUrl('/user?id=123')
+  await undiciFetch(
+    url,
+    {
+      method: 'HEAD',
+      headers: {
+        'x-custom-header': 'yes',
+      },
+    },
+  )
+
+  expect(requests).toHaveLength(1)
+
+  const [request] = requests
+  expect(request.method).toEqual('HEAD')
+  expect(request.url).toBeInstanceOf(URL)
+  expect(request.url.href).toEqual(httpServer.https.makeUrl('/user?id=123'))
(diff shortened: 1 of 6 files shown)


test/intercept/undici/undici.fetch.test.ts
```diff
@@ -0,0 +1,190 @@
+/**
```
kettanaito, line 1: We can put these test suites into `test/integration` folder. It's there to test this library's integration with other libraries (like `got`), as those may have certain corner-cases to account for.

chentsulin: As I see it, `undici` is a replacement for the old `http` and `https` module, and many libraries [include got](https://github.com/sindresorhus/got/issues/1419) may use it instead of `http` in the future. So, I suppose this test should be placed nearby the tests of the `http` interceptor.

kettanaito: @chentsulin, well, `undici` is still a _third-party library_. Until it's a part of the Node.js standard library, we should treat it as third-party integration.

kettanaito:
I've rebased your feature branch, as well as adjusted it to the latest test structure refactoring (#182). Your newly added undici tests are now under `test/third-party/undici`.

test/intercept/undici/undici.fetch.test.ts
chentsulin: Fair enough. We can rethink about that by then.

kettanaito:
Thank you for adding these tests!

The support for unidici will roughly follow these steps:

1. Research how `undici` performs requests internally. See what modules it utilizes, what modules can be extended to support interception.
2. Discuss the findings, assess the support strategy.
2. If agreed, create a new `interceptors/undici` interceptor with the barebone implementation to cover the tests you've added.
3. Ensure sufficient test coverage, merge and release the change. 

If you have time, feel free to tackle the first step. Gathering knowledge often takes the most time, and I could definitely use your expertise in that. Thank you!

chentsulin:
In my early research, undici has its own mocking mechanism:

```js
import { MockAgent, setGlobalDispatcher, } from 'undici'

const mockAgent = new MockAgent();

setGlobalDispatcher(mockAgent);

// Provide the base url to the request
const mockPool = mockAgent.get('http://localhost:3000');

// intercept the request
mockPool.intercept({
  path: '/bank-transfer',
  method: 'POST',
  headers: {
    'X-TOKEN-SECRET': 'SuperSecretToken',
  },
  body: JSON.stringify({
    recepient: '1234567890',
    ammount: '100'
  })
}).reply(200, {
  message: 'transaction processed'
})
```

We should take advantage of this [setGlobalDispatcher](https://github.com/nodejs/undici#undicisetglobaldispatcherdispatcher) stuff to implement the interceptor.

chentsulin:
@kettanaito,
I don't have enough knowledge to experiment and finish all implementation details, but I do a naive and maybe buggy implementation for you to reference:
https://github.com/mswjs/interceptors/commit/a7e21d1a929e0914cdd472e73ac68bdce676c75a

That's what i learned:

- `setGlobalDispatcher` and `setGlobalDispatcher` are prefect for mocking and restoring.
- undici classes are written in ES2015, so set `compilerOptions.target` to `ES5` won't work.
- The `dispatch` method is synchronous, so have to find a way to inject asynchronous body parsing and response resolving.
- Relying on `dispatch` and `handler` seem to be right, because undici use them: https://github.com/nodejs/undici/blob/f951002dfb86f502f1129588c021a26d7f2a0afb/lib/mock/mock-utils.js#L168-L174

jtlopez18:
@chentsulin Thanks for sharing your implementation, it was super helpful.  I have been playing with it and seeing an issue where undici expects @types/node to be at least 14.18 (this has the Blob type from buffer) in order to import the Dispatcher.DispatchHandlers type for the resolver.

However, updating this seems to cause issues with other intercepters, such as XML and http.get.  Did you run into anything like that?  How did you resolve?


> @kettanaito, I don't have enough knowledge to experiment and finish all implementation details, but I did a naive and maybe buggy implementation for you to refer: [a7e21d1](https://github.com/mswjs/interceptors/commit/a7e21d1a929e0914cdd472e73ac68bdce676c75a)

kettanaito:
Thanks for preparing a reference implementation, @chentsulin! 

I'm not sure to which extend we can reuse it, as with interceptors we're aiming at low-level request interception. Relying on `MockAgent` is a high-level thing, meaning we're going deeper into implementation specifics rather than relying on Node.js primitives. 

I'd love to start this feature from research. We should first understand what kind of requests does undici make (it _has_ to make a `Socket` connection at least, there are no other ways to request things in Node.js) and, ideally, model our interceptor around that. 

I don't think I'll have time to work on this in the foreseeable future, so anybody is welcome to grab the research and implementation if they find this interesting.

Issue #163, opened 2021-10-27
Prevent nested "observer" from emitting events


kettanaito:
## Current behavior

XMLHttpRequest in Node.js is often polyfilled via `ClientRequest`. This means that when an XHR request is mocked, both XHR and ClientRequest interceptors will be working. As each interceptor is designed standalone, each emits the `observer` events separately. This results in events like "request" to be emitted _twice_ for an XHR request in Node.js. 

## Expected behavior

Internal `observer` events must be emitted once. The library must understand when it nests interceptors and allow only the root interceptor to emit events.

Issue #121, opened 2021-05-26
xhr.upload is null


jameskerr:
I've got code that looks like this:

```js
const xhr = new XMLHttpRequest()
xhr.upload.addEventListener("progress", (e) => {
  // ...
})
```
When I run my test I get this error:
```
    TypeError: Cannot read property 'addEventListener' of null

      40 |         const xhr = new XMLHttpRequest()
      41 |         console.log(xhr)
    > 42 |         xhr.upload.addEventListener("progress", (e) => {
```

That console.log reveals that the upload property of the class is set to null.

```
   XMLHttpRequestOverride {
      upload: null
    }
```

If someone points me in the right direction, I'd be open to filling in this gap.

kettanaito:
Hey, @jameskerr. Thank you for raising this up.

The `upload` property of the XMLHttpRequest override class is currently set to `null` in the constructor and isn't modifier since then:

https://github.com/mswjs/interceptors/blob/d693b1ee90479f7e6bd48b5d9a08b33b663d2139/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L134

According to the documentation, `XMLHttpRequest.upload` is an `XMLHttpRequestEventTarget` with the support for certain progress events. My initial thought is to introduce a custom event handling for the `upload` property, similar to how the regular `xhr.addEventListner()` events are collected and executed at the right stage of the request:

https://github.com/mswjs/interceptors/blob/d693b1ee90479f7e6bd48b5d9a08b33b663d2139/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L463-L481

https://github.com/mswjs/interceptors/blob/d693b1ee90479f7e6bd48b5d9a08b33b663d2139/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L277

https://github.com/mswjs/interceptors/blob/d693b1ee90479f7e6bd48b5d9a08b33b663d2139/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L159-L183

What do you think about this approach?

kettanaito:
We may want to put this issue on hold. It's possible we'll rewrite the XHR interception from the ground-up to extend the native XMLHttpRequest class instead of patching it (similar to what we did with `ClientRequest` in #164).

kettanaito:
[...]

PR #184, opened 2021-11-12
refactor(XMLHttpRequest): extend native XMLHttpRequest class

### LINKED ISSUE #103: XMLHttpRequest: readystatechange listener is called twice for unhandled requests
## Code

https://github.com/mswjs/node-request-interceptor/blob/b19c82db06fc9c92a8606bd02ea967a1391d7837/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L358-L364

See the workaround test:

https://github.com/mswjs/node-request-interceptor/blob/b19c82db06fc9c92a8606bd02ea967a1391d7837/test/compliance/XMLHttpRequest/events-order.test.ts#L75-L88

## Expected behavior

The order of the XMLHttpRequest events must be this:

```
['loadstart', 1],
['readystatechange', 2],
['readystatechange', 4],
['load', 4],
['loadend', 4],
```

There must be no duplicate `['readystatechange', 4]` event.

## Why this is happening?

XMLHttpRequest polyfill from JSDOM executes the attached `readystatechange` listeners as a part of XMLHttpRequest life-cycle, _which is correct_. However, somehow the way that the listener is attached makes it being called twice.

### LINKED ISSUE #183: Rewrite XMLHttpRequest override to extend XMLHttpRequest
Similar to what we did with `ClientRequest` in #164, we should rewrite the XMLHttpRequest interceptor to extend the native `window.XMLHttpRequest` class instead of implementing it:

https://github.com/mswjs/interceptors/blob/1d35db531d14a7f95126a382b005451af927924c/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L49

Currently, we have what effectively is a partial polyfill, which may result in all sorts of issues due to certain pieces of XHR logic simply not being present (#121, #70). If we introduce extension, we will leverage the entire XHR logic underneath, overriding only the methods necessary.

### LINKED ISSUE #187: XMLHttpRequest: Support "upload" events
The `upload` property on the XMLHttpRequest returns a `XMLHttpRequestUpload` instance, which is similar to an event emitter. The consumer may use the `upload` instance to attach listeners on the _request body_ upload progress.

```js
req.open('POST', '/upload')
req.upload.addEventListener('progress', (event) => {
  if (event.lengthComputable) {
    console.log('%d / %d', event.loaded, event.total)
  } 
})
```

We should enable the `upload` events for both mocked and bypassed requests. 

## Event order

The upload events are called in the following order, given request body:

1. `loadstart`, request body is known, the `lengthComputable` value on the `ProgressEventInit` is set to a boolean representing if there's any upload work to be done (request has body).
2. `progress`, request body is known and read, the progress event contains `loaded` and `total` of transferred request body bytes.
3. `load`.
4. `loadend`, the upload is over.

kettanaito:
- Closes #183 
- Fixes #103 
- Closes #187 
- Potentially solves #121 (needs tests)
- Potentially solves #163 (needs tests)

## Changes

- Adds a new `NodeXMLHttpRequest` class that extends the native `XMLHttpRequest` instead of replacing it.
- Adjusts the tests accordingly. Fixes test issues that were found once we execute the underlying request module. 

## Roadmap

- [ ] Fix the TypeScript's `target` change issue.
- [ ] Add support for different `data` types in `.send()`, add tests.
- [x] Add a test for `xhr.upload` value when performing mocked/bypass upload.
- [x] Ensure that no connection is made while in the "mock" response source (XHR polyfill in JSDOM constructs a native `ClientRequest`instance now).


kettanaito:
## TypeScript `target` change regression

```sh
/home/circleci/release/lib/index.js:1
export * from './createInterceptor';
^^^^^^

SyntaxError: Unexpected token 'export'
    at wrapSafe (internal/modules/cjs/loader.js:1054:16)
    at Module._compile (internal/modules/cjs/loader.js:1102:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1158:10)
    at Module.load (internal/modules/cjs/loader.js:986:32)
    at Function.Module._load (internal/modules/cjs/loader.js:879:14)
    at Module.require (internal/modules/cjs/loader.js:1026:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (/home/circleci/release/test/features/remote/child.js:2:37)
    at Module._compile (internal/modules/cjs/loader.js:1138:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1158:10)
```

Setting the `target: "ES6"` was necessary to support class extension of `XMLHttpRequest` module. This change causes  some tests to fail.

kettanaito:
Extending native classes is prohibited in the browser. We cannot augment the native `window.XMLHttpRequest` and have the library transpiled with TypeScript at the same time. Sadly, this change is impossible as of now.

Issue #185, opened 2021-11-12
Modify requests


PaperStrike:
Hello!

I would like to propose a feature that will allow us to modify the sent requests. That is, apart from returning responses from the `resolver`, we should be able to return requests as well. In that case, the returned request should be treat as the real request to send. This allows us to intercept the network in all aspects.

My initial motivation is to bypass specific requests without exposing specific headers. I know there is a relevant #34, which suggests this is what should be done in downstream packages, but it seems this upstream interceptor don't have the ability to allow so. Take the `x-msw-pass` as an example, if we simply return a falsy value on this header's existence, the request will pass through, but be sent with this header. Allowing modifying headers would be a solution, but why not being able to modify the whole request?

Thanks!

kettanaito:
Hey, @PaperStrike. Thanks for suggesting this.

You can modify outgoing requests at the moment by accessing the second argument of the `resolver` function:

```js
createInterceptor({
  modules: [...],
  resolver(req, ref) {
    // use "ref" to operate on the raw request instance
  }
})
```

While `req` is an isomorphic request representation, the `ref` is a reference to the actual request instance. That instance is different depending on what made the request:

- `ClientRequest` if the request was issued by `http`/`https`;
- `XMLHttpRequest` if the request was issued by the respective class;
- `Request` if the request was issued by `fetch` (applicable only when used in a browser). 

Access that ref, and call the respective methods you wish to modify the request. 

It's worth mentioning that you should be careful modifying things that weren't set by you (like the `x-msw-bypass` header). This may break the functionality that depends on those things (such as life-cycle events if you remove the said header).

PaperStrike:
This isn't true at least for `fetch`. You can see the `pureFetch` isn't using the `ref`. So this is a bug?

https://github.com/mswjs/interceptors/blob/1d35db531d14a7f95126a382b005451af927924c/src/interceptors/fetch/index.ts#L61

PaperStrike:
@kettanaito Also the changes we can made in `ref` are _very_ limited. For [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request), most properties of it are `readonly`. Actually we can only change the `headers`.

And this seems not to be a "bug" that we can't change the request via `ref`, since it will be an breaking change if we start to support it - users may have already been using `ref`'s [body functions](https://developer.mozilla.org/en-US/docs/Web/API/Request#methods), which can only run once.

kettanaito:
[...]

PR #186, opened 2021-11-12
fix(fetch): use native ref to parse request props

PaperStrike:
The first argument passed to `fetch` may be a `Request` which will be used as the base of the second argument. E.g.,

```ts
const request = new Request('', { method: 'POST' });
fetch(request); // <- This is a POST fetch.
```

Parsing the request properties only from the second argument `init` will not cover this case.
We can parse from `ref` that is constructed from the native `Request` constructor here.

I haven't add test cases since I think it better to add after the test tool supports passing `Request`s.

### DIFF
--- a/src/interceptors/fetch/index.ts
+++ b/src/interceptors/fetch/index.ts
@@ -22,7 +22,7 @@ export const interceptFetch: Interceptor = (observer, resolver) => {
   window.fetch = async (input, init) => {
     const ref = new Request(input, init)
     const url = typeof input === 'string' ? input : input.url
-    const method = init?.method || 'GET'
+    const method = ref.method
 
     debug('[%s] %s', method, url)
 
@@ -30,8 +30,8 @@ export const interceptFetch: Interceptor = (observer, resolver) => {
       id: uuidv4(),
       url: new URL(url, location.origin),
       method: method,
-      headers: new Headers(init?.headers || {}),
-      credentials: init?.credentials || 'same-origin',
+      headers: new Headers(ref.headers),
+      credentials: ref.credentials,
       body: await ref.text(),
     }
     debug('isomorphic request', isoRequest)
--- a/src/interceptors/fetch/index.ts
+++ b/src/interceptors/fetch/index.ts
@@ -32,7 +32,7 @@ export const interceptFetch: Interceptor = (observer, resolver) => {
       method: method,
       headers: new Headers(request.headers),
       credentials: request.credentials,
-      body: await request.text(),
+      body: await request.clone().text(),
     }
     debug('isomorphic request', isoRequest)
     observer.emit('request', isoRequest)
@@ -58,7 +58,7 @@ export const interceptFetch: Interceptor = (observer, resolver) => {
 
     debug('no mocked response found, bypassing...')
 
-    return pureFetch(input, init).then(async (response) => {
+    return pureFetch(request).then(async (response) => {
       const cloneResponse = response.clone()
       debug('original fetch performed', cloneResponse)
 
--- a/src/interceptors/fetch/index.ts
+++ b/src/interceptors/fetch/index.ts
@@ -20,9 +20,9 @@ export const interceptFetch: Interceptor = (observer, resolver) => {
   debug('replacing "window.fetch"...')
 
   window.fetch = async (input, init) => {
-    const ref = new Request(input, init)
+    const request = new Request(input, init)
     const url = typeof input === 'string' ? input : input.url
-    const method = ref.method
+    const method = request.method
 
     debug('[%s] %s', method, url)
 
@@ -30,15 +30,15 @@ export const interceptFetch: Interceptor = (observer, resolver) => {
       id: uuidv4(),
       url: new URL(url, location.origin),
       method: method,
-      headers: new Headers(ref.headers),
-      credentials: ref.credentials,
-      body: await ref.text(),
+      headers: new Headers(request.headers),
+      credentials: request.credentials,
+      body: await request.text(),
     }
     debug('isomorphic request', isoRequest)
     observer.emit('request', isoRequest)
 
     debug('awaiting for the mocked response...')
-    const response = await resolver(isoRequest, ref)
+    const response = await resolver(isoRequest, request)
     debug('mocked response', response)
 
     if (response) {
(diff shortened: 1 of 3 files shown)


src/interceptors/fetch/index.ts
```diff
@@ -22,16 +22,16 @@ export const interceptFetch: Interceptor = (observer, resolver) => {
   window.fetch = async (input, init) => {
     const ref = new Request(input, init)
```
kettanaito, line 23: How about we call this `request`?

kettanaito (commented):
Thanks for opening this pull request. I'll help you with the tests on this.

src/interceptors/fetch/index.ts
PaperStrike: Done in https://github.com/mswjs/interceptors/pull/186/commits/3dee65e9c153623812575c32bbc23af6342ddc87

src/interceptors/fetch/index.ts
```diff
@@ -20,25 +20,25 @@ export const interceptFetch: Interceptor = (observer, resolver) => {
   debug('replacing "window.fetch"...')
 
   window.fetch = async (input, init) => {
-    const ref = new Request(input, init)
+    const request = new Request(input, init)
     const url = typeof input === 'string' ? input : input.url
```
PaperStrike, line 24: How about we parse with `request`?

PaperStrike:
The latter commit `fix(fetch): use exposed request in pureFetch` ([#549d960](https://github.com/mswjs/interceptors/pull/186/commits/549d9606db7674468c87ddf4ecabd427eb133a97)) is to resolve https://github.com/mswjs/interceptors/issues/185#issuecomment-967174889.

@kettanaito But wait, shouldn't we reconstruct a request from `isoRequest`? With the other two interceptors we can do this on the isomorphic request (see [mswjs/msw/src/utils/request/parseIsomorphicRequest.ts#L16](https://github.com/mswjs/msw/blob/fe6751f76108047fb27b16a875497693da1eb66d/src/utils/request/parseIsomorphicRequest.ts#L16)):

```js
request.headers.set('x-msw-request-id', requestId)
```

Reconstruct a request from `isoRequest` provides a consistent API.

PaperStrike:
At least reconstruct the `headers`?

kettanaito:
> But wait, shouldn't we reconstruct a request from isoRequest?

I'm not sure what you mean. The isomorphic request is meant to represent the original request. It cannot be used to reconstruct it, as it may not contain all the necessary options. The isomorphic request is _always_ derived from the original request, not the other way around. It's not mutable.

src/interceptors/fetch/index.ts
kettanaito: I don't think that's a good idea. We should keep the actual logic as close to the original request/options as possible, those being `input` and `init` the consumer provided when constructing fetch. Can we guarantee that the original options will be 100% the same after we provided them to the `Request` class? I think that's how `fetch` works internally, but it's hard to assume that, as we can't see its source code.

PaperStrike:
> I'm not sure what you mean. The isomorphic request is meant to represent the original request. It cannot be used to reconstruct it, as it may not contain all the necessary options. The isomorphic request is _always_ derived from the original request, not the other way around. It's not mutable.

I thought you were modifying the sent request via [mswjs/msw/src/utils/request/parseIsomorphicRequest.ts#L16](https://github.com/mswjs/msw/blob/fe6751f76108047fb27b16a875497693da1eb66d/src/utils/request/parseIsomorphicRequest.ts#L16) when using the other two interceptors. Whatever, I'm tired of checking the internal logic in `msw`. Let's focus on `@mswjs/interceptors` now.

src/interceptors/fetch/index.ts
PaperStrike: It is 100% the same according to [`fetch()` specification](https://fetch.spec.whatwg.org/#dom-global-fetch). You can see it uses the `Request` constructor internally, with no related modifications in the other steps.

PaperStrike: I think it quite a common thing to follow the standard step by step when developing a polyfill / plugin that simulates the standard behaviour. And it's rare to see an implementation (browser in our context) uses its own way. So when the standard states it, it should be safe to rely on -)

---

https://github.com/mswjs/interceptors/blob/549d9606db7674468c87ddf4ecabd427eb133a97/src/interceptors/fetch/index.ts#L31

Also, we should use `document.baseURI` in above if insist on parsing it on our own (without `Request`). A advantage I can think of is to satisfy users that expect to see what was directly passed to `fetch` in

https://github.com/mswjs/interceptors/blob/549d9606db7674468c87ddf4ecabd427eb133a97/src/interceptors/fetch/index.ts#L27

I have no preference.

kettanaito:
> I thought you were modifying the sent request via mswjs/msw/src/utils/request/parseIsomorphicRequest.ts#L16 

We are not modifying the _original_ request, only its isomorphic representation. We later rely on that modification when implementing the life-cycle events. For instance, to pair requests with responses:

https://github.com/mswjs/msw/blob/fe6751f76108047fb27b16a875497693da1eb66d/src/node/createSetupServer.ts#L83

The appended `x-msw-request-id` header is _not_ present on the actual request, had such to happen. I believe you may be confusing this with the worker exposing that header in actual requests. The worker has its own request processing pipeline.

kettanaito:
This is a great change, we need a test that reproduces the scenario this change covers: when constructing `fetch` with a `Request` instance. Ideally, a scenario that would fail the previous implementation. 

Feel free to write the test, I can help you in case of any questions. Otherwise, I'll come to this pull request when I have time. Thank you for the effort!

src/interceptors/fetch/index.ts
kettanaito: I'd like to discuss it as a separate issue if you don't mind. It makes such decisions more traceable. Thanks.

alexviar (approved):
This fix is needed. I use RTK query and this library use Request object to invoke fetch, therefore at least two bugs happens:
1. All request are intercepted as GET requests
2. Unhandled requests gets the error "Failed to execute 'fetch' on 'Window': …with a Request object that has already been used."

kettanaito:
I've added integration tests for Node.js and browser to ensure we can intercept requests constructed with `Request` as an input to `fetch`.

kettanaito (approved):
Thank you for making this library better, @PaperStrike! 🎉

Issue #188, opened 2021-11-16
RequestOptions payloads for http/https request generated from url.parse end up having headers removed from the request


dg-harris:
If a RequestOptions payload is created using the node url.parse function, the normalizeClientRequestArgs function ends up stripping the request of additional configuration such as headers.

I've created a test to demonstrate the issue at: https://github.com/dg-harris/interceptors/pull/1/files

Here is an example of a popular library using this pattern: https://github.com/EventSource/eventsource/blob/master/lib/eventsource.js#L84-L139

Here is a repo with a small example of the `msw` library preventing authentication of the `launchdarkly-node-server-sdk` due to this issue:
https://github.com/dg-harris/ld-msw-authentication-issue

Please let me know if there is any additional information I can provide or work I can do to help.

kettanaito:
Hey, @dg-harris. Thanks for reporting this!

We've finished a full rewrite of ClientRequest interception recently (#164), which included fixes for a few issues. I think what you've described has been covered. I recall there was a bug that we disregarded certain request options during `normalizeClientRequestArgs`, which led to those options being lost. 

Could you try cloning the current state of the repo and adding that reproduction test straight to the respective test suite? If it fails, I'd be thankful for a pull request with that failing test.

dg-harris:
Can do! Thanks for the quick response.

gerbyzation:
I ran into this issue as well, this still occurs in the rewritten version (`@mswjs/interceptors@0.13.0`) as well.
Another repro: https://github.com/gerbyzation/msw-launchdarkly-401-repr0

The value of `legacyUrl`:
<img width="2168" alt="Screenshot 2021-12-09 at 12 18 19" src="https://user-images.githubusercontent.com/2500973/145546504-dd8a46d7-ca21-4518-9e6e-a9268f5fa440.png">

Is this something MSW could gracefully handle despite the fact that strictly speaking they are passing down values that do not belong in request options?

I also believe launchdarkly uses their own evensource library instead: https://github.com/launchdarkly/js-eventsource

eli-darkly:
Hi - I'm one of the maintainers of the LaunchDarkly fork of that library, and this was just brought to my attention. I'm a bit confused by this statement from @gerbyzation:

> strictly speaking they are passing down values that do not belong in request options

[...]

PR #189, opened 2021-11-16
Tests for handling request options from parsed url #188

dg-harris:
# Related Issue
#188 
# Motivation
Some node libraries (notably the eventsource library) use the `url.parse` method to generate a RequestOption payload to pass to `https.request`. When this happens the interceptor strips any provided headers (and possibly other configuration) from the request.
Here is an example repo of this behavior causing the launch darkly client to fail when mock service worker is running:
https://github.com/dg-harris/ld-msw-authentication-issue


kettanaito:
Hey, @dg-harris. Thanks for writing the tests.

I'm proposing to close this because the way `eventsource` composes the request options is not according to the Node.js specification (see [my reasoning](https://github.com/mswjs/interceptors/issues/188#issuecomment-1065644045)).

PR #193, opened 2021-12-16
fix: change upload property to object

johnsonlin:
change `upload` to object to avoid `upload.onprogress` runtime error.
some npm file upload deps use `xhr.upload.onprogress = () => {}` for pogress reporting. 
Changing `upload` to an object will at least **_temporarily_** avoid this runtime error and unblock teams on writing tests.

Reference:
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/upload


kettanaito (approved):
Thank you for handling this, @johnsonlin! Welcome to contributors 🎉 

There's a proposal to implement `upload` on XHR: #121. If you feel like this is an interesting area, I could use your expertise to make this improvement happen. Let me know.

PR #203, opened 2022-01-25
test: use name import of "request" in the https tests

kettanaito:
Checks if requests issued by named imports of `https` can be intercepted.


kettanaito:
This was an illustratory pull request. There are no issues when importing `request`/`get` by name from `http`/`https`.

PR #210, opened 2022-02-13
fix(deps): use "headers-polyfill"

kettanaito:
Just a dependency name change.


kettanaito:
Need to update `page-with` too to get rid of `headers-utils` in the lockfile.

PR #214, opened 2022-02-26
fix(cloneIncomingMessage): inherit prototype properties

### LINKED ISSUE #212: cloneIncomingMessage does not handle `headers` field in Node.js v16
I met the problem described in the title. I've investigated it and found the cause.
The `src/interceptors/ClientRequest/utils/cloneIncomingMessage` is not properly.
But I don't know how to fix it properly. I'm planning to consider it but first I share it via this issue.

--- Edit

I removed the meaningless tests.

hrsh7th:
Fix #212 

The current implementation will clone the properties defined in `instance`.
But the `headers` property defined in `prototype` in Node.js v16 or higher.

So this PR aims to support prototype inheriting in `cloneIncomingMessage`.

> The final pitfalls is this is not support `instanceof IncomingMessage` and `instanceof PassThrough`. If the msw uses those operators,  I think this must not be merged.

Fixed.

### DIFF
--- a/src/interceptors/ClientRequest/utils/cloneIncomingMessage.test.ts
+++ b/src/interceptors/ClientRequest/utils/cloneIncomingMessage.test.ts
@@ -12,6 +12,7 @@ test('clones a given IncomingMessage', () => {
   expect(clone.statusCode).toEqual(200)
   expect(clone.statusMessage).toEqual('OK')
   expect(clone.headers).toHaveProperty('x-powered-by', 'msw')
+  expect(clone).toBeInstanceOf(IncomingMessage);
 
   // Cloned IncomingMessage must be marked respectively.
   expect(clone[IS_CLONE]).toEqual(true)
--- a/src/interceptors/ClientRequest/utils/cloneIncomingMessage.test.ts
+++ b/src/interceptors/ClientRequest/utils/cloneIncomingMessage.test.ts
@@ -1,21 +1,25 @@
 import { Socket } from 'net'
 import { IncomingMessage } from 'http'
+import { Stream, Readable, EventEmitter } from 'stream'
 import { cloneIncomingMessage, IS_CLONE } from './cloneIncomingMessage'
 
 test('clones a given IncomingMessage', () => {
-  const source = new IncomingMessage(new Socket())
-  source.statusCode = 200
-  source.statusMessage = 'OK'
-  source.headers = { 'x-powered-by': 'msw' }
-  const clone = cloneIncomingMessage(source)
+  const message = new IncomingMessage(new Socket())
+  message.statusCode = 200
+  message.statusMessage = 'OK'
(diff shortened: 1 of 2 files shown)


kettanaito:
Hey, @hrsh7th! Thank you for submitting this fix. 

> The final pitfalls is this is not support instanceof IncomingMessage and instanceof PassThrough

This may be an issue, as we're exposing the cloned `IncomingMessage` publicly:

https://github.com/mswjs/interceptors/blob/4563ce7493b696830702877d4a61d0fd44d5feda/src/interceptors/ClientRequest/NodeClientRequest.ts#L238

When the consumer adds a `response` listener on `ClientRequest`, they will get the cloned message when using this library. They may have `instanceof` and other logic around that, which will break with this change if I understand correctly. 

Maybe we can think of reusing parts of your solution to preserve the prototype? Would that be possible?

trevoro:
I've tested this locally on node16 re-using the same workflow that the github actions has and it works great!

Regarding the types, this function as it existed before and after this PR forces the type by casting to unknown and then to ClonedIncomingMessage. The stream itself has the same overall shape but it now has the prototype methods added in. If you remove the 'to unknown' part tsc has a way better idea about the type of the stream and it seems like it's been the wrong one for quite some time. Like if you look at the shape of an IncomingMessage vs what's there in `main` it's not the same at all. I think that's out of scope of this issue for now though.

hrsh7th:
I'd changed the implementation to support `stream instanceof IncomingMessage`.

This approach can be broken if both `IncomingMessage` and` PassThrough` have the same properties in different implementations. (This fact is the same in the existing implementation. So I think we don't care about it in this PR for now)

hrsh7th:
I think this PR is ready to review and merge.  👍🏼

src/interceptors/ClientRequest/utils/cloneIncomingMessage.test.ts
```diff
@@ -1,20 +1,25 @@
...
+  message.statusCode = 200
+  message.statusMessage = 'OK'
+  message.headers = { 'x-powered-by': 'msw' }
+  const clone = cloneIncomingMessage(message)
+
+  // Prototypes must be preserved.
+  expect(clone).toBeInstanceOf(IncomingMessage)
+  expect(clone).toBeInstanceOf(EventEmitter)
+  expect(clone).toBeInstanceOf(Stream)
+  expect(clone).toBeInstanceOf(Readable)
```
kettanaito, line 17: I'm adding assertions to check that the cloned message indeed extends all the prototypes of the original message.

kettanaito (approved):
Absolutely fantastic job on this one, @hrsh7th! 🎉 
Thank you for making this library better. Welcome to contributors!

hrsh7th:
Thank you for review, fix and merge!  🎉

PR #215, opened 2022-02-26
fix: use 127.0.0.1 in test to explicitly listen on IPv4 interface

lionralfs:
Hey,

tests in node >= 17.0.0 are not passing due to the way in which `localhost` gets resolved. From (https://github.com/nodejs/node/issues/40537#issuecomment-948047193):

> Node.js now returns IP addresses in the order they are returned from the name resolver/DNS. There's a --dns-result-order command line option you can set to control the behaviour -- the difference in Node.js 17 is the default changed.

This change was introduced in https://github.com/nodejs/node/pull/39987 (node v17.0.0), causing this test to fail:

https://github.com/mswjs/interceptors/blob/4563ce7493b696830702877d4a61d0fd44d5feda/src/interceptors/ClientRequest/NodeClientRequest.test.ts#L151-L170

`localhost` resolves to the IPv6 address `::1` which means that the line `expect(error.address).toEqual('127.0.0.1')` fails.

This PR changes the test to explicitly listen on the IPv4 equivalent.


kettanaito:
Hey, @lionralfs. Thanks for submitting this pull request!

So, does this mean you can no longer construct a `ClientRequest` with a `localhost` hostname?

lionralfs:
This is just a change for the test case, so that tests doesn't break in newer node versions.

Also, you should still be able to use localhost in the `NodeClientRequest` but you might run into issues where if you pass `localhost`, it resolves to `::1` (IPv6) instead of `127.0.0.1` (IPv4). That is out of scope for this library though.

More info (from https://github.com/nodejs/node/pull/39987):

> Switch the default from false (reorder the result so that IPv4 addresses
come before IPv6 addresses) to true (return them exactly as the resolver
sent them to us.)

kettanaito:
Hey, @lionralfs. Thanks for opening this!

We don't currently have official support for Node 17. While the library should work there, our test suite is meant to be run on Node 14/16. I'm putting this change on hold until we adopt Node 17 as the official development version.

kettanaito:
## Released: v0.21.1 🎉

This has been released in v0.21.1!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.21.1)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.21.1)

---

PR #219, opened 2022-03-01
feat: migrate to node 16.14.0 (lts)

### LINKED ISSUE #217: What versions of node do we want to support?
Hey,

Lots of the issues in this package seem to be related to the version of node that people are running. The test suite uses node v12 which lots of people have moved past at this point. If we can have a list of LTS releases that we want to support, then we can write the test suite in a way that supports all of them, then we can sort the issues out and bucket them. If anyone runs into a problem, we can just tag it with the version of node they're using and add the test, and that way we can have the library up to date for the known versions.

Here's an example of a test suite we could adopt for the versions that we want; essentially just loop through the different node versions. 

WDYT?

kettanaito:
- Closes #217 
- Updates to Node.js 16.14.0 as the officially supported version.
- Updates any related dependencies.


src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -116,101 +116,106 @@ export class NodeClientRequest extends ClientRequest {
...
-        '(MOCKED)'
-      )
 
-      callback?.()
-
-      this.log('emitting the custom "response" event...')
-      this.observer.emit('response', isomorphicRequest, isomorphicResponse)
+    // Execute the resolver Promise like a side-effect.
+    // Node.js 16 forces "ClientRequest.end" to be synchronous and return "this".
+    until(async () => this.resolver(isomorphicRequest, this.response)).then(
```
kettanaito, line 122: Resolver execution is a side-effect right now. This needs to be tested.

kettanaito:
The `headers` issue has been resolved (#214).

src/interceptors/ClientRequest/NodeClientRequest.ts
kettanaito: Automated tests report no issues. Response handling logic is asynchronous by design. I think when Node 16 forced `end` to return `this` they accounted for it. As in: return the next (potentially unresolved) state of `ClientRequest`.

kettanaito:
Need to publish 7d8f6bdc3d5e1e664f0a9331911acc5b6802decf first.

PR #250, opened 2022-05-28
fix: remove "RemoteHttpInterceptor" export from the main

kettanaito:
Cannot export `RemoteHttpInterceptor` because it hard relies on `ClientRequestInterceptor`, which makes it usable only in a Node.js environment. If it's explored from `src/index.ts`, then any imports to `@mswjs/interceptors` assume a Node.js environment, which is not the case (fetch interceptor can run in a browser).


kettanaito:
## Released: v0.16.2 🎉

This has been released in v0.16.2!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.16.2)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.16.2)

---

PR #262, opened 2022-06-15
fix(BatchInterceptor): remove listeners between fast refresh

kettanaito:
- Fixes https://github.com/mswjs/msw/issues/1271

## Changes

- Batch interceptor now prunes any existing listeners from the child interceptors. Such listeners may persist between fast refresh, causing the same request to be handled multiple times.
- This adds support to run interceptors in modern apps that use HMR (Fast Refresh). 

## Roadmap

- [ ] Ensure this also works for isolated interceptor usage (not batched). 
- [ ] Mention that you cannot attach listeners to interceptors if you then pass those interceptors to a `BatchInterceptor`. Always attach listeners to the batch interceptor. Maybe because of this accept class references and not instances?

### DIFF
--- a/src/BatchInterceptor.ts
+++ b/src/BatchInterceptor.ts
@@ -31,12 +31,21 @@ export class BatchInterceptor<
     BatchInterceptor.symbol = Symbol(options.name)
     super(BatchInterceptor.symbol)
     this.interceptors = options.interceptors
+
+    /**
+     * Ensure that child interceptors have no event listeners
+     * when the batch interceptor is constructed. The listeners may
+     * persists between Fast Refresh in modern apps, causing
+     * the same intercepted request to be handled multiple times.
+     * @see https://github.com/mswjs/msw/issues/1271
+     */
+    for (const interceptor of this.interceptors) {
+      interceptor['emitter'].pruneListeners()
+    }
   }
 
   protected setup() {
     const log = this.log.extend('setup')
-
-    log('applying all %d interceptors...', this.interceptors.length)
 
     for (const interceptor of this.interceptors) {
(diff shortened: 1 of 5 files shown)


src/BatchInterceptor.ts
```diff
@@ -31,13 +31,22 @@ export class BatchInterceptor<
...
+
+    /**
+     * Ensure that child interceptors have no event listeners
+     * when the batch interceptor is constructed. The listeners may
+     * persists between Fast Refresh in modern apps, causing
+     * the same intercepted request to be handled multiple times.
+     * @see https://github.com/mswjs/msw/issues/1271
+     */
+    for (const interceptor of this.interceptors) {
+      interceptor['emitter'].pruneListeners()
```
kettanaito, line 43: Coupling internal `_events_` pruning and `.removeAllListeners()` causes issues. Looks like we cannot remove the internal record all the time when the consumer removes listeners.

I'm adding a designated method that'd do both only in this particular scenario.

src/utils/AsyncEventEmitter.ts
```diff
@@ -180,4 +207,15 @@ export class AsyncEventEmitter<
     this.readyState = AsyncEventEmitterReadyState.DEACTIVATED
     log('set state to:', this.readyState)
   }
+
+  private removeInternalListeners<Event extends keyof EventMap>(
+    event?: Event
+  ): void {
+    if (event) {
+      this._events[event] = []
```
kettanaito, line 215: Missing `return` statement: passing an explicit event name will still clear _all the events because this condition doesn't short-circuit.

PR #265, opened 2022-06-25
feat(ClientRequest): use proxies for interception

kettanaito:
This is a third generation of the request interception algorithm. To improve on the previous class augmentation principle, we are introducing proxies. Proxies allow us to keep the original request modules implementations intact while spying only on specific scenarios of properties/methods of those classes. 

## Motivation

This should increase the integrity of the tested code even further since there are even fewer things that are mocked right now. We reduce the surface of mocks/spies to the minimum. Less mock code also means better performance and resilience of the entire approach, which benefits our users in the form of faster tests and production-grade requests behavior. 

Because this change moves the response resolution to a mock `Agent` instance, we can control the underlying socket connection and ignore it entirely if there's mock. Previously, we recorded any socket/request errors during this connection and replayed them in the case where there weren't any mocks. 

## Roadmap

- [ ] Finish logging for the new interception.
- [ ] Remove `requestUrl` from the mock agents. They cannot know about individual requests, they receive request info in `addRequest`. Consider parsing the request there instead? Doesn't look like we care about `options` and `callback` now with proxies.
- [ ] Replace `uuid4` with the alternative (see MSW).


src/interceptors/ClientRequest/MockAgent/DuplexPair.ts
```diff
@@ -0,0 +1,53 @@
+import { Duplex } from 'stream'
+import { invariant } from 'outvariant'
+
+export const kOtherSide = Symbol('kOtherSide')
+const kCallback = Symbol('kCallback')
+
+export class DuplexPair {
```
kettanaito, line 7: Boy have I gone into depth here. Brgh.

kettanaito:
Good for reference, not so much for production. Superseded by #515.

PR #267, opened 2022-07-03
Expose new Request API which allow access to raw buffer

95th:
Expose new Request API which allow access to raw buffer

### DIFF
--- a/src/BufferedRequest.test.ts
+++ b/src/BufferedRequest.test.ts
@@ -1,0 +1,52 @@
+import { Headers } from 'headers-polyfill/lib'
+import { BufferedRequest } from './BufferedRequest'
+import { encodeBuf } from './utils/bufferCodec'
+
+describe('Request', () => {
+  const url = new URL('http://dummy')
+  const buf = encodeBuf(JSON.stringify({ hello: 'world' }))
+
+  it('returns JSON, text and array buffer', async () => {
+    const request = new BufferedRequest(url, buf, {})
+    const json = await request.json()
+    const text = await request.text()
+    const arrayBuffer = await request.arrayBuffer()
+    expect(json).toEqual({ hello: 'world' })
+    expect(text).toEqual(JSON.stringify({ hello: 'world' }))
+    expect(arrayBuffer).toEqual(encodeBuf(`{"hello":"world"}`))
+  })
+
+  it('return default method', async () => {
+    const request = new BufferedRequest(url, buf, {})
+    expect(request.method).toEqual('GET')
+  })
+
+  it('return given method', async () => {
+    const request = new BufferedRequest(url, buf, { method: 'POST' })
+    expect(request.method).toEqual('POST')
+  })
+
+  it('return given credentials', async () => {
+    const request = new BufferedRequest(url, buf, { credentials: 'include' })
+    expect(request.credentials).toEqual('include')
+  })
+
+  it('return default credentials', async () => {
+    const request = new BufferedRequest(url, buf, {})
+    expect(request.credentials).toEqual('same-origin')
+  })
+
+  it('return empty headers if not provided', async () => {
+    const request = new BufferedRequest(url, buf, {})
+    expect(request.headers).toEqual(new Headers())
+  })
+
+  it('return empty headers if not provided', async () => {
+    const request = new BufferedRequest(url, buf, {
+      headers: { 'Content-Type': 'application/json' },
+    })
+    expect(request.headers).toEqual(
+      new Headers({ 'Content-Type': 'application/json' })
+    )
+  })
+})
--- a/src/BufferedRequest.ts
+++ b/src/BufferedRequest.ts
@@ -1,0 +1,31 @@
+import { Headers } from 'headers-polyfill/lib'
+import { decodeBuf } from './utils/bufferCodec'
+
+export class BufferedRequest {
+  public readonly method: string
+  public readonly credentials: RequestCredentials
+  public readonly headers: Headers
+
+  constructor(
+    public readonly url: URL,
+    private readonly body: ArrayBuffer,
+    init: RequestInit
+  ) {
+    this.method = init.method || 'GET'
+    this.credentials = init.credentials || 'same-origin'
+    this.headers = new Headers(init.headers)
+  }
+
+  public async text(): Promise<string> {
+    return decodeBuf(this.body)
+  }
+
+  public async json<T = any>(): Promise<T> {
+    const text = await this.text()
+    return JSON.parse(text)
+  }
+
+  public async arrayBuffer(): Promise<ArrayBuffer> {
+    return this.body
+  }
+}
--- a/src/interceptors/ClientRequest/NodeClientRequest.test.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.test.ts
@@ -138,7 +138,7 @@ test('performs the request as-is given resolver returned no mocked response', (d
 test('emits the ENOTFOUND error connecting to a non-existing hostname given no mocked response', (done) => {
   const emitter = new AsyncEventEmitter<HttpRequestEventMap>()
   const request = new NodeClientRequest(
-    normalizeClientRequestArgs('http:', 'http://non-existing-url.com'),
+    normalizeClientRequestArgs('http:', 'http://dummy'),
     {
       emitter,
       log,
@@ -180,7 +180,7 @@ test('does not emit ENOTFOUND error connecting to an inactive server given mocke
   const emitter = new AsyncEventEmitter<HttpRequestEventMap>()
   const handleError = jest.fn()
   const request = new NodeClientRequest(
-    normalizeClientRequestArgs('http:', 'http://non-existing-url.com'),
+    normalizeClientRequestArgs('http:', 'http://dummy'),
     {
       emitter,
       log,
--- a/src/utils/bufferCodec.ts
+++ b/src/utils/bufferCodec.ts
@@ -1,0 +1,13 @@
+export function encodeBuf(input: string): ArrayBuffer {
+  const encoder = new TextEncoder()
+  const encoded = encoder.encode(input)
+  return encoded.buffer.slice(
+    encoded.byteOffset,
+    encoded.byteOffset + encoded.byteLength
+  )
+}
+
+export function decodeBuf(input: ArrayBuffer): string {
+  const decoder = new TextDecoder('utf-8')
+  return decoder.decode(input)
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,7 @@
 export * from './glossary'
 export * from './Interceptor'
 export * from './BatchInterceptor'
+export * from './BufferedRequest'
 
 /* Utils */
 export { getCleanUrl } from './utils/getCleanUrl'
--- a/test/modules/XMLHttpRequest/response/xhr.browser.test.ts
+++ b/test/modules/XMLHttpRequest/response/xhr.browser.test.ts
@@ -92,7 +92,7 @@ test('bypasses a request not handled in the resolver', async () => {
 
 test('bypasses any request when the interceptor is restored', async () => {
   const scenario = await prepareRuntime()
-  // Uusing the "createRawBrowserXMLHttpRequest" because when the interceptor
+  // Using the "createRawBrowserXMLHttpRequest" because when the interceptor
   // is restored, it won't dispatch the "resolver" event.
   const callXMLHttpRequest = createRawBrowserXMLHttpRequest(scenario)
 


src/interceptors/ClientRequest/NodeClientRequest.test.ts
```diff
@@ -180,7 +180,7 @@ test('does not emit ENOTFOUND error connecting to an inactive server given mocke
   const emitter = new AsyncEventEmitter<HttpRequestEventMap>()
   const handleError = jest.fn()
   const request = new NodeClientRequest(
-    normalizeClientRequestArgs('http:', 'http://non-existing-url.com'),
```
95th, line 183: seems like `non-existing-url.com` exists.

95th: Also, CI seems to fail to lookup `dummy`

src/BufferedRequest.ts
```diff
@@ -0,0 +1,38 @@
...
+  public arrayBuffer(): ArrayBuffer {
+    return this.body
+  }
+
+  public get credentials(): RequestCredentials {
+    return this.init.credentials || 'same-origin'
+  }
+
+  public get headers(): Headers {
+    return new Headers(this.init.headers)
```
kettanaito, line 36: Constructing headers on each getter is rather costly. Let's move this to the constructor itself, so we create a `Headers` instance only once, when the Request itself is created.

```ts
constructor(...args) {
  this.headers = new Headers(init.headers)
}
```

src/BufferedRequest.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  public get method(): string {
+    return this.init.method || 'GET'
+  }
+
+  public text(): string {
+    return decodeBuf(this.body)
+  }
+
+  public json<T = any>(): T {
```
kettanaito, line 25: To keep compatibility with Fetch API Request, let's make `text`, `json`, and `arrayBuffer` return a Promise that resolves the values you have now.

src/interceptors/ClientRequest/NodeClientRequest.test.ts
```diff
@@ -138,7 +138,7 @@ test('performs the request as-is given resolver returned no mocked response', (d
 test('emits the ENOTFOUND error connecting to a non-existing hostname given no mocked response', (done) => {
   const emitter = new AsyncEventEmitter<HttpRequestEventMap>()
   const request = new NodeClientRequest(
-    normalizeClientRequestArgs('http:', 'http://non-existing-url.com'),
+    normalizeClientRequestArgs('http:', 'http://dummy'),
```
kettanaito, line 141: The actual issue here is that we are making a side-effect that is not managed by this test. I've tried migrating most of such things to use a locally spawned server but, apparently, I've missed a few tests. 

Let's spawn a server and route the requests there:

https://github.com/mswjs/interceptors/blob/98e64aa75c877d57da28a79d5939d28b5fbc0a66/test/modules/http/intercept/http.get.test.ts#L11-L15

For the test cases when we don't want the test to hit the endpoint, I'm usually returning `res.status(500)` from the actual server route so that even if the intention breaks, the response is not what the test expects.

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -84,7 +86,7 @@ export const createXMLHttpRequestOverride = (
     public statusText: string
     public user?: string
     public password?: string
-    public data: string
+    public data: ArrayBuffer
```
kettanaito, line 89: Can we make this assumption that XHR `data` will always be buffer?

Note that this `data` represents whichever payload was fed to XHR upon construction. Could you please share some more context as to why you're making this change?

src/utils/bufferCodec.ts
```diff
@@ -0,0 +1,15 @@
...
+  const encoder = new TextEncoder()
+  const encoded = encoder.encode(input)
+  return encoded.buffer.slice(
+    encoded.byteOffset,
+    encoded.byteOffset + encoded.byteLength
+  )
+}
+
+export function decodeBuf(input: ArrayBuffer): string {
+  const decoder = new TextDecoder('utf-8')
```
kettanaito, line 13: I think we should accept the buffer encoding as an input to this function. We should respect the original request's encoding. What do you think about this?

src/BufferedRequest.test.ts
```diff
@@ -0,0 +1,52 @@
+import { Headers } from 'headers-polyfill/lib'
+import { BufferedRequest } from './BufferedRequest'
+import { encodeBuf } from './utils/bufferCodec'
+
+describe('Request', () => {
+  const url = new URL('http://dummy')
+  const buf = encodeBuf(JSON.stringify({ hello: 'world' }))
+
+  it('returns JSON, text and array buffer', () => {
```
kettanaito, line 9: I would split this into separate tests representing individual request method:

```ts
it('returns arrayBuffer')
it('returns plain text')
it('returns json')
```

src/BufferedRequest.ts
```diff
@@ -0,0 +1,38 @@
+import { Headers } from 'headers-polyfill/lib'
+import { decodeBuf } from './utils/bufferCodec'
+import { uuidv4 } from './utils/uuid'
+
+export class BufferedRequest {
```
kettanaito, line 5: I'd call this `IsomorphicRequest` or just `Request`. We are really aiming at Fetch API's `Request` compatibility so whether we handle request body as a buffer is an implementation detail of this class. The consumer should not know about that.

src/BufferedRequest.ts
```diff
@@ -0,0 +1,38 @@
...
+  public id = uuidv4()
+
+  constructor(
+    public readonly url: URL,
+    private readonly body: ArrayBuffer,
+    private readonly init: RequestInit
```
kettanaito, line 12: Unfortunately, `RequestInit` is only defined in `lib/dom` in TypeScript. It will not exist in Node.js.

We need to re-defined the request init manually:

```ts
export interface RequestInit {
  method: string
  headers: Record<string, string | string[]> | Headers
  credentials: RequestCredentials // our internal type, can reuse
  body: ArrayBuffer
}
```

Let's also change the constructor signature to match that of the Fetch API Request:

```ts
new Request(url)
new Request(url, { headers: {}, body: new ArrayBuffer() })
```

src/glossary.ts
```diff
@@ -19,8 +20,14 @@ export interface IsomorphicRequest {
   body?: string
 }
 
-export interface InteractiveIsomorphicRequest extends IsomorphicRequest {
-  respondWith: LazyCallback<(mockedResponse: MockedResponse) => void>
+export class InteractiveIsomorphicRequest extends BufferedRequest {
+  constructor(
+    buffered: BufferedRequest,
+    readonly respondWith: LazyCallback<(mockedResponse: MockedResponse) => void>
```
kettanaito, line 26: I'd define the `respondWith()` method on the `InteractiveIsomorphicRequest` class straight away. You cannot have an interactive IR without the `respondWith` method. 

The method itself is a deferred method, so it doesn't need to accept anything from the consumer to be defined. The consumer then can await this method being called, and get its return value, by doing `await interactiveRequest.respondWith.invoked()`.

kettanaito (changes requested):
What a fantastic kick off of this idea! Thank you, @95th! 

I've left a few comments and thoughts on the changes. I would like to know what you think about those. Let's discuss it, make whichever adjustments are necessary, and have a unified request handling!

src/BufferedRequest.ts
95th: Fetch api has those for response. also, currently most of my mocks are synchronous. Making these return promise will force the user to use async await.

src/BufferedRequest.ts
95th: Types are stripped at compile time. So, this is not required at runtime by nodejs. 

I'll change the constructor signature to be more fetch compatible.

src/glossary.ts
95th: also probably should move the `invoked` on IIR as well?

src/BufferedRequest.ts
95th: Calling it request makes it weird at place where we are using fetch api as well. I'll call it Isomorphic request.. 

Or may be `InterceptedRequest` ? wdyt?

src/interceptors/ClientRequest/NodeClientRequest.test.ts
95th: We want `ENOTFOUND` error in that test case. Will calling `localhost:xxx` give that error?

src/utils/bufferCodec.ts
95th: We can provide that and also accept it in `text` function

kettanaito: The `text` should return the text representation of the given buffer with the encoding that was initially given to that request instance (we can fallback to utf8 if not set). If you wish to create a text representation with a different encoding you should do `req.arrayBuffer` and then something like `.toString('latin1')` based on the ArrayBuffer API (you get the idea). 

I'm trying to say that accepting a custom encoding in the `text` function is not idiomatic.

src/BufferedRequest.ts
kettanaito: I vote for `IsomorphicRequest`. It describes what it stands for the best—an origin-agnostic representation of a request. It's a middle ground to describe any intercepted request regardless of its origin: `http.ClientRequest`, `XMLHttpRequest`, `fetch.Request`, and, in the future, things like Undici's request. Maybe there's a better engineering term for this kind of abstraction but I grabbed the term isomorphic as in "having the same shape (morph)".

src/BufferedRequest.ts
kettanaito: Types are stripped from the code but not from the type definitions. `BufferedRequest.d.ts` will reference global scope `Request` and will cause a type violation when people will run it Node in TypeScript projects.

src/glossary.ts
kettanaito: No, `invoked()` is a deferred promise from the `respondWith()` method. It doesn't belong to any particular class, only to that lazy function created with `createLazyCallback`. I wouldn't touch this behavior, it works so let's not concern ourselves with it at this moment.

src/BufferedRequest.ts
kettanaito: It looks like it's the same for Request:

- [`Request.arrayBuffer()`](https://developer.mozilla.org/en-US/docs/Web/API/Request/arrayBuffer)
- [`Request.json()`](https://developer.mozilla.org/en-US/docs/Web/API/Request/json)
- [`Request.text()`](https://developer.mozilla.org/en-US/docs/Web/API/Request/text)

Those methods return promises. Let's keep our abstraction compatible. 

I'm advocating for fetch API compatibility because if we achieve that, we will be able to use an intercepted request instance as an input to regular `window.fetch()`, which is a great way to perform [response patching](https://mswjs.io/docs/recipes/response-patching).

src/interceptors/ClientRequest/NodeClientRequest.test.ts
kettanaito: Did you discover that `https://non-existing-url.com/` exist? I don't think it does, at least when I try to access it. Could it been some hiccup during local testing? I'd like to learn more, maybe we don't need this change.

kettanaito: We are getting `EAI_AGAIN` on CI right now because `http://dummy` is not a valid DNS record. I hope if we revert it back to the previous URL, it will produce connection error, and the test will pass. Maybe something else interfered while you were testing.

src/BufferedRequest.ts
kettanaito: Also, an interesting detail from the docs:

> The response is always decoded using UTF-8.
> — Request.text()

To the point of `text()` accepting a custom encoding. Turns out it can't.

src/interceptors/ClientRequest/NodeClientRequest.test.ts
95th: Well thats flakey

src/glossary.ts
95th: I don't understand how respondWith will be defined. Currently we can call `request.respondWith(whatever)`

kettanaito: We should define it in the `InteractiveIsomorphicRequest` class:

```ts
class InteractiveIsomorphicRequest {
  public respondWith(response: MockedResponse) {
    return createLazyCallback({ maxCalls: 1, maxCallsCallback, })
  }
}
```

I'd expect this to work unless I'm missing something very important.

src/interceptors/ClientRequest/NodeClientRequest.test.ts
kettanaito: How do you think we could improve this? We need to connect to a non-existing address to assert how NodeClientRequest behaves in this scenario.

kettanaito:
The tests should be passing now! I've also added a few minor changes and left a roadmap in the pull request's description of what's remaining. You don't have to tackle everything that remains, just let me know what interests you and I can take care of the rest (like updating the docs).

src/IsomorphicRequest.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export interface RequestInit {
+  method?: string
+  headers?: Record<string, string | string[]> | Headers
+  credentials?: RequestCredentials
+  body?: ArrayBuffer
+}
+
+export class IsomorphicRequest {
+  public id: string
```
kettanaito, line 13: Need to keep `id` as writable because our test helpers re-construct `IsomorphicRequest` instance using an intercepted request object in Playwright.

src/RemoteHttpInterceptor.ts
```diff
@@ -118,17 +116,19 @@ export class RemoteHttpResolver extends Interceptor<HttpRequestEventMap> {
...
-      log('parsed intercepted request', isomorphicRequest)
+      const body = requestJson.body ? bufferFrom(requestJson.body) : undefined
 
-      const interactiveIsomorphicRequest: InteractiveIsomorphicRequest = {
-        ...isomorphicRequest,
-        respondWith: createLazyCallback(),
-      }
+      const isomorphicRequest = new IsomorphicRequest(requestJson.url, {
+        ...requestJson,
+        body: body?.buffer || new ArrayBuffer(0),
```
kettanaito, line 126: `new ArrayBuffer(0)` is the default iso request body value if `init.body` is undefined. Perhaps we could remove this pipe `||`?

src/interceptors/ClientRequest/NodeClientRequest.test.ts
```diff
@@ -151,6 +148,10 @@ test('emits the ENOTFOUND error connecting to a non-existing hostname given no m
     done()
   })
 
+  request.on('response', () => {
```
kettanaito, line 151: Looks like a debugging leftover.

95th: it is executed on my local where this test fails

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -115,21 +109,10 @@ export class NodeClientRequest extends ClientRequest {
...
-          invariant(
-            false,
-            'Failed to respond to "%s %s" request: the "request" event has already been responded to.',
-            isomorphicRequest.method,
-            isomorphicRequest.url.href
-          )
-        },
-      }),
-    }
+    const request = this.toIsomorphicRequest(requestBody)
```
kettanaito, line 112: I would keep `const isomorphicRequest` for clarity. There's always a `request` in the context, which is the intercepted request. Just so it doesn't get us confused.

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -388,16 +371,16 @@ export class NodeClientRequest extends ClientRequest {
...
 
-    const resolvedRequestBody = bodyBufferToString(
-      Buffer.concat(this.requestBody)
-    )
-
+    const resolvedRequestBody = Buffer.concat(this.requestBody)
     this.log('resolved request body:', resolvedRequestBody)
 
-    return resolvedRequestBody
+    return resolvedRequestBody.buffer.slice(
```
kettanaito, line 377: This looks like what `bufferUtils.encodeBuffer()` is doing, is it not? Can we use that utility here as well?

src/interceptors/ClientRequest/NodeClientRequest.test.ts
kettanaito: Could you post the test runner's result when this test fails? I wonder what exactly makes it fail: it actually connects to a non-existing-url.com?

95th: image

kettanaito: Here's what I'm getting:
```sh
% ping non-existing-url.com
ping: cannot resolve non-existing-url.com: Unknown host
```

Perhaps you have some custom host configuration on your machine? 

In any case, try setting a different bust still valid URL and see if it results in the same test failure maybe? It doesn't really matter what URL we try to connect here, only that the connection fails. It can be `<uuid>.com`.

src/interceptors/ClientRequest/NodeClientRequest.test.ts
95th: I think there is no way around it.

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
95th: Removed

src/glossary.ts
```diff
@@ -1,28 +1,11 @@
 import type { HeadersObject, Headers } from 'headers-polyfill'
-import type { LazyCallback } from './utils/createLazyCallback'
+import type { InteractiveIsomorphicRequest } from './InteractiveIsomorphicRequest'
```
kettanaito, line 2: We used to export the `InteractiveIsomorphicRequest` type from the root of the package as well, as it was in the glossary module before. Could you please add the IIR export from its own module in the `index.ts`?

src/interceptors/ClientRequest/NodeClientRequest.test.ts
kettanaito, line 151: Should we remove this if the issue has been resolved?
The response callback shouldn't even be called in this scenario, as the request errors on trying to connect to a non-existing endpoint.

test/helpers.ts
```diff
@@ -216,22 +217,20 @@ export async function extractRequestFromPage(
 
       window.addEventListener(
         'resolver' as any,
-        (event: CustomEvent<string>) => {
+        (event: CustomEvent<StringifiedIsomorphicRequest>) => {
           clearTimeout(timeoutTimer)
-          resolve(JSON.parse(event.detail))
+          resolve(event.detail)
```
kettanaito, line 222: We were previously parsing this, as `event.detail` is a string. Do we not need to parse it into object anymore? 🤔

test/modules/fetch/intercept/fetch.body.runtime.js
```diff
@@ -2,7 +2,7 @@ import { FetchInterceptor } from '@mswjs/interceptors/lib/interceptors/fetch'
 
 const interceptor = new FetchInterceptor()
 interceptor.on('request', (request) => {
-  window.requestBody = request.body
+  window.requestBody = request.text()
```
kettanaito, line 5: We should await this, as `request.text()` returns a Promise.

test/modules/fetch/intercept/fetch.browser.test.ts
```diff
@@ -82,7 +82,7 @@ describe('HTTP', () => {
     expect(request.url.href).toEqual(url)
     expect(request.headers.get('x-custom-header')).toEqual('yes')
     expect(request.credentials).toEqual('same-origin')
-    expect(request.body).toEqual('{"body":"post"}')
+    expect(await request.text()).toEqual('{"body":"post"}')
```
kettanaito, line 85: This is actually a great opportunity to have some transient assertions: let's read request body as JSON here! Would make the "toEqual" part easier to manage than stringified objects.

test/third-party/follow-redirect-http.test.ts
```diff
@@ -71,17 +71,21 @@ test('intercepts a POST request issued by "follow-redirects"', async () => {
   expect(initialRequest.credentials).toBe('same-origin')
   expect(initialRequest.headers.get('Content-Type')).toBe('application/json')
   expect(initialRequest.headers.get('Content-Length')).toBe('23')
-  expect(initialRequest.body).toBe(JSON.stringify({ todo: 'Buy the milk' }))
+  expect(await initialRequest.text()).toBe(
```
kettanaito, line 74: Another opportunity to extract request body as json.

test/third-party/supertest.test.ts
```diff
@@ -44,5 +44,5 @@ test('preserves original POST request JSON body', async () => {
   expect(requests).toHaveLength(1)
   const [request] = requests
   expect(request.method).toBe('POST')
-  expect(request.body).toEqual(JSON.stringify({ query: 'foo' }))
+  expect(await request.text()).toEqual(JSON.stringify({ query: 'foo' }))
```
kettanaito, line 47: And another one!

kettanaito:
I'm also thinking of how the IIR will be propagated to the request listener. It'd be great to just reuse the `IsomorphicRequest` instance provided by this library in MSW, but what this library exposes is an _interactive_ instance, meaning it has the `respondWith() method. At the same time, the `req` in MSW handlers must not have such method, and should be a barebones iso request instance. 

I'm thinking of something like `IsomorphicRequest.clone(): IsomorphicRequest` that'd create a clean instance of request upon use. Since IIR extends IR, it will also have the `.clone()` method but will return a plain request without the `respondWith()` method. 

What do you think about this approach?

test/helpers.ts
95th: It is automatically serialized and deserialized

test/modules/fetch/intercept/fetch.body.runtime.js
95th: We already return promise for it in the test:

```
function getRequestBody(page: Page): Promise<string> {
  return page.evaluate(() => window.requestBody)
}
```

and documentation for `evaluate` says:

> If the function passed to the [page.evaluate(pageFunction[, arg])](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) returns a [Promise], then [page.evaluate(pageFunction[, arg])](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) would wait for the promise to resolve and return its value.

src/glossary.ts
95th: Done

src/interceptors/ClientRequest/NodeClientRequest.test.ts
95th: Removed

95th:
I think we should have a `MockedRequest extends IsomorphicRequest` which include cookies etc and it won't expose the `respondWith` details in the type system.

src/IsomorphicRequest.ts
```diff
@@ -0,0 +1,57 @@
...
+  public readonly url: URL
+  public readonly method: string
+  public readonly headers: Headers
+  public readonly credentials: RequestCredentials
+  public readonly body: ArrayBuffer
```
95th, line 18: Any reason to keep this public?

kettanaito: No. Following our discussion in Discord, let's make `body` private because we currently don't comply with the spec in this regard. We are not returning `body` as a readable stream.

kettanaito (approved):
This is by far the most enjoyable pull request in my open source journey. Thank you for collaborating with me on this!

kettanaito:
## Released: v0.17.0 🎉

This has been released in v0.17.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.17.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.17.0)

---

PR #270, opened 2022-07-04
feat(BatchInterceptor): batch multiple events with the same id

kettanaito:
- Related to https://github.com/mswjs/interceptors/issues/163


kettanaito:
I'm leaving this for posterity. This doesn't fix anything because every interceptor will generate a unique ID for the same intercepted request. There's virtually no knowing if multiple interceptors are handling the same request right now.

PR #274, opened 2022-07-19
fix: support 0 as an explicit status code

tanem:
Bumped into this in MSW, setting `ctx.status(0)` results in a `200` status in the mocked response. Guessing this change will more closely match the intent of the code since `status` is a `number`?

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -304,7 +304,7 @@ export const createXMLHttpRequestOverride = (
           // Trigger a loadstart event to indicate the initialization of the fetch.
           this.trigger('loadstart')
 
-          this.status = mockedResponse.status || 200
+          this.status = mockedResponse.status ?? 200
           this.statusText = mockedResponse.statusText || 'OK'
           this._responseHeaders = mockedResponse.headers
             ? objectToHeaders(mockedResponse.headers)
--- a/src/utils/toIsoResponse.ts
+++ b/src/utils/toIsoResponse.ts
@@ -6,7 +6,7 @@ import { IsomorphicResponse, MockedResponse } from '../glossary'
  */
 export function toIsoResponse(response: MockedResponse): IsomorphicResponse {
   return {
-    status: response.status || 200,
+    status: response.status ?? 200,
     statusText: response.statusText || 'OK',
     headers: objectToHeaders(response.headers || {}),
     body: response.body,


kettanaito:
Hey, @tanem. Thanks for opening this.

Could you tell me more about your use case? To my best knowledge, HTTP status codes must be within 100-5xx range in order to comply with the spec. While `ctx.status(0)` results in status code 200, I'd argue that the mistake here is setting the status code to 0.

tanem:
Hey @kettanaito 👋 

I'm migrating tests for a legacy app. We see this scenario reported in our monitoring tools and want it handled in a specific way. It's for these sorts of scenarios described in [XMLHttpRequest.status](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/status):

> Before the request completes, the value of status is 0. Browsers also report a status of 0 in case of XMLHttpRequest errors.

Also described [here](https://www.ibm.com/support/pages/what-causes-status-code-0-indicating-reqcancelled).

Yep agree with you, it's a bit ham-fisted and I'm guessing there's a better way of testing this... but I noticed I can successfully set `ctx.status(1)` in MSW which is also outside the 100-5xx range, so figured if that range wasn't enforced and the type is `number` then `0` should be allowed too 😅 

Also wasn't sure if something needed to change elsewhere (e.g. should `UNSENT` have status `0` [here](https://github.com/mswjs/interceptors/blob/main/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L132=)?), so thought I'd raise a PR for this lo-fi option to start with.

kettanaito (approved):

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -304,7 +304,7 @@ export const createXMLHttpRequestOverride = (
           // Trigger a loadstart event to indicate the initialization of the fetch.
           this.trigger('loadstart')
 
-          this.status = mockedResponse.status || 200
+          this.status = mockedResponse.status ?? 200
```
kettanaito, line 307: We need to add an integration test for mocking a status code that'd fall under this change. Mocking `0` status code may be a good example.

tanem:
> What do you think about this roadmap? Would you have the time/interest to make it happen together?

Sounds good, would love to help out 👍 Did you want the changes made under this PR, or a new PR/PRs?

kettanaito:
I think your changes are great and we can ship them straight away. I've opened #281 to follow up whenever you're interested. Thanks for the help!

kettanaito:
## Released: v0.17.4 🎉

This has been released in v0.17.4!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.17.4)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.17.4)

---

tanem:
> I think your changes are great and we can ship them straight away. I've opened #281 to follow up whenever you're interested. Thanks for the help!

No worries and thanks for the merge 🙏 Will hit that issue up if I get the chance to make those other changes.

Issue #282, opened 2022-08-26
Where is the source code for `page-with` and `@open-draft/test-server` packages?


milesrichardson:
Hi, thanks for the great libraries!

I'm working on a PR to (hopefully) improve support for Node 18 and native fetch. So far I've found a few related bugs caused by incorrect serialization of IPv6 host in URLs, which is a problem on Node 18 when running in dual stack environments like GitHub Actions because `localhost` can resolve to `::1`, and some code in `page-with` and `test-server` libraries incorrectly serializes any IPv6 host to an invalid URL. The fix is to surround an IPv6 host with square brackets. 

This is an easy bug to fix, but I don't know where to make the PR. I haven't been able to find the source code for [`page-with`](https://www.npmjs.com/package/page-with) (no source listed) or [`@open-draft/test-server`](https://www.npmjs.com/package/@open-draft/test-server) (source link 404). Is the source code for these packages available somewhere? I see that @kettanaito is their maintainer, but I cannot find the source repositories on his profile nor anywhere in the mswjs organization.

In the meantime, I'm able to workaround the bug by fixing it with re-exports and prototype patching, but that's not ideal

kettanaito:
Hey, @milesrichardson. 

Well, that's a mess. I'm glad to have this opportunity to clean things up. 

You can find the source code for `page-with` here: https://github.com/kettanaito/page-with. If you could also add the `repository` url in `package.json` as a part of your changes I'd be grateful! 

As for the `@open-draft/test-server`, it's a private package I've developed, so its source code is private for now. I don't have intention to open source it yet, but I can add the fix you propose manually and publish a new version so we can continue with proper testing of this, and many other dependent libraries.

On the related note, is the issue you're experiencing related to https://github.com/mswjs/interceptors/pull/215 by any chance?

milesrichardson:
Thanks! I pushed my changes with the hacky patch version of these fixes in #283. I made sure the commits with the patches are atomic so that reverting them can be done with `git revert`. 

> You can find the source code for page-with here: https://github.com/kettanaito/page-with. If you could also add the repository url in package.json as a part of your changes I'd be grateful!

Cool, I will take a look at this and hopefully open a PR (probably should spend the next few hours on other priorities -_-)

[...]

PR #283, opened 2022-08-29
feat: support Node.js 18

milesrichardson:
**In this PR**

This PR migrates the development environment to Node v18 (which [replaces v16
as active LTS in October](https://nodejs.org/en/about/releases/)), and adds tests in CI for v14, v16 and v18. It fixes
compatibility issues running existing tests in a Node v18 environment and does
not break tests in earlier environments. It does not include any new user-facing features.

It **does not** add support for "native fetch" (if that's even missing - see #159)
but it creates the baseline for adding it, which is the next step. (I assume
this will be easy, and will hopefully only require adding new integration
tests to use the native `fetch` instead of the one from `cross-fetch`.)

It's smaller than it looks. Most of the file changes are changing import
paths in 8de6a91568dcde044c15a9231285b38952e7ba5d, which patches a bug in upstream `@opendraft/test-server`
by mutating and re-exporting it, hence the import path changes.

_Fixes_:

- #282
  - this contains the patches referenced there
- #215
  - I think this is one of the bugs fixed here. I originally tried that same fix, of explicitly using `127.0.0.1` instead of `localhost`, but there were still some tests failing - I forget the details. The fix in this PR doesn't change the tests and should work correctly with both IPv4 and IPv6 hosts (which will be tested as long as GitHub Actions `ubuntu:latest` continues to resolve `localhost` to an IPv6 address).

_Makes progress on_:

- #159
  - the next step is to add support for global fetch interception, which might not
  need undici, but that could be added as a third-party interceptor

_Related_:

- #180
  - I did [try rebasing this branch](https://github.com/milesforks/msw-interceptors/pull/2), but then I realized the tests were using
  a now-deprecated API. Also, I don't think using undici is necessary, if we can
  use global fetch interceptor, although undici could be a sensible third-party
  interceptor, perhaps using its native mocks.
- #217
  - Assuming this PR is merged, we will now support Node v14, v16 and v18, while using v18 as the default development version (via `.nvmrc`)

_Upstream PR_:

- The upstream fix for the `page-with` bug (which is fixed here via 1c4309e and 4291844) is open at https://github.com/kettanaito/page-with/pull/9

**How to review**

I suggest reviewing this commit-by-commit, because unfortunately one of the
commits (8de6a91568dcde044c15a9231285b38952e7ba5d) required updating an import
path in every test file.

If you want to run this locally, make sure you switch to Node 18. You might
also need to delete `node_modules` to fix the `memfs` resolution issue
(see 1c4309e), but I'm not sure.

Here's how I do that with `nvm`, while retaining my global package installations:

```bash
nvm install --reinstall-packages-from=current $(cat .nvmrc)
nvm use $(cat .nvmrc)

# maybe needed?
rm -rf node_modules

yarn install --frozen-lockfile
```

**Fix IPv6 compatibility issues**

Multiple bugs shared the same root cause, which was incorrect serialization of
IPv6 hosts in URLs. This only surfaced as an issue because Node v18 now
[binds to a hostname by choosing the first IP address returned by the OS](https://github.com/nodejs/node/issues/40537#issuecomment-948047193),
including for the `localhost` domain name. Previously, this would only resolve to `127.0.0.1`,
but now on some hosts (such as GitHub Actions), it resolves to `::1`.

We use this IP address, which is returned by `HttpServer.listen()` when we bind
on `localhost`, to construct a URL in multiple places, mostly in tests but also
user facing (anyone using the library and binding to `localhost` would encounter
this same bug). The bug is caused by code that incorrectly
assumes `${scheme}://${host}:${port}` serializes to a valid URL even for an IPv6
host, which should be surrounded by square brackets, like `${scheme}://[${host}]:${port}`.

This bug was repeated in multiple places:

- In this `interceptors` codebase
  - I fixed these bugs directly. I believe all changes should only affect tests, except for the
  two commits marked `fix` (eb46b837c67ba80ee1f614b19ceda4760eb87661 and 22abd1c7fb2f6190227d3e9dd61ec0db727d1623)
- In the `@opendraft/test-server` codebase (dev dependency, only affects tests)
  - closed source, can be fixed upstream
    - see #282
  - I fixed this with a patch and re-export in 8de6a91568dcde044c15a9231285b38952e7ba5d, which is
  the commit that changes a bunch of files, because it needs to update the imports to use
  the re-exported server.
    - See the diff of `OpenDraftTestServer.ts` in that commit for the patch changes which
    could be replicated to the upstream codebase
  - If you fix the bugs in upstream, you should be able to simply revert this commit
  to remove the patch (I was careful to make it atomic)
- In the `page-with` codebase (dev dependency, only affects tests)
  - [recently open source](https://github.com/kettanaito/page-with), can be fixed upstream
    - see #282
    - upstream PR with "proper" fix: https://github.com/kettanaito/page-with/pull/9 (can be merged independently of this current PR)
  - bug 1: the `memfs` dependency needed to be upgraded. I fixed this by setting
  a compatible version in the `resolutions` field of our `package.json` so that
  the transitive dependency is forced to resolve to the latest `memfs` (1c4309e190078ec8ef0a686b7b72ef7c804ff407)
  - bug 2: The IPv6 serialization is buggy in `PreviewServer`. I fixed this by
  patching the singleton `server` object that is created at browser setup time (4291844df928b0f78323574df83d0e9d2e0b7cb6).
  When the upstream bug is fixed, this commit can be reverted.


milesrichardson:
Note: the change includes an edit to `ci.yml` to add tests for Node v18, but it looks like those new checks are not running for this PR, probably for security reasons. FYI you can see the Node v18 tests passing successfully in the run against my fork: https://github.com/milesforks/msw-interceptors/pull/4

src/interceptors/fetch/index.ts
```diff
@@ -40,7 +40,13 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
...
       const request = new Request(input, init)
 
-      const url = typeof input === 'string' ? input : input.url
+      const url =
+        typeof input === 'string'
+          ? input
+          : input instanceof URL
+          ? input.href
+          : input.url
+
```
milesrichardson, line 49: I believe this is the one user-facing change, to select `.href` in the case that `input` is an instance of `URL`.

I don't fully grok the root cause of this - update to JSDom maybe? But I note that it fixed a type error, which was reported in my VSCode with latest TypeScript but not by any build script. It also fixed some tests, but I can't remember exactly which, since this was one of multiple bugs fixed simultaneously.

It looks like a safe change to me, though, since the new branch should only evaluate when `input instanceof URL`, which previously would only result in an error since [there is no valid `url` property on an instance of `URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL). And it still works on Node v14/v16, although that's not unexpected if the root cause was an update to `jest-environment-jsdom`.

test/patched/OpenDraftTestServer.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  var origGetServerAddress = HttpServer.getServerAddress
+  /**
+   * Patch `static HttpServer.getServer` to fix URI serialization of IPv6 hosts,
+   * by calling the original method and then modifying the return value.
+   *
+   * @see HttpServer.getServerAddress
+   */
+  HttpServer.getServerAddress = function () {
```
milesrichardson, line 56: This is the static method you would need to fix in the upstream `@opendraft/test-server`. It's the `static getServerAddress()` method on `HttpServer`.

Note this patch also includes a change to `buildHttpServer`, but that's only because it calls the static method and we need to make sure it calls the patched version. I believe the only upstream change required is in the `static getServerAddress()` method, but definitely double check that.

test/patched/PageWithPreviewServer.ts
```diff
@@ -0,0 +1,36 @@
...
+ * the connection info of the preview server which was created for the page.
+ *
+ * This function needs to be called from a scope where that mutable variable
+ * is available. That's why we call it in `jest.browser.setup.ts`.
+ *
+ * When the upstream bug is fixed, remove this file and remove the code that
+ * calls it in `jest.browser.setup.ts`
+ *
+ */
+export const patchServerConnectionInfo = (server: PreviewServer) => {
```
milesrichardson, line 13: This isn't really a patch - we just mutate the invalid object (`connectionInfo` with invalid `url`) that is created by `PreviewServer.listen()` when it returns `this.connectionInfo`. So the upstream fix is in the `PreviewServer.listen()` method.

We call this function in `jest.browser.setup.ts`, in order to mutate the `server` object exported from `page-with` and updated via the `createBrowser` promise (subtle, but it works).

test/jest.browser.setup.ts
```diff
@@ -11,7 +12,20 @@ beforeAll(async () => {
     serverOptions: {
       webpackConfig,
     },
-  })
+  }).then(
+    // NOTE: when upstream patch is fixed, remove this last promise handler
+    (browser) => {
+      patchServerConnectionInfo(
```
milesrichardson, line 18: Here is where we mutate `server` (see previous comment on `patched/PageWithPreviewServer.ts`) to fix `server.connectionInfo.url`.

test/jest.browser.setup.ts
```diff
@@ -11,7 +12,20 @@ beforeAll(async () => {
...
+        // note: we know server connection is up because createBrowser waited `until`
+        // it was up, meaning PreviewServer.listen() set `this.connectionInfo`
+        server as NonNullable<typeof server> & {
+          connectionInfo: NonNullable<typeof server['connectionInfo']>
+        }
+      )
+
+      return browser
+    }
+  )
```
milesrichardson, line 28: Here is where we mutate `server` (see previous comment on `patched/PageWithPreviewServer.ts`) to fix `server.connectionInfo.url`.

test/patched/OpenDraftTestServer.ts
```diff
@@ -0,0 +1,115 @@
...
+              ? `[${this.host}]`
+              : this.host
+          }:${this.port}`
+        ).href
+      },
+      enumerable: true,
+    })
+
+    return address
+  } as PatchedGetServerAddress
```
milesrichardson, line 80: This is the static method you would need to fix in the upstream `@opendraft/test-server`. It's the `static getServerAddress()` method on `HttpServer`.

Note this patch also includes a change to `buildHttpServer`, but that's only because it calls the static method and we need to make sure it calls the patched version. I believe the only upstream change required is in the `static getServerAddress()` method, but definitely double check that.

src/interceptors/fetch/index.ts
kettanaito: Do we need this change for type compliance only? Afaik Fetch supports `URL` as input. Is that no longer the case for Node.js fetch?

test/jest.browser.setup.ts
kettanaito: Do I get this right that we can remove this mutation once your change to `page-with` is published?

kettanaito:
Btw, Node 18 pipeline on CI is passing 🎉 I believe it wasn't running because all CIs need to be approved my maintainers before running. I've encountered a flaky browser test in Node 14, rerun, and now the entire CI is green.

test/patched/PageWithPreviewServer.ts
```diff
@@ -0,0 +1,36 @@
...
+  // bug: `PreviewServer.listen()` serializes IPv6 hosts to invalid URL (missing square brackets)
+  // see: node_modules/page-with/lib/server/PreviewServer.js / PreviewServer.listen
+  const conn = server.connectionInfo
+
+  // fix: re-serialize any URL containing an IPv6 host without surrounding brackets
+  if (conn.host.includes(':') && !conn.url.includes(`[${conn.host}]`)) {
+    // note: scheme is hardcoded to `http` to match `PreviewServer.listen()`
+    server!.connectionInfo!.url = `http://[${conn.host}]:${conn.port}`
+  }
+}
```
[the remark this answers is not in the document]
kettanaito: I've released `page-with@0.6.0`. Would you mind updating this pull request please?

kettanaito:
I've released `@open-draft/test-server@0.5.0` that adds IPv6 support similar to what you've added in `page-with`.

milesrichardson:
Sorry for the annoying delay! This should be ready to merge now.

I (carefully) dropped the commits adding patches for bugs in upstream packages, and replaced them with new commits upgrading the packages. All went smoothly, and now this PR is much cleaner.

The tests are passing for [this branch in my fork](https://github.com/milesforks/msw-interceptors/tree/chore/migrate-to-node-v18), but here you will need to manually approve the workflow again, since I force-pushed the commit. 

Reminder that this PR only adds support for running existing tests in a Node v18 environment. I have a subsequent PR which is based on this that adds support for node-native fetch. I'm opening that next, and will edit this comment with a link. In the meantime, you can [see the changes in my fork](https://github.com/milesforks/msw-interceptors/pull/5).

test/jest.browser.setup.ts
milesrichardson: Yes, that's correct! It's now gone, since it was part of the dropped commit adding the patch. :)

test/patched/PageWithPreviewServer.ts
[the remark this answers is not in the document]
milesrichardson: All set! See root-level comment for details.

src/interceptors/fetch/index.ts
milesrichardson: Good question. I'm looking more closely at this now. Reverting this change does not cause any tests to fail on my local (non-IPv6) system. I also checked in GitHub Actions, and no tests failed there either. So we could revert it, but let me first check if it breaks any tests in the next branch following this one. I *think* it could fix a real issue that was only *exposed* by a TypeScript error, but I don't have any regression test to prove that.

The reason it's a type error in VSCode but not any of the build scripts is that VSCode uses version 4.7.3 while this package uses version 4.3.5, and the type error is related to the latest version of the `fetch` API from `lib.dom.d.ts`. So the type error is likely indicating a real bug, even if there are no tests for it. Likely the bug would surface in some environments (browser vs. Node) but not others.

Specifically, the error is:

```
Property 'url' does not exist on type 'Request | URL'.
  Property 'url' does not exist on type 'URL'.ts(2339)
```

as reported in VSCode:

image

Note that it's inferring the type of `input` as `RequestInfo` which it infers from the definition of `globalThis.fetch` in the latest TypeScript `lib.dom.d.ts`:

```ts
declare function fetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response>;
```

visually:

image

Whether or not this indicates a bug at runtime depends on the `fetch` API provided by the environment. If we're using a subset of that API, then we should use module augmentation to keep the TypeScript API consistent with what we expect at runtime (although admittedly it won't be an issue until we upgrade the version of TypeScript used in the build scripts).

test/patched/OpenDraftTestServer.ts
milesrichardson: I'm resolving this comment since the upstream package was fixed and it's no longer relevant. AFAICT the fix was correct (replacing patch with the upgraded package caused no regressions).

src/interceptors/fetch/index.ts
milesrichardson: Ok, so reverting this change does have an effect on a downstream dependency - namely, the msw branch where I'm adding support for node-native interception. It causes [tests to fail on IPv6 systems](https://github.com/milesforks/msw/actions/runs/3154340675/jobs/5131825375#step:7:16) that were otherwise already fixed with [this commit](https://github.com/milesforks/msw/pull/2/commits/4ded61dcb25ccac51c86d4bd52275697bf9e408d).

So, from a backwards compatibility perspective, it would be safer to revert this change. But from a forward compatibility perspective, we need it if we want to operate properly on RequestInfo objects in Node 18. The balance seems to weigh in favor of keeping the commit.

If you're okay with keeping this, then the PR is ready to merge

mdingena:
I'm eagerly awaiting this to drop. 👀

carlosrsabreu:
> I'm eagerly awaiting this to drop. 👀

Me as well 👀

kettanaito (approved):
These changes are absolutely stunning! Thank you for making this happen, @milesrichardson 👏 

There are a few issues with `@remix-run/web-fetch` imports coming from the main:

```
    /mswjs/interceptors/node_modules/@remix-run/web-fetch/src/lib.js:4
    export { ReadableStream, Blob, FormData  } from './package.js';
    ^^^^^^

    SyntaxError: Unexpected token 'export'

    > 1 | import { Response } from '@remix-run/web-fetch'
        | ^
      2 | import { stringToHeaders } from 'headers-polyfill'
      3 |
      4 | export function createResponse(

      at Runtime.createScriptFromCode (../node_modules/jest-runtime/build/index.js:1796:14)
      at Object.<anonymous> (../src/interceptors/XMLHttpRequest/utils/createResponse.ts:1:1)
```

If you have any suggestion about why this happens, I'd be thankful. I will look into those as well so we could have the tests passing and merge this.

kettanaito:
Hey, @milesrichardson. I'm trying to push a rebased version of this branch into your fork but getting a permissions error:

```
ERROR: Permission to milesforks/msw-interceptors.git denied to kettanaito.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Could you please take a look at your fork? Seems like it needs to be made accessible for external contributors or something so I could update it. Thanks.

mattcosta7:
> These changes are absolutely stunning! Thank you for making this happen, @milesrichardson 👏 
> 
> There are a few issues with `@remix-run/web-fetch` imports coming from the main:
> 
> ```
>     /mswjs/interceptors/node_modules/@remix-run/web-fetch/src/lib.js:4
>     export { ReadableStream, Blob, FormData  } from './package.js';
>     ^^^^^^
> 
>     SyntaxError: Unexpected token 'export'
> 
>     > 1 | import { Response } from '@remix-run/web-fetch'
>         | ^
>       2 | import { stringToHeaders } from 'headers-polyfill'
>       3 |
>       4 | export function createResponse(
> 
>       at Runtime.createScriptFromCode (../node_modules/jest-runtime/build/index.js:1796:14)
>       at Object.<anonymous> (../src/interceptors/XMLHttpRequest/utils/createResponse.ts:1:1)
> ```
> 
> If you have any suggestion about why this happens, I'd be thankful. I will look into those as well so we could have the tests passing and merge this. 

Could this be something we need to add to 'transformIgnoreModules' in jest config, so have jest handle the esm->cjs transform  (viewing on mobile, but I didn't see that edited,and potentially explains this)

kettanaito:
@mattcosta7, thanks for the insight, Matthew. It does look like module transformation issue. However, other users reported that this also happens when they add MSW to a fresh Next project (see https://github.com/mswjs/msw/discussions/1464#discussioncomment-4159031). It makes me thing we should handle the ESM -> CJS thing when bundling `@remix-run/web-fetch`, should we not? 

From the user's standpoint, if they don't explicitly decide to use ESM, I wouldn't ask them to add any additional configuration to handle ESM that comes from MSW.

mattcosta7:
> @mattcosta7, thanks for the insight, Matthew. It does look like module transformation issue. However, other users reported that this also happens when they add MSW to a fresh Next project (see [mswjs/msw#1464 (comment)](https://github.com/mswjs/msw/discussions/1464#discussioncomment-4159031)). It makes me thing we should handle the ESM -> CJS thing when bundling `@remix-run/web-fetch`, should we not?
> 
> From the user's standpoint, if they don't explicitly decide to use ESM, I wouldn't ask them to add any additional configuration to handle ESM that comes from MSW.

We could try including it in the bundle we ship to consumers, which would avoid that issue - although that requires a build process for this library, that we don't have yet (since we don't currently bundle it), and likely some additional work to handle the multiple entrypoints necessary to only use specific interceptors in specific environments?

kettanaito:
@mattcosta7, I think you're right. We can add a build process to this library as well. I think `tsup` + `esbuild` like we have in MSW should be a good start. Does this sound interesting to you to work on?

kettanaito:
@Y2zz, no there's not. There's a discussion above suggesting we need additional work to make this support happen. As absolutely everything that I ship, I do so if it's work personally interesting for me and I have time for it. 

Anyone can help with this support, just as a few contributors above already did. Bundling the library is a standalone change and can be added and merged separately, unblocking Node 18 support.

mattcosta7:
> @mattcosta7, I think you're right. We can add a build process to this library as well. I think `tsup` + `esbuild` like we have in MSW should be a good start. Does this sound interesting to you to work on?

I'd be happy to work on it, but not sure what level of time I have over the next month - If anyone else has time sooner, I'd be happy to offer support though!

cdimitroulas:
@mattcosta7 I might have some time to look into it. Could you give some guidance on what exactly we want here?

I see that in `msw` the tsup config has 4 main sections -> `main`, `iife` (browser), `node` and `native` + the type definitions. Do we need to follow the same kind of setup here for the interceptors lib?

mattcosta7:
> @mattcosta7 I might have some time to look into it. Could you give some guidance on what exactly we want here?
> 
> I see that in `msw` the tsup config has 4 main sections -> `main`, `iife` (browser), `node` and `native` + the type definitions. Do we need to follow the same kind of setup here for the interceptors lib?

I don't think we need those specifically, but we probably need to consider separate imports (with shared code between them) for each of the individual interceptors, since they have code that might not evaluate in every environment, and a single file bundle won't necessarily shake things out.  

https://github.com/mswjs/interceptors/tree/main/src/interceptors

I think we'd need to move the polyfills we want to bundle into `devDependencies` (like this web-std one) otherwise `tsup` will treat them as external.

cdimitroulas:
@mattcosta7 first roadblack - the `tsconfig.json` in the interceptors lib targets ES5, but esbuild doesn't support such a target. A basic test with tsup yields various errors like "Transforming const to the configured target environment ("ES5") is not supported yet". How should we proceed here?

cdimitroulas:
@mattcosta7 I've opened a [draft PR](https://github.com/mswjs/interceptors/pull/313) for this work with some initial config based on our discussions above. Let's continue discussing the details of what exactly we want/need there.

kettanaito:
> I think we're probably ok to at least es2019 (or even esnext)

I'd go for `esnext` 👍 

@cdimitroulas, thank you for opening that pull request! I will take a look soon.

milesrichardson:
hey all - sorry for the delay on my end. I have this all working, but I haven't rebased it for a few weeks. I diagnosed the memory leak problem as being due to the number of tests being run (each test is leaking some memory for some reason), which could be fixed by just splitting it into two runs. 

Otherwise, I haven't touched this since I saw some PRs in progress that mine would likely conflict with, and rebasing these PRs is pretty annoying since it touches multiple repositories

I'm not sure I will get to clean this up this week, or even if it will still be necessary. But all written in my comment above still applies.

In the meantime, if you want to do something **really hacky** you can use the tarball of `msw` built from my PR. If you're using Yarn you can just put this in your package.json:

```
"msw": "https://github.com/milesforks/msw/blob/released/patched/intercept-node-v18-native-fetch/unreleased-msw-1190f5560178babd232abbb9a96ce8ba4f1abfb1.tgz?raw=true",
```

Just to emphasize, this is really hacky and I offer no guarantees about the stability of that URL. That tarball is a build of my `msw` fork at this commit: https://github.com/milesforks/msw/commits/1190f5560178babd232abbb9a96ce8ba4f1abfb1 and it transitively depends on my fork of `msw-interceptors` via:

```
    "@mswjs/interceptors": "https://github.com/milesforks/msw-interceptors/blob/released/patched/pr-283-support-node-v18/msw-interceptors-6aaf5a14562f9deffc044aa3b2872e649612f957.tgz?raw=true",
```

which is a tarball built from https://github.com/milesforks/msw-interceptors/commits/6aaf5a14562f9deffc044aa3b2872e649612f957

milesrichardson:
Also, I just saw the comment from @kettanaito about my fork. If you still want to help with that, I am adding you as an external contributor now (to both the interceptors and msw forks). feel free to rebase whatever you need (although i'd prefer if you push to a new branch where possible, since I'm not sure where my local copy is)

justrhysism:
Is it just the rebasing which needs to be fixed for this PR to get merged? I'm interested in pitching in to get this happening.

kettanaito:
I think we're about done with distributing this package as CJS _and_ ESM in #313, which should be merged before this. I will rebase this branch and see if there are any loose ends.

milesrichardson:
@kettanaito Feel free to email me if you need help with this (miles [at] splitgraph.com). I don't have the bandwidth to do all the rebasing myself atm but happy to help if you get stuck with my code

kettanaito:
## Released: v0.21.0 🎉

This has been released in v0.21.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.21.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.21.0)

---

kettanaito:
Addressed and merged this in #349. We are dropping support for Node < 18 from the next minor release. No more polyfills.

kettanaito:
## Released: v0.22.0 🎉

This has been released in v0.22.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.0)

---

PR #292, opened 2022-10-12
feat: adopt Fetch API Request and Response

### LINKED ISSUE #158: Arraybuffer response in XMLHttpRequest assumes always to be a string?
I am trying to mock a XMLHttpRequest which get image file data as an arraybuffer. However, I observed some unexpected behaviors with the mocked response.

1. The response attribute always set to be a Uint8Array rather than ArrayBuffer.
2. Even if the body of MockedResponse is already set an arraybuffer, 'bufferfrom' function assumes it as a string and encodes it. It results an encoded string like '<Object ArrayBuffer>'.
3. ResponseText attribute is directly set to be the original response body. So if I send an arraybuffer, response attribute is a string and responseText attribute is an arraybuffer.

### LINKED ISSUE #187: XMLHttpRequest: Support "upload" events

### LINKED ISSUE #281: XMLHttpRequest: Keep "status" as 0 until request succeeds
- Source: https://github.com/mswjs/interceptors/pull/274#issuecomment-1189479436

The `this.status` value of XMLHttpRequest must be `0` until the request succeeds. If it fails (request exception), it must return `0`.

### LINKED ISSUE #288: Expose `formData`  method on incoming request from the `interceptor`
I was using Remix and had tried to do the following in my server.js file (I'm using Remix express template)

```ts
interceptor.on('request', async (request) => {
  if(request.method === "POST"){
      const body = await request.formData() // ❌ error obviously since its not an supported on [InteractiveIsomorphicRequest](https://github.com/mswjs/interceptors/blob/e305d3815851dbdfd674e2b412f2c2e0288fc1de/src/IsomorphicRequest.ts)
```
In my remix app I was attempting  to capture a request from my server and rewrite the request (proxy the request) so that an external API provider thinks its coming from my proxy server (example-proxy.com). Details here I wrote up a [stackoverflow question](https://stackoverflow.com/questions/73739436/nodejs-server-how-can-i-intercept-a-3rd-party-request-and-forward-it-the-reque)

### Aside
Then I remembered that [formData](https://developer.mozilla.org/en-US/docs/Web/API/Request/formData), [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request), [fetch](https://nodejs.org/de/blog/announcements/v18-release-announce/) were only introduced a few months ago in in [nodejs v18](https://nodejs.org/de/blog/announcements/v18-release-announce/)

Would bringing in `formData` be contingent upon this discussion that Kent wrote or the work @milesrichardson is doing to migrate this library to node v18?
- https://github.com/mswjs/interceptors/pull/283
- https://github.com/mswjs/msw/issues/1404

kettanaito:
This change is a pre-requisite for https://github.com/mswjs/msw/issues/1404.

## Changes

- Adds support for XMLHttpRequest uploads (closes #187). Now, response readable streams are streamed to the consumer, respecting the `progress` event.
- Fixes XMLHttpRequest events order to be correct.
- Fixes XMLHttpRequest events for when the request fails (middleware throws an exception). Updates `readyState` to `DONE`, emits `loadend`.
- Fixes #158 by d52d71d
- Fixes #281 by d8868333ead58620c44894b19146cb98a5a7d29a
- Closes #288 because now the intercepted request has all body reading methods. 

## Dependencies

- [ ] https://github.com/remix-run/web-std-io/pull/21


kettanaito:
The CI is failing due to the `credentials` bug in `@remix-run/web-fetch`. The fix opened in https://github.com/remix-run/web-std-io/pull/21.

kettanaito:
Fetch API support is done here. I'm adding a custom `createRequestWithCredentials` abstraction until Remix respects the credentials property in the `Request` class.

kettanaito:
## Released: v0.18.0 🎉

This has been released in v0.18.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.18.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.18.0)

---

PR #300, opened 2022-11-08
fix(xhr): removes @xmldom/xmldom, uses native DOMParser

### LINKED ISSUE #297: XMLHttpRequestOverride: remove `getResponseXML` and `responseXML`, drop xmldom
XMLHttpRequestOverride contains a property `responseXML`, which appears to have a meaningful value given only the appropriate response content type (`application/xml` or `text/xml`). This suggests that the property is only useful when dealing with XML responses, which would form a smaller proportion to JSON or HTML responses, amongst others.

To support this property however, XMLHttpRequestOverride implements `getResponseXML()`, a thin wrapper around xmldom. This package has a significant size (134kB for 0.8.5) relative to what it is being used for, obliges a dependent to maintain it even if it does not process XML, and is currently maintained only by [one active author](https://github.com/xmldom/xmldom/graphs/contributors).

Given that the few applications who actually depend on this property would in all likelihood explicitly bring in a package to parse and manipulate the response, I would consider it reasonable to propose the removal of `XMLHttpRequestOverride.responseXML` and its supporting cast. Dependents on this property will not be too inconvenienced, as they can trivially build an XML document from `responseText`, even reproducing `getResponseXML()` to do so. 

Most other users of this package would stand to benefit by no longer depending on - and having to maintain - a package of little relevance to them.

kettanaito:
- Closes #297 

## Changes

- `req.responseXML` will not rely on the native (global) `DOMParser`. It's present in the browser and it must be polyfilled in browser-like environments for XHR to work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,6 @@
     "@open-draft/until": "^1.0.3",
     "@remix-run/web-fetch": "^4.3.1",
     "@types/debug": "^4.1.7",
-    "@xmldom/xmldom": "^0.8.3",
     "debug": "^4.3.3",
     "headers-polyfill": "^3.1.0",
     "outvariant": "^1.2.1",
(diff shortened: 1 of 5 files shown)


kettanaito:
## Released: v0.18.3 🎉

This has been released in v0.18.3!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.18.3)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.18.3)

---

Issue #309, opened 2022-11-26
XMLHttpRequest: "responseType" is not being honored when making non-mocked requests


avm99963:
When running the following code in a browser:

``` js
import {XMLHttpRequestInterceptor} from '@mswjs/interceptors/lib/interceptors/XMLHttpRequest';

const URL = 'https://httpbin.org/get';

const interceptor = new XMLHttpRequestInterceptor();
interceptor.apply();

// Send a request expecting to receive an ArrayBuffer.
const xhr = new XMLHttpRequest();
xhr.responseType = 'arraybuffer';
xhr.open('GET', URL);

console.log('xhr.responseType (original):', xhr.responseType);

xhr.addEventListener('load', () => {
  console.log('xhr.responseType (new):', xhr.responseType);
  console.log('xhr.response:', xhr.response);
});

xhr.send();
```

The following output is shown in the JS console:

```
xhr.responseType (original): arraybuffer

xhr.responseType (new): 

xhr.response: {
  "args": {}, 
  "headers": {
    [censored]
  }, 
  "origin": "[censored]", 
  "url": "https://httpbin.org/get"
}
```

I'd expect `xhr.responseType` to continue being `arraybuffer` and `xhr.response` to be an ArrayBuffer.

PS: Thanks for developing this! It's awesome :) I'm trying to use it for [a Chrome extension](https://gerrit.avm99963.com/plugins/gitiles/infinitegforums/) which needs to modify some responses to improve how a website is shown.

avm99963:
See my attempt to fix this here: https://github.com/mswjs/interceptors/compare/main...avm99963:interceptors:b/309

I added a test to `test/modules/XMLHttpRequest/intercept/XMLHttpRequest.test.ts`, but it fails with

> `Error: Uncaught [InvalidStateError: The object is in an invalid state.]`

due to the added line in `src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts` (which I think would fix this).

kettanaito:
Thanks for reporting this, @avm99963! 

It makes me wonder if this logic makes sense: 

https://github.com/mswjs/interceptors/blob/ddf0389c0a0f2f98350f3857d1e301f806c66587/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L419-L423

[...]

PR #310, opened 2022-11-27
fix(XMLHttpRequest): forward "responseType" and "withCredentials" to the bypassed request

avm99963:
Fixes: #309

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -416,12 +416,6 @@ export const createXMLHttpRequestOverride = (
             }
           })
 
-          originalRequest.addEventListener('loadstart', () => {
-            // Forward the response type to the patched instance immediately.
-            // Response type affects how response reading properties are resolved.
-            this.responseType = originalRequest.responseType
-          })
-
           originalRequest.addEventListener('progress', () => {
             this._responseBuffer = concatArrayBuffer(
               this._responseBuffer,
@@ -462,9 +456,13 @@ export const createXMLHttpRequestOverride = (
           this.propagateCallbacks(originalRequest)
           this.propagateListeners(originalRequest)
 
+          // Infer properties from the intercepted XHR instance to the original XHR
+          // instance.
           if (this.async) {
             originalRequest.timeout = this.timeout
           }
+          originalRequest.withCredentials = this.withCredentials;
+          originalRequest.responseType = this.responseType;
 
           /**
            * @note Set the intercepted request ID on the original request
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -61,6 +61,7 @@ export const createXMLHttpRequestOverride = (
     _events: XMLHttpRequestEvent<InternalXMLHttpRequestEventTargetEventMap>[] =
       []
 
+    id: string
     log: Debugger = log
 
     /* Request state */
@@ -123,6 +124,7 @@ export const createXMLHttpRequestOverride = (
     ) => any = null as any
 
     constructor() {
+      this.id = uuidv4()
       this.url = ''
       this.method = 'GET'
       this.readyState = this.UNSENT
@@ -239,7 +241,6 @@ export const createXMLHttpRequestOverride = (
       this.log('request headers', this._requestHeaders)
 
       // Create an intercepted request instance exposed to the request intercepting middleware.
-      const requestId = uuidv4()
       const capturedRequest = new RequestWithCredentials(url, {
         method: this.method,
         headers: this._requestHeaders,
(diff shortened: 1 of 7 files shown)


test/modules/XMLHttpRequest/intercept/XMLHttpRequest.test.ts
```diff
@@ -347,3 +349,29 @@ test('sets "credentials" to "omit" on isomorphic request when "withCredentials"
   const [request] = resolver.mock.calls[0]
   expect(request.credentials).toBe('omit')
 })
+
+test('responds with an ArrayBuffer when "responseType" equals "arraybuffer"', async () => {
```
avm99963, line 353: This test is broken and should be fixed (it fails [as mentioned here](https://github.com/mswjs/interceptors/issues/309#issuecomment-1328052958)).

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -682,5 +609,88 @@ export const createXMLHttpRequestOverride = (
...
+          this.log(
+            'original response headers (normalized)',
+            this._responseHeaders
+          )
+        }
+      })
+
+      request.addEventListener('progress', () => {
+        const chunk = request.response
+        const nextChunk =
```
kettanaito, line 641: @avm99963, so the "Invalid state" error was caused by the interceptor accessing `request.responseText` of the original request when its `responseType` value is anything else but `"text"`. Turns out, that's a no-op (I'd expect `responseText` to return a text representation of the response body but that's not how JSDOM works, at least).

I've added the logic that checks the value of `request.response`, and if it's an array buffer, we skip the encoding part, create a uint8array out of the response buffer, and add it to the internal response buffer representation. That seems to work.

test/modules/XMLHttpRequest/intercept/XMLHttpRequest.test.ts
kettanaito: I've fixed this one, see my comment below.

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -123,6 +124,7 @@ export const createXMLHttpRequestOverride = (
     ) => any = null as any
 
     constructor() {
+      this.id = uuidv4()
```
kettanaito, line 127: I've moved the request ID to the request itself (out from the `send()` method) so it's accessible within this class.

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -682,5 +609,88 @@ export const createXMLHttpRequestOverride = (
...
+        this.setReadyState(this.DONE)
+        this.log(
+          'original request DONE!',
+          this.status,
+          this.statusText,
+          this.response
+        )
+      })
+
+      // Forward request headers.
```
avm99963, line 668: Since we're no longer using `propagateHeaders()`, we could maybe delete the function definition? WDYT?

kettanaito: A good idea.

kettanaito:
Setting the withCredentials property now throws a jsdom exception during tests. It's odd since it fails on the preflight check for Acess-Controlled-Allow-Credentials header missing from the OPTIONS response (preflight) but jsdom never reaches that endpoint.

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
avm99963: Nice catch! So it seems like since we didn't forward `responseType` before, it was always `''`, the default (which is the same as `'text'`), so `responseText` always returned the response text.

LGTM!

avm99963:
I have very little experience with Express, but I think the `cors` package in your test server handles the OPTIONS request. To return that header, it seems like we should use the `{ credentials: true }` configuration for the cors middleware. ([more info](https://www.npmjs.com/package/cors#configuration-options)).

kettanaito:
@avm99963, yeah, that's a good point. But I would still expect my `app.options('/')` route to be hit during that preflight check. I will try your suggestion, it sound like what we need.

src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
kettanaito: Removed.

kettanaito:
I've found a bug in the `@open-draft/test-server` package that prevented custom middleware to take effect if it conflicts with the default middleware, like CORS. Merged the fix, will update and the CI should be green.

avm99963:
Awesome, thank you so much!! :)

kettanaito:
This still needs some work, I will try to continue whenever I have a minute. I like that this unblocks the actual XHR behavior in regard to credentials. A great change.

kettanaito:
## Released: v0.20.0 🎉

This has been released in v0.20.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.20.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.20.0)

---

Issue #52, opened 2020-08-30
Distribute interceptors properly


kettanaito:
## Expected behavior

- Exported under `node-request-interceptor/interceptors/[name]`
- Distributed in a format that would allow them to be imported from both ESM and CJS context (currently importing an interceptor directly from a ESM context doesn't compile, it strips away named export of `SocketPolyfill`).

kettanaito:
I've tried building each interceptor as a separate CJS module, but that produces the same result as with the current setup.

It comes down to which format the entire NRI should be distributed. I believe that should be CJS, as you can use CJS in another CJS context, while in ESM context you are likely to ignore transformation of CJS modules.

kettanaito:
Can use symlinks or empty folders like we do in `msw/node`:

```
interceptors/
  fetch/
    package.json
  ClientRequest/
    package.json
```

Can also use `"exports"` in package.json but TypeScript doesn't play nicely with it.

kettanaito:
We can use "exports" in package.json to do this but it'll only work in ESM, I assume.

kettanaito:
Already implemented in #313.

PR #313, opened 2022-11-30
feat: distribute the library for browser and node in ESM and CJS

cdimitroulas:
re #283

Bundle the library, including `remix-run` and other polyfill libs, so that consumers of the lib can consume it without having to bundle/transpile it themselves.


tsup.config.ts
```diff
@@ -0,0 +1,48 @@
...
+  {
+    name: 'index',
+    entry: ['./src/index.ts'],
+    outDir: './lib',
+    format: ['esm', 'cjs'],
+    legacyOutput: true,
+    sourcemap: true,
+    clean: true,
+    bundle: true,
+    splitting: false,
```
mattcosta7, line 13: does splitting share code between these entrypoints?  if so, we might want to enable it?

I'm not sure how these output?  are they 4 discrete bundles, or 4 modules with shared code between them? (could probably look at imports after running the build locally). 

I think we want to share most of the stuff between them outside of the interceptor specific files.

tsup.config.ts
```diff
@@ -0,0 +1,48 @@
+import { defineConfig } from 'tsup'
+
+export default defineConfig([
+  {
+    name: 'index',
```
mattcosta7, line 5: I think we have a couple files in msw that are referenced from `interceptors/util` like https://github.com/mswjs/msw/blob/ab8b1accf5db10b30647759dde8c1585b8dc4c14/src/utils/matching/matchRequestUrl.ts#LL2C3-L2C3

do we need to expose those as well, or can we export them from the root index?

cdimitroulas: They are already exported from the root index - I don't know if the reason for importing like this in msw was for tree-shaking purposes? If so we'll need to separate the utils into another bundle which is then re-exported from the index somehow

tsup.config.ts
cdimitroulas: > does splitting share code between these entrypoints? if so, we might want to enable it?

I tried to set this to true (actually that's the default setting) but I couldn't notice any difference in the build output 🤔 I tried to use the `--splitting` [flag](https://tsup.egoist.dev/#code-splitting) but that also didn't seem to make any difference. The shared code seemed to be placed inside each interceptor bundle each time.

I will do some more digging to see how this is supposed to work. 

> I'm not sure how these output? are they 4 discrete bundles, or 4 modules with shared code between them? (could probably look at imports after running the build locally).

Right now it seems they are 4 discrete bundles. This is not what we want obviously, I need to tinker more with it.

cdimitroulas:
OK I switched to using multiple entrypoints in the tsup config and this looks more like what we want.

Here is the structure of the outputted files:
Screenshot 2022-11-30 at 16 43 44

The interceptors import from the chunks which are all the shared bits of code

tsup.config.ts
mattcosta7: cool - i'm not sure either, but if we can use them from another path that might be helpful - but it might be worth adding the entrypoint to be consistent for now, and change imports later if we need to!

mattcosta7:
> OK I switched to using multiple entrypoints in the tsup config and this looks more like what we want.
> 
> Here is the structure of the outputted files: Screenshot 2022-11-30 at 16 43 44
> 
> The interceptors import from the chunks which are all the shared bits of code

Nice! yeah, that's more like what I think we're looking for!

tsup.config.ts
cdimitroulas: Looking at how msw imports these utils more closely, it seems to import from specific files like `interceptors/utils/getCleanUrl.js` and `interceptors/utils/bufferUtils.js`. There is no `utils/index.ts` file which we can use to bundle all the utils together.

Based on this, we would need to add an entrypoint for all the individual files under `utils/` which seems a bit unusual to me. Are we sure we want that? I'd like to get @kettanaito's opinion here as well

kettanaito: Please, let's make those a part of the public API, exporting them from the root of the package. That was a historic oversight on my part and now seems like a good opportunity to finally fix it. Let's not reference things that way. 

So, what I'm proposing is this:

```
import { getCleanUrl, bufferUtils } from '@mswjs/interceptors'
```

kettanaito:
> The interceptors import from the chunks which are all the shared bits of code

@cdimitroulas, _This_ is fantastic! 

While we're at it, folks, what is your opinion on #52? I think it would be great to be able to import interceptors directly from the root of the package like so:

```js
import { ClientRequestInterecptor } from '@mswjs/interceptors/ClientRequestInterceptor'
```

This is possible via `exports` in `package.json` but I don't know if that will be universally supported by all consumers (CJS/ESM). This can also be done entirely separately from this change, I'm just asking for your opinion on this.

package.json
```diff
@@ -49,11 +50,13 @@
     "body-parser": "^1.19.0",
     "commitizen": "^4.2.4",
     "cors": "^2.8.5",
+    "cross-env": "^7.0.3",
     "cz-conventional-changelog": "3.3.0",
     "express": "^4.17.3",
     "express-rate-limit": "^6.3.0",
     "follow-redirects": "^1.15.1",
     "got": "^11.8.3",
+    "headers-polyfill": "^3.1.0",
```
kettanaito, line 59: Moving it to dev dependencies gets it bundled, is that right? 

I wonder if we have any issues with `headers-polyfill` in this regard. I thought of bundling ESM-only dependencies as a way to enable this library for legacy consumers. But headers polyfill distributes like both CJS and ESM, does it not? 🤔

package.json
```diff
@@ -62,18 +65,18 @@
...
     "ts-jest": "^27.1.1",
+    "tsup": "^6.5.0",
     "typescript": "4.4.4",
-    "wait-for-expect": "^3.0.2"
+    "wait-for-expect": "^3.0.2",
+    "web-encoding": "^1.1.5"
   },
   "dependencies": {
     "@open-draft/until": "^1.0.3",
     "@remix-run/web-fetch": "^4.3.1",
```
kettanaito, line 75: We need to ensure that we bundle this Fetch polyfill from Remix. Should we move it to dev dependencies as well?

tsup.config.ts
```diff
@@ -0,0 +1,59 @@
+import { defineConfig } from 'tsup'
+
+export default defineConfig({
+  entry: [
+    "./src/index.ts",
+    "./src/interceptors/ClientRequest/index.ts",
+    "./src/interceptors/fetch/index.ts",
+    "./src/interceptors/XMLHttpRequest/index.ts"
+  ],
+  outDir: "./lib"
```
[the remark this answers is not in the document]
mattcosta7: yep I think so

mattcosta7:
> > The interceptors import from the chunks which are all the shared bits of code
> 
> @cdimitroulas, _This_ is fantastic!
> 
> While we're at it, folks, what is your opinion on #52? I think it would be great to be able to import interceptors directly from the root of the package like so:
> 
> ```js
> import { ClientRequestInterecptor } from '@mswjs/interceptors/ClientRequestInterceptor'
> ```
> 
> This is possible via `exports` in `package.json` but I don't know if that will be universally supported by all consumers (CJS/ESM). This can also be done entirely separately from this change, I'm just asking for your opinion on this.



> > The interceptors import from the chunks which are all the shared bits of code
> 
> @cdimitroulas, _This_ is fantastic!
> 
> While we're at it, folks, what is your opinion on #52? I think it would be great to be able to import interceptors directly from the root of the package like so:
> 
> ```js
> import { ClientRequestInterecptor } from '@mswjs/interceptors/ClientRequestInterceptor'
> ```
> 
> This is possible via `exports` in `package.json` but I don't know if that will be universally supported by all consumers (CJS/ESM). This can also be done entirely separately from this change, I'm just asking for your opinion on this.

I think this should work everywhere that `package.json exports` work. 

We may need to add fake folders/package.json for older versions of typescript and node to handle, but otherwise i think this is the path we should be on here.

we should add a block like this to the package.json (once we configure both esm and cjs builds from tsup)

```
"exports": {
    ".": {
        "types": "...path to index.d.ts for root package",
        "require": "...path to cjs root pacakge",
        "default": "...path to esm root package"
    },
    "./ClientRequest": {
       ...same format as above
    },
    "./XMLHttpRequest": {
       ...same format as above
    },
    "./fetch": {
       ...same format as above
    }
}
```

then we'll probably also need a folder for each of these (empty, just with a package.json)
`ClientRequestIntercepot`, `XMLHTTPRequest` and `fetch` with the `main, types, and module` fields corresponding to the same data in the exports blocks.

tsup.config.ts
cdimitroulas: `getCleanUrl`, `encodeBuffer` and `decodeBuffer` are already part of the public API exported from the root of the package so no change is needed unless there are more utils that should be part of the public API

package.json
cdimitroulas: Indeed it does distribute both CJS and ESM

tsup.config.ts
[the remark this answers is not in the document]
cdimitroulas: What format do you want this outputted in exactly?

For example, it can look like this:
Screenshot 2022-12-02 at 11 24 37

Or like this (tsup's "legacy output"):
Screenshot 2022-12-02 at 11 25 43

I really don't know enough about ESM to judge what's best here

package.json
cdimitroulas: Done

cdimitroulas:
@mattcosta7 I've added the "exports" field to package.json, please take a look and maybe also test out the build locally to see if it matches your expectations.

> I think it would be great to be able to import interceptors directly from the root of the package like so:
>
> `import { ClientRequestInterecptor } from '@mswjs/interceptors/ClientRequestInterceptor'`

I think this should work fine based on the current build output (`yarn build:new` for now, I'll overwrite `yarn build` when we're ready to finalize this PR)

package.json
kettanaito: Let's move it back to `dependencies` then?

package.json
```diff
@@ -31,11 +32,34 @@
...
+      "require": "./lib/index.js",
+      "import": "./lib/index.mjs",
+      "types": "./lib/index.d.ts"
+    },
+    "./lib/interceptors/ClientRequest": {
+      "require": "./lib/interceptors/ClientRequest/index.js",
+      "import": "./lib/interceptors/ClientRequest/index.mjs",
+      "types": "./lib/interceptors/ClientRequest/index.d.ts"
+    },
+    "./lib/interceptors/fetch": {
```
kettanaito, line 46: Should we drop `./lib/interceptors/` from the exports paths? So that you could import interceptors like so:

```js
import X from '@mswjsw/interceptors/ClientRequest'
```

The more we work on this, the more I think it needs to be a separate change. I wouldn't want to ask you to restructure all of your work, so if you feel it's already under substantial progress, we can ship this together with the `remix-run` change. If not, I think we're safer releasing the `exports` changes separately.

kettanaito:
Hey, @cdimitroulas. I was having a couple of busy weeks at work. I've updated all the discussions, summarizing them below:

- Move `headers-polyfill` to `dependencies` as it ships both as CJS and ESM so it's safe to depend on it as-is (no need to bundle it).
- Consider moving the `exports` changes to a separate pull request. I leave this to you and whether you feel comfortable extracting those changes. If not, we can ship it as a part of this change as well, just going to bump a minor version. 

I wouldn't want to introduce empty placeholder directories like @mattcosta7 mentions but I think we need that for full compatibility. Do you know if there's a support map for `exports` in package.json or something? I believe all modern bundlers and the latest Node respects that field but I wonder how far the support goes in terms of previous versions.

package.json
cdimitroulas: I've moved it to a separate PR. Let's tackle that once we get the new bundling merged since I'm not sure if it depends on this work or not.
I've opened a draft PR on my fork just to keep a record of the work: https://github.com/cdimitroulas/interceptors/pull/1

kettanaito: I don't think these changes are dependent. Thanks for moving it away 👍

cdimitroulas:
> Do you know if there's a support map for exports in package.json or something? I believe all modern bundlers and the latest Node respects that field but I wonder how far the support goes in terms of previous versions.

I'm not sure such a thing exists, I did find this helpful SO answer which is relevant:
https://stackoverflow.com/a/74551879/7133623

Not sure about bundler support but I imagine it's as you say and they all support it (at least on more recent versions)

kettanaito:
I've added `RemoteHttpInterceptor` as the entrypoint so it can be imported.

Now, seems like all browser tests are failing because some Node code ends up in `lib/index.js`:

```
    BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
    This is no longer the case. Verify if you need this module and configure a polyfill for it.

    If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { \"http\": require.resolve(\"stream-http\") }'
        - install 'stream-http'
    If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { \"http\": false }",
        "moduleId": "./lib/chunk-EYK25V6X.js",
        "moduleIdentifier": "/mswjs/interceptors/lib/chunk-EYK25V6X.js",
        "moduleName": "./lib/chunk-EYK25V6X.js",
        "moduleTrace": Array [
          [Object],
          [Object],
        ],
        "stack": "ModuleNotFoundError: Module not found: Error: Can't resolve 'http' in '/mswjs/interceptors/lib'
```

@cdimitroulas, do you have some suspicion as to why this happens?

cdimitroulas:
No idea, I'll try to figure it out

cdimitroulas:
It seems remix is configured to automatically deal with this stuff... https://remix.run/docs/en/v1/guides/constraints#server-code-pruning

Screenshot 2022-12-07 at 16 46 43

mattcosta7:
> It's not all the browser tests that are failing, the `fetch` ones are passing. Only the `XMLHttpRequest` ones are failing.
> 
> It seems that a chunk that is split off by the code-splitting that is used by both the `XMLHttpRequest` interceptor and the `ClientRequest` interceptor has this in it:
> 
> ```
> // node_modules/@remix-run/web-fetch/src/headers.js
> var _util = require('util');
> var _http = require('http'); var _http2 = _interopRequireDefault(_http);
> ```
> 
> This suggests that it's somehow related to `remix-run/web-fetch` :/

OOof hmm.  I think since we're bundling this in node, we're probably only getting the node imports for remix-run/fetch. 

I wonder if we actually need a 'browser' build that's separate from the node build, which could potentially be a separate tsup config?   the docs specificy - https://tsup.egoist.dev/#target-environment that target environment is node14 by default?  do we need one that targets browsers to get the browser build of that library? (just guessing)

cdimitroulas:
I think the correct flag is esbuild's `platform` which can be set to `node` or `browser`. Unfortunately, trying to build with `platform: "browser"` just results in the same errors but at build-time instead:
```
ESM Build start
✘ [ERROR] Could not resolve "http"

    src/interceptors/ClientRequest/index.ts:1:17:
      1 │ import http from 'http'
        ╵                  ~~~~~~

  The package "http" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "platform: 'node'" to do that, which will remove this error.

✘ [ERROR] Could not resolve "http"

    src/interceptors/ClientRequest/index.ts:1:17:
      1 │ import http from 'http'
        ╵                  ~~~~~~

  The package "http" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "platform: 'node'" to do that, which will remove this error.

✘ [ERROR] Could not resolve "https"

    src/interceptors/ClientRequest/index.ts:2:18:
      2 │ import https from 'https'
        ╵                   ~~~~~~~

  The package "https" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "platform: 'node'" to do that, which will remove this error.

```

cdimitroulas:
I was able to get the tests to pass, but I'm not sure how the build output should look like when building for node/browser separately.
For now, I just arbitrarily created `lib/node/` and `lib/browser/` and updated the imports for the tests:

Here's a gist showing the changes I made as I don't feel like this is something I want to commit here:
https://gist.github.com/cdimitroulas/caf930091aa7c63fc696e6521ade465a

kettanaito:
> It's not all the browser tests that are failing, the fetch ones are passing. Only the XMLHttpRequest ones are failing.

This is a fantastic find. 

> I wonder if we actually need a 'browser' build that's separate from the node build, which could potentially be a separate tsup config?

This may be the case. We use `@remix-run/web-fetch` only in Node to polyfill Fetch API primitives. I don't believe that package is ever meant to run in the browser, as it's a Node.js polyfill for Fetch API. 

I think we do need to ship Node and browser builds separately, which means that XMLHttpRequest and Fetch interceptors should be compiled to both Node (depends on web-fetch) and browser versions (doesn't need web-fetch, uses browser API). 

I believe a separate config is the way here, as we mustn't compile `ClientRequest` interceptor since it's meant for Node-only execution. That's precisely why those errors happen when you tried setting the platform to "browser", @cdimitroulas. 

Something like this should work?

```js
import { defineConfig } from "tsup";

export default defineConfig([
  {
    name: "node",
    entry: [
      "src/index.ts",
      "src/interceptors/ClientRequest",
      "src/interceptors/XMLHttpRequest",
      "src/interceptors/fetch",
    ],
    platform: "node",
    outDir: "./lib",
    format: ["cjs", "esm"],
  },
  {
    name: "browser",
    entry: [
      "src/index.ts",
      "src/interceptors/XMLHttpRequest",
      "src/interceptors/fetch",
    ],
    platform: "browser",
    outDir: "./lib",
    format: ["cjs", "esm"],
  },
]);
```

The fact that we depend on `@remix-run/web-fetch` through direct imports may need to change as well. What I used to do in some other projects is this:

```js
// some-module.ts
// Imagine a module (e.g. an interceptor) that imports
// some fetch api utils.
import { Request } from './fetch-api'
```

```js
// fetch-api.ts
// Define those fetch api utils for Node.
// Here, depend on the polyfill.
export * from '@remix-run/web-fetch'
```

```js
// fetch-api.browser.ts
// Then, create a browser-side version of these utils,
// reusing the native browser API.
export {
  Request,
  Response, 
  // ...
}
```

Lastly, instruct the code to import different `fetch-api` based on the environment where it runs:

```package.json
{
  "browser": {
    "./lib/fetch-api.js": "./lib/fetch-api.browser.js"
  },
}
```

There may be a better way of doing this since I'm not a fan of re-exporting browser globals from `fetch-api.browser.js`. I suppose we can inject `@remix-run/web-fetch` as a polyfill on build-time and rely on Fetch primitives as if they are globals:

```js
// Reference globalThis.Request which is:
// A. A regular browser global if run in the browser;
// B. The same export from @remix-run/web-fetch in Node.js.
new Request('/foo')
```

Maybe there are other options you have in mind, please share them.

cdimitroulas:
>  I suppose we can inject @remix-run/web-fetch as a polyfill on build-time and rely on Fetch primitives as if they are globals:

This sounds better and simpler to me than the alternative you proposed

cdimitroulas:
Unfortunately I don't think the config you proposed will work as both node/browser configs are outputting the same files to the same place (e.g. "lib/interceptors/fetch.js" is an output from both configs).
We'll need something like `lib/interceptors/node/fetch.js` and `lib/interceptors/fetch.js` to distinguish them unless I'm missing some trick

cdimitroulas:
@kettanaito @mattcosta7 do you have any thoughts/suggestions relating to my previous comment?

kettanaito:
@cdimitroulas, you're right about the emitted files. 

I will give this some thought and return to this next year. I'm extremely thankful for all your work on this, I'm looking forward to finalizing it together 🎄

cdimitroulas:
Perfect, enjoy the holiday season 🎅 !

cdimitroulas:
@kettanaito happy new year, hope you enjoyed some nice relaxing holiday time. Let me know when you're ready and have some thoughts about how we should proceed here :)

kettanaito:
Hey, folks. 

I've pushed a few changes to this branch. Mainly, I've added the `exports` file to package.json and used [RxJS](https://github.com/ReactiveX/rxjs/blob/630d2b009b5ae4e8f2a62d9740738c1ec317c2d5/package.json) as an inspiration on how to structure it. With these changes, the library now distributes CJS and ESM and I've verified it in CJS/ESM Node.js and CJS/ESM TypeScript example modules. 

What I haven't tested yet is the browser support. I have the "browser" field listed in relevant exports, like "XMLHttpRequest" and "fetch" but I haven't tested whether that works at all. 

If somebody has a spare minute, I'd appreciate it if they conducted an in-browser test and shared their findings. Thanks!

cdimitroulas:
@mattcosta7 are you familiar with how to do an in-browser test? If you are, it would be great if you could look into that as it will take me a while to figure out how to set that up and I don't have much time this week or next week.

kettanaito:
@cdimitroulas, I think we can use the existing browser tests in this repo. Although they don't feature ESM and CJS, I believe they should still pass.

Here's how you can run it:

```js
yarn install
yarn build
yarn test:integration:browser <PATH_TO_TEST_SUITE>
```

You can see some of the browser integration tests failing. I suspect that's because they cannot import the right thing, and end up importing the Node bundle.

kettanaito:
Okay, so I've got the browser tests working. The culprit was of two parts:

1. The browser `*.runtime.js` files were importing interceptors at the old path: `/lib/interceptors/X`. Changed that to `@mswjs/interceptors/X`.
2. Our browser testing config (webpack) doesn't support ESM so it was ignoring the "exports" field in root-level package.json. I had to configure `./XMLHttpRequest/package.json` correctly. Since those root-level directories are _fallbacks_, I also removed the "exports" field in them as well since it won't make sense (if a bundler is referencing those package.json's it doesn't support "exports"). 

Now the CI should pass. The only thing remaining is to test the ESM imports in the browser. I will see what effort it would require to support ESM as a part of our testing setup (ideally, on demand!). But a manual test for now may also do.

kettanaito:
Hey, @cdimitroulas 👋 Please, would you have a minute to give these changes a review? The main changes are in `package.json`, most of the diff noise is about updating the import paths in tests.

kettanaito:
Oh, one more test to conduct is to link this build of the Interceptors to MSW and see its test report.

kettanaito:
Testing this on MSW (after some import path changes), the IIFE build is failing because it attempts to import from `/lib/node` for some reason:

```
Error: Build failed with 4 errors:
../interceptors/lib/node/chunk-XXHY6MZ6.js:4507:20: ERROR: Could not resolve "http"
../interceptors/lib/node/chunk-XXHY6MZ6.js:4510:22: ERROR: Could not resolve "stream"
../interceptors/lib/node/chunk-XXHY6MZ6.js:5047:22: ERROR: Could not resolve "crypto"
../interceptors/lib/node/chunk-XXHY6MZ6.js:5975:19: ERROR: Could not resolve "url"
    at failureErrorWithLog (/Projects/mswjs/msw/node_modules/esbuild/lib/main.js:1624:15)
    at /Projects/mswjs/msw/node_modules/esbuild/lib/main.js:1266:28
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

That's despite the `platform: "browser"` set for the IIFE build config of TSUP in MSW:

https://github.com/mswjs/msw/blob/2ae9fa9e4993233832e205d4d3f2f1a7e4b3b3f1/tsup.config.ts#L29

I wonder if it's the problem with this change or not 🤔

package.json
```diff
@@ -2,47 +2,52 @@
...
         "require": "./lib/browser/interceptors/fetch/index.js",
-        "default": "./lib/browser/interceptors/fetch/index.mjs"
+        "default": "./lib/browser/interceptors/fetch/index.js",
+        "import": "./lib/browser/interceptors/fetch/index.mjs"
       }
     },
     "./RemoteHttpInterceptor": {
       "types": "./lib/node/interceptors/RemoteHttpInterceptor.d.ts",
-      "node": "./lib/node/interceptors/RemoteHttpInterceptor.js",
+      "default": "./lib/node/interceptors/RemoteHttpInterceptor.js",
```
mattcosta7, line 48: I think default has to come last in the object

These are resolved in the order they're listed in the object, so we might need 

```
browser
require
default (or import)
```

package.json
```diff
@@ -2,47 +2,52 @@
...
+  "module": "./lib/node/index.mjs",
   "types": "./lib/node/index.d.ts",
   "exports": {
     ".": {
       "types": "./lib/node/index.d.ts",
-      "node": "./lib/node/index.js",
+      "default": "./lib/node/index.js",
       "require": "./lib/node/index.js",
-      "default": "./lib/node/index.mjs"
+      "import": "./lib/node/index.mjs"
```
mattcosta7, line 13: I think we need default to come last - these are matched in order from top to bottom, so default will always beat out require/import in this case (and might throw in some tools?)

mattcosta7:
> Testing this on MSW (after some import path changes), the IIFE build is failing because it attempts to import from `/lib/node` for some reason:
> 
> ```
> Error: Build failed with 4 errors:
> ../interceptors/lib/node/chunk-XXHY6MZ6.js:4507:20: ERROR: Could not resolve "http"
> ../interceptors/lib/node/chunk-XXHY6MZ6.js:4510:22: ERROR: Could not resolve "stream"
> ../interceptors/lib/node/chunk-XXHY6MZ6.js:5047:22: ERROR: Could not resolve "crypto"
> ../interceptors/lib/node/chunk-XXHY6MZ6.js:5975:19: ERROR: Could not resolve "url"
>     at failureErrorWithLog (/Projects/mswjs/msw/node_modules/esbuild/lib/main.js:1624:15)
>     at /Projects/mswjs/msw/node_modules/esbuild/lib/main.js:1266:28
>     at processTicksAndRejections (node:internal/process/task_queues:96:5)
> ```
> 
> That's despite the `platform: "browser"` set for the IIFE build config of TSUP in MSW:
> 
> https://github.com/mswjs/msw/blob/2ae9fa9e4993233832e205d4d3f2f1a7e4b3b3f1/tsup.config.ts#L29
> 
> I wonder if it's the problem with this change or not 🤔

Mentioned a couple places, where I'm hoping this is mostly a matter of ordering in the exports fields (assuming those are used when bundling with tsup in this way), but maybe ordering is casuing the node version to get picked up, because 'browser' fields are after 'default' and 'import' fields, so the node bundles are getting chosen instead?

cdimitroulas:
@mattcosta7 I reordered the keys in the exports as per your suggestion

kettanaito:
@mattcosta7, thanks for the review! And @cdimitroulas, thank you so much for addressing those points so quickly! I've rebased this branch, and solved the package.json conflict. Will continue with testing in MSW, let's see if the order of those fields solved the issue for the IIFE build.

package.json
kettanaito: Didn't know about that. Thanks, @mattcosta7. Do you happen to know where this order is defined? I didn't see it in the Node's ESM docs.

mattcosta7: Within the "exports" object, key order is significant. During condition matching, earlier entries have higher priority and take precedence over later entries. The general rule is that conditions should be from most specific to least specific in object order.

https://nodejs.org/api/packages.html#conditional-exports

kettanaito: Fantastic. Thank you!

cdimitroulas:
@kettanaito did the latest changes fix the problems you saw when testing in MSW?

kettanaito:
Hey, @cdimitroulas. I haven't tested that yet. Planning on returning to open source this/next week, need to arrange a few other things. 

The changes to Interceptors look good from what I can gather. The order sensitivity that @mattcosta7 mentioned was a great addition to this, now to see what exactly gets bundled in when tsup builds into IIFE. For some reason, I suspect it doesn't respect the `exports` field despite the platform being the "browser". I'd expect it to respect that.

kettanaito:
The IIFE integration test passes in MSW with these changes to Interceptors linked:

```
 PASS  test/msw-api/distribution/iife.test.ts
  ✓ supports the usage of the iife bundle in a <script> tag (1169 ms)
```

🎉

kettanaito:
There are, however, failing unit tests because somewhere in Interceptors we rely on `TextEncoder`, which Jest just bluntly ignores (Jest is notoriously ~~bad~~ strict with ignoring accepted Node.js standard API).

```
 FAIL  src/utils/logging/serializeRequest.test.ts
  ● Test suite failed to run

    ReferenceError: TextEncoder is not defined

      1 | "use strict";Object.defineProperty(exports, "__esModule", {value: true});// node_modules/web-encoding/src/lib.mjs
    > 2 | var Encoder = TextEncoder;
        |               ^
      3 | var Decoder = TextDecoder;
```

kettanaito:
## Solution to `web-encoding`

I just added a manual shim to Interceptors to provide ambiguous export for environments like Jest.

kettanaito:
## Issue: Jest doesn't understand the `exports` field

Looks like Jest doesn't handle `exports` correctly in its resolver:

```
 FAIL  test/msw-api/setup-server/life-cycle-events/on.test.ts
  ● Test suite failed to run

    Cannot find module '@mswjs/interceptors/ClientRequest' from '../lib/node/index.js'

    Require stack:
      /Users/kettanaito/Projects/mswjs/msw/lib/node/index.js
      msw-api/setup-server/life-cycle-events/on.test.ts

    > 1 | import { ClientRequestInterceptor } from '@mswjs/interceptors/ClientRequest'
        |                                          ^
      2 | import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/XMLHttpRequest'
      3 | import { RequestHandler } from '../handlers/RequestHandler'
      4 | import { SetupServerApi } from './SetupServerApi'

      at Resolver.resolveModule (../node_modules/jest-resolve/build/index.js:306:11)
      at Object.<anonymous> (../src/node/setupServer.ts:1:42)
```

- See https://github.com/facebook/jest/issues/9771

The issue appears to be fixed in Jest@28 (27 too?). But jumping two major versions of Jest foreshadows a traumatic experience. I will try that to prove that MSW tests pass with this Interceptors change.

cdimitroulas:
Jumping from Jest v26 -> v27 doesn't seem too bad https://jestjs.io/blog/2021/05/25/jest-27#features-coming-with-breaking-changes

Jumping from v27 -> v28 requires working through [this guide](https://jestjs.io/docs/28.x/upgrading-to-jest28) though

kettanaito:
I wonder if Jest 27 brings in ESM support... The references I find mention Jest 28 if not 29. 

All and all, this looks like an endless loop of testing setup update that I've been postponing for forever. I don't mind doing it in MSW's repo, it just takes time. Everything does. 

Working on that https://github.com/mswjs/msw/pull/1394

kettanaito:
I wonder if we just used `undici` in the Fetch API branch, and removed remix-run dependencies that force us to bundle _everything_ as ESM, would that be less changes all around the stack? 🤔

mattcosta7:
> I wonder if we just used `undici` in the Fetch API branch, and removed remix-run dependencies that force us to bundle _everything_ as ESM, would that be less changes all around the stack? 🤔

offhand, i'm not sure of node support, would we need to lose backward support? (ie, would we only support versions where unidici already exists? and referenceable on globalThis?)

kettanaito:
@mattcosta7, 

> would we need to lose backward support? (ie, would we only support versions where unidici already exists? and referenceable on globalThis?)

Eventually, yes! That's the plan for this autumn when Node 16 goes to pass and we only have the versions of Node that ship global fetch throughout the board. Until then, we need to resort to some polyfill to support Node 16.

kettanaito:
## Trying this with the Fetch API branch

✅ Build
✅ Unit tests
❌ Integration tests

```
 FAIL  test/msw-api/setup-server/scenarios/response-patching.test.ts
  ● Test suite failed to run

    Cannot find module '@mswjs/interceptors/ClientRequest' from '../lib/node/index.js'

    Require stack:
      /mswjs/msw/lib/node/index.js
      msw-api/setup-server/scenarios/response-patching.test.ts

    > 1 | import { ClientRequestInterceptor } from '@mswjs/interceptors/ClientRequest'
```

All the errors seem to have the same origin: they fail to resolve the "exports" to the interceptors.

kettanaito (approved):
@cdimitroulas @mattcosta7 thank you so much for all the work on this. I think the changes are good to go and I'm going to merge them once the CI passes. I still cannot fully test them in MSW due to the old Jest being used there but I've decided to slowly make our way towards that, as updating its test dependencies is a time-consuming endeavor and I don't wish to stall this change any longer.

kettanaito:
## Released: v0.21.0 🎉

This has been released in v0.21.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.21.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.21.0)

---

blaenk:
Is it just me, or does it appear that this change made the presets unreachable now? This sample code no longer works for importing the `browserInterceptors` preset:

```js
import browserInterceptors from '@mswjs/interceptors/lib/presets/browser'

  name: 'my-interceptor',
  interceptors: browserInterceptors,
})
interceptor.on('request', listener)
```

cdimitroulas:
Most probably yes @blaenk - we'll need to add these exports explicitly at the top-level somewhere I think and update the docs to reflect this. Do you have capacity to open a PR on that?

blaenk:
Sorry, I don't have the bandwidth at the moment. For now I pinned to the previous release.

PR #318, opened 2022-12-08
fix(xhr): do not pass response `Content-Encoding` to `TextDecoder`

### LINKED ISSUE #308: Got request is silently failing on compressed data when an interceptor is set up
Hi,

I'm using [Got](https://github.com/sindresorhus/got) and it seems like the request is silently failing while fetching compressed data (like brotli or gzip).
In the request option, Got provides a `decompress` option (default: true) that use [decompress-response](https://github.com/sindresorhus/decompress-response) internally.
Of course, everything is working fine when the interceptor is not set up.

Thanks

Repro: https://replit.com/@qyfjs9bfwd/Interceptor-and-Got-with-compressed-data

```js
import got from 'got'

import { ClientRequestInterceptor } from '@mswjs/interceptors/lib/interceptors/ClientRequest/index.js'
import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/lib/interceptors/XMLHttpRequest/index.js'

  name: 'my-interceptor',
  interceptors: [
  ]
})

interceptor.apply()

  console.log('interceptor listener')
})

// Working
const resp = await got('https://httpbin.org/brotli', { decompress: false })
console.log(resp.body)

// Not working
const resp2 = await got('https://httpbin.org/brotli')
console.log(resp2.body) // Does not show up

// Both requests are working if interceptor.apply() is commented out
```

blaenk:
- Fixes #308 

Thank you so much for working on this project, it is immensely useful!

TLDR: I am using the browser interceptors on a site that is serving responses with `Content-Encoding: br` or `gzip` or `deflate` etc. This breaks every request since `Content-Encoding` is typically for compression algorithms, like `gzip`, like this site responds with, but `TextDecoder` takes a character encoding, like `utf-8`. This breaks every request even if nothing is done by the user because the library logs the response, which implicitly calls `responseText()`. Avoiding passing the `Content-Encoding` to `TextDecoder` fixes the issue.

---

Currently, the interceptors break on every request without me doing anything because the `loadend` event logs the `response` which implicitly calls `responseText()`:

https://github.com/mswjs/interceptors/blob/d8f785d2571a6941fd67e7d2f74c43db6aa528b5/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L447

The `responseText()` method passes the response' `Content-Encoding` value to `decodeBuffer()` which passes it to `TextDecoder`.

However, it seems that the [`Content-Encoding`](https://github.com/mswjs/interceptors/blob/main/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L483-L491) header is conventionally for denoting the response' compression such as `gzip`, `br`, `deflate` etc.

> Content encoding is mainly used to compress the message data without losing information about the origin media type

See the [example syntax](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding#syntax) and [directives](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding#directives).

Meanwhile, `TextDecoder` appears to be more concerned with [character encoding](https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API/Encodings), such as `utf-8`.

For this reason, I recommend not passing through the `Content-Encoding` value to `TextDecoder`, since it leads to errors on every request of this form:

```
Uncaught RangeError: Failed to construct 'TextDecoder': The encoding label provided ('br') is invalid.
```

I confirmed that this change fixes the issue.

Once again thank you for all your great work on this project!

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
@@ -483,11 +483,7 @@ export const createXMLHttpRequestOverride = (
     public get responseText(): string {
       this.log('responseText()')
 
-      const encoding = this.getResponseHeader('Content-Encoding') as
-        | BufferEncoding
-        | undefined
-
-      return decodeBuffer(this._responseBuffer, encoding || undefined)
+      return decodeBuffer(this._responseBuffer)
     }
 
     public get response(): unknown {
--- a/test/modules/XMLHttpRequest/intercept/XMLHttpRequest.browser.test.ts
+++ b/test/modules/XMLHttpRequest/intercept/XMLHttpRequest.browser.test.ts
@@ -7,7 +7,6 @@ import { HttpServer } from '@open-draft/test-server/http'
 import { RequestHandler } from 'express-serve-static-core'
 import { createBrowserXMLHttpRequest, XMLHttpResponse } from '../../../helpers'
 import { headersContaining } from '../../../jest.expect'
-import zlib from 'zlib'
 
 const httpServer = new HttpServer((app) => {
   const requestHandler: RequestHandler = (_req, res) => {
@@ -16,15 +15,6 @@ const httpServer = new HttpServer((app) => {
 
   app.get('/user', requestHandler)
   app.post('/user', requestHandler)
-
-  const handleCompressedRequest: RequestHandler = (_req, res) => {
-    res
-      .status(200)
-      .setHeader("Content-Encoding", "gzip")
-      .send(zlib.gzipSync(Buffer.from('compressed-body'))).end()
-  }
-
-  app.get('/compressed', handleCompressedRequest)
 })
 
 function prepareRuntime() {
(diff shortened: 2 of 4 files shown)


kettanaito (approved):
Thanks for fixing this, @blaenk!

You're totally right, we shouldn't forward `Content-Encoding` to `TextDecoder`, those expect quite different kinds of encoding. 

The fix looks great. One thing I would add is an automated test for this scenario. You can take a look at the existing XHR tests in `test/modules/XMLHttpRequest` or let me know and I will add the test once I have a minute. 

The purpose of the test would be to:

1. Establish a [test server](https://github.com/mswjs/interceptors/blob/d8f785d2571a6941fd67e7d2f74c43db6aa528b5/test/modules/XMLHttpRequest/intercept/XMLHttpRequest.test.ts#L19-L30) that responds with a custom `Content-Encoding` (e.g. compression).
2. Assert that the mocked requests are delivered correctly.

blaenk:
Sure. I took a shot at it. Pardon the force-push, was trying to adapt to the commit lint.

kettanaito:
Awesome job on that test, @blaenk. I've moved it out of the happy path tests into a `regressions` folder and added a Node.js test as well. Verified that those tests fail without your fix. This is ready!

kettanaito:
Updated against the latest main, waiting for the final round of CI before merging this.

kettanaito:
## Released: v0.19.2 🎉

This has been released in v0.19.2!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.19.2)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.19.2)

---

PR #322, opened 2022-12-11
fix(xhr): finalize the response on "load"

### LINKED ISSUE #321: XHR `loadend` event may be too late for interception finalization
Curious if my analysis here is correct. I tried this change locally and it appeared to work for this specific use case but bringing this up for discussion in case I'm misunderstanding something or taking the wrong approach. I'll file a tentative PR soon.

I've been debugging and stepping-through a site that I'm intercepting with this package using a chrome extension to inject my code into the same execution context as the page. To be clear: this strategy does work and I've used it successfully with this same package, it's just that the site is in the process of rolling out a new design/version which is giving me some issues.

Anyway, I _believe_ that I've arrived at the issue (aside from #318 which was also an issue).

It appears that the website has an `onload` event handler for a GraphQL fetcher which essentially checks the `xhr.status`:

```javascript
s.onload = function() {
    var n = {
        status: s.status,
        statusText: s.statusText,
        headers: b(s.getAllResponseHeaders() || "")
    };
    n.url = "responseURL"in s ? s.responseURL : n.headers.get("X-Request-URL");
    var e = "response"in s ? s.response : s.responseText;
    r(new E(e,n))
}
```

Pardon the pretty-printed minified code, this is a website I have no control over.

In the code above, `E`'s constructor is this:

```javascript
function E(n, e) {
    e || (e = {}),
    this.type = "default",
    this.status = void 0 === e.status ? 200 : e.status,
    this.ok = this.status >= 200 && this.status < 300,
    this.statusText = "statusText"in e ? e.statusText : "OK",
    this.headers = new d(e.headers),
    this.url = e.url || "",
    this._initBody(n)
}
```

So note `this.ok` is based on the `xhr.status`. However, at this point, `xhr.status` is `0` according to the debugger. This is causing the higher-level GraphQL code to believe that every request is failing.

If I'm understanding correctly, it seems like this library "finalizes" the intercepted request, wiring in the `.status` field and so on, only at `loadend`:

https://github.com/mswjs/interceptors/blob/3c0b5ec1995b76b6bceb88c08177b4d3654a3995/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L432-L456

Maybe the reason for doing this was because this single handler can handle both success and failure. Unfortunately it seems to mean that the underlying interceptor-unaware code may try to access e.g. the `.status` field in `load` or `error`, which won't be set since `loadend` runs only after that event?

If my analysis is correct, it seems one option might be to factor the finalization code out into a function and call it in handlers for both `load` and `error`. If so, not sure if this should also be done for `abort` and `timeout`.

MDN [appears to confirm](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#events) that `loadend` runs only after these other events:

> Fired when a request has completed, whether successfully (after [load](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/load_event)) or unsuccessfully (after [abort](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort_event) or [error](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/error_event)). Also available via the onloadend event handler property.

Alternatively it occurred to me, if the analysis is correct, that it might be done in a single location in `readystatechange` when `originalRequest.readyState === XMLHttpRequest.DONE` but I have no clue if that's correct. The [spec](https://xhr.spec.whatwg.org/#dom-xmlhttprequest-readystate) appears to say that `DONE` is emitted for success and any kind of error as well. MDN [appears to corroborate this](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState). The existence of https://github.com/mswjs/interceptors/issues/103 means that the finalization code would need some kind of `isFinalized` variable to avoid running it twice (may be worth doing anyway 🤷🏼‍♂️).

blaenk:
This change moves request interception finalization to occur earlier than the current `loadend`. Request interception finalization involves, for example, surfacing the underlying `xhr.status`. This is notable because unaware code may have `onload` handlers that read `xhr.status` to determine whether the request succeeded, but `loadend` explicitly emits _after_ `load`, so in the current state of things, the `xhr.status` will not be initialized and so will be `0`, leading unaware client code to believe that every intercepted request failed.

The approach I took here was to run finalization during `readystatechange` when it is `DONE` instead. I made sure finalization only runs once since I saw another open issue https://github.com/mswjs/interceptors/issues/103 where it gets triggered twice for `DONE`.

This completely fixed the issue I was observing, for whatever that's worth.

Closes #321


src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts
```diff
@@ -413,6 +415,25 @@ export const createXMLHttpRequestOverride = (
                 'original response headers (normalized)',
                 this._responseHeaders
               )
+            } else if (originalRequest.readyState === XMLHttpRequest.DONE) {
```
kettanaito, line 418: What if we move this logic to `load`? I see the issue as this:

- Coupling updates to `this.status`, `this.statusText`, etc. with emitting the "response" event is a mistake. 

Instead, I think this also fixed the problem:

- Update internal response properties in `load`;
- Emit the response event only on `loadend` so it corresponds to successful responses. 

What do you think, @blaenk?

kettanaito: Then we can also drop `finalized` internal state and rely on XHR respecting the spec when calling `load`.

blaenk: I think that was the other option outlined in #321. I defer to you on which one you prefer.

I went with this one for this pass of the PR because it seemed like it would handle it in every case, success or error, whereas the other approach I think I would have to handle it not only in `load` (success) but also in `error` and `abort` and maybe even `timeout`? I'm not sure. This way seemed more fundamental/universal. There's more info in #321 with links to the spec etc.

Curious what you think.

blaenk: As for only emitting the response in `loadend` I think that is captured by the current state of the PR but let me know if I misunderstood you.

kettanaito: Afaik, `loadend` is called _always_, in both success and error scenarios, regardless of the error reason (rejection/timeout):

> The loadend event is fired when a request has completed, whether successfully (after [load](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/load_event)) or unsuccessfully (after [abort](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort_event) or [error](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/error_event)).
— [Source](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/loadend_event)

On the other hand, `load` is called only upon successful response. 

I believe what I'm proposing covers your issue rather well. There aren't any events at which response info is available prior to `load`. We populate that response info on `load`, and we finalize the response on `loadend`, which indicates a successful response. That should work.

blaenk: It seems there may have been a misunderstanding. I do understand that `loadend` runs in both cases. What I am referring to is your suggestion that we set the response info e.g. `.status` in `load`, my concern being that I believe that only runs on 'success', and so then unsuspecting code may have an `error` handler that checks `.status` or another one of those fields and in that case they will be uninitialized too.

Or is it the case that the `error` handler is for a kind of error which _wouldn't_ set those fields like `.status`, and they're only expected to be populated on success/`load`?

So to recap: I am concerned that setting the response info like `.status` only on `load` would mean that someone trying to read those fields from an `onerror` will see those fields as uninitialized. If so, then maybe response-field setting should also be set on other events as well?

Does that make sense now? If not let me know and I'll be happy to rephrase. If you don't believe that is an issue.

blaenk: By the way, to be clear, I am not against your suggestion, as it was one of the options I mentioned in #321, I'm simply bringing up that open question I had to see what you think, but I don't feel strongly one way or the other so I'm happy to implement it that way if you decide that's what you want!

kettanaito: > we set the response info e.g. .status in load, my concern being that I believe that only runs on 'success',

That's the thing, `load` runs only on success but `loadend` runs in both success and failure. That's why the separation like I've suggested makes sense to me. I do believe that internally, what XHR does is exactly this:

```js
if (nextReadyState === this.DONE) {
  this.onloadend?(...)
}
```

> It may be the case that .status can only be expected to be set on a successful load

As we're speaking about this purely in the bypassed requests' case, any error/abort will not give us any status to infer, so when we do that in `loadend`, we will be doing nothing:

```js
originalRequest.addEventListener('loadend', () => {
  // If error happened, "originalRequest.status" is undefined/0/etc.
  // So we don't disrupt the original behavior. If anything, we respect it. 
  this.status = originalRequest.status
})
```

> I'm simply bringing up that open question I had to see what you think,

I think this proposal is great but it can be implemented in simpler terms (for us, as maintainers). As I see it, `readyState === DONE` check is the same as a `loadend` listener. 

The only concern here is the timing when these events run:

- `readystatechange` from X to DONE;
- and the dispatch of the `loadend` listeners.

I'm looking at some examples from the [XMLHttpRequest Specification](https://xhr.spec.whatwg.org/), and based on this error example I see that `readystatechange` always fries before any listeners:

> The request error steps for an [XMLHttpRequest](https://xhr.spec.whatwg.org/#xmlhttprequest) object xhr, event, and optionally exception are:
>
> Set xhr’s [state](https://xhr.spec.whatwg.org/#concept-xmlhttprequest-state) to done.
> 
> Unset xhr’s [send() flag](https://xhr.spec.whatwg.org/#send-flag).
> 
> Set xhr’s [response](https://xhr.spec.whatwg.org/#response) to a [network error](https://fetch.spec.whatwg.org/#concept-network-error).
> 
> If xhr’s [synchronous flag](https://xhr.spec.whatwg.org/#synchronous-flag) is set, then [throw](https://webidl.spec.whatwg.org/#dfn-throw) exception.
> 
> **[Fire an event](https://dom.spec.whatwg.org/#concept-event-fire) named [readystatechange](https://xhr.spec.whatwg.org/#event-xhr-readystatechange) at xhr.**
> 
> At this point it is clear that xhr’s [synchronous flag](https://xhr.spec.whatwg.org/#synchronous-flag) is unset.
> 
> If xhr’s [upload complete flag](https://xhr.spec.whatwg.org/#upload-complete-flag) is unset, then:
> 
> Set xhr’s [upload complete flag](https://xhr.spec.whatwg.org/#upload-complete-flag).
> 
> If xhr’s [upload listener flag](https://xhr.spec.whatwg.org/#upload-listener-flag) is set, then:
> 
> [Fire a progress event](https://xhr.spec.whatwg.org/#concept-event-fire-progress) named event at xhr’s [upload object](https://xhr.spec.whatwg.org/#upload-object) with 0 and 0.
> 
> [Fire a progress event](https://xhr.spec.whatwg.org/#concept-event-fire-progress) named [loadend](https://xhr.spec.whatwg.org/#event-xhr-loadend) at xhr’s [upload object](https://xhr.spec.whatwg.org/#upload-object) with 0 and 0.
> 
> [Fire a progress event](https://xhr.spec.whatwg.org/#concept-event-fire-progress) named event at xhr with 0 and 0.
> 
> **[Fire a progress event](https://xhr.spec.whatwg.org/#concept-event-fire-progress) named [loadend](https://xhr.spec.whatwg.org/#event-xhr-loadend) at xhr with 0 and 0**.

This means if we update internal properties on `loadend`, they won't yet be available in the preceding `readystatechange`:

```js
xhr.addEventListener('readystatechange', () => {
  if (xhr.readyStateChange === xhr.DONE) {
    console.log(xhr.status, xhr.statusText)
    // Will these be 0? 
  }
})
```

@blaenk, would you have a minute to see how regular XMLHttpRequest behaves in this regard? Are the updated response-related properties available in `readystatechange` on `DONE` immediately?

blaenk: I investigate some more and it seems like `onerror` for example is for network-level errors, so that definitely wouldn't set `status` since `status` assumes that there _was_ a request received. So in this case `load` being triggered on 'success' just means that a response was received, regardless of the http status code.

So it does make complete send to set those fields in `load` and I will make that change 👍🏼

blaenk:
@kettanaito no rush just in case you didn't notice the update here. Sorry for the back and forth!

blaenk:
@kettanaito happy new year! Is there anything I can help with to move this along?

To recap: you were right! I adapted the PR with the simpler approach that you mentioned.

Thanks again for your continued work on this project!

kettanaito (approved):
This looks fantastic! 

Happy New Year, @blaenk! Thanks for all the help with the libraries. Really happy to collaborate with you. Let's wait on the CI and if it passes, I'll merge this.

kettanaito:
If you would like to team up with me again, I could use some help with #324. This improves the browser operability of this library by moving away from the `StrictEventEmitter` and using `EventTarget` since it's present in both browser and Node.js. The change itself is simple and done but it causes a number of tests to fail seemingly for no reason. We could investigate that, if it sounds interesting to you.

blaenk:
Thank you so much, I look forward to the release when you get a chance!

I'll take a look at #324 when I get a chance.

kettanaito:
## Released: v0.19.3 🎉

This has been released in v0.19.3!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.19.3)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.19.3)

---

PR #336, opened 2023-02-13
fix(XHR Failed to construct Response): the status provided 0 is outside the range 200 599

### LINKED ISSUE #335: XHR: Failed to construct 'Response': the status provided (0) is outside the range [200, 599]
@kettanaito - We encountered the error shown below. Block the request URL in the browser network tab and call the API again to reproduce it. Any ideas on how to solve the following problem?

image

_Originally posted by @smo043 in https://github.com/mswjs/interceptors/issues/307#issuecomment-1428061952_

smo043:
- Fixes #335 

this fix address the below issue:

image


src/interceptors/XMLHttpRequest/utils/createResponse.ts
```diff
@@ -5,9 +5,13 @@ export function createResponse(
   request: XMLHttpRequest,
   responseBody: Uint8Array
 ): Response {
-  return new Response(responseBody, {
-    status: request.status,
-    statusText: request.statusText,
+  // this fix will help to return Error when status is 0
+  const {status} = request;
+  const body = status === 0 ? 'Error' : responseBody;
+  const req = status === 0 ? {status: 500, statusText: 'Internal Server Error'} : request;
```
smo043, line 11: I can also update the condition to use `!status` instead of `status === 0`

kettanaito: We should go from the intention here: the intention is to call `createRequest` with a _settled_ request. Settled requests have a non-0 status code. So, in the context of this function, receiving a 0 status code is a breach of the intention, and so this function must throw on those situations. 

You can use the `outvariant` package used in this library to throw on such conditions:

```js
import { invariant } from 'outvariant'

function createResponse(request, responseBody) {
  invariant(request.status !== 0, 'Failed to create a Fetch API Response out of XMLHttpRequest: received a non-settled request (status code 0)')
}
```

This will not solve the issue but it will guard against this function being called improperly.

kettanaito: To solve the issue, let's look at _why_ `createResponse` is called in here:

https://github.com/mswjs/interceptors/blob/d73624b222fd80a291486e13eb368d440bf21816/src/interceptors/XMLHttpRequest/XMLHttpRequestOverride.ts#L454

From the code, you can see that this only happens on a _bypassed request_. If you have a reproduction scenario (where you're taking the error screenshot), link the `@mswjs/interceptors` package to that application and set up some logging/`debugger`s to see _why_ that function is called while `this.status` is still `0` in XMLHttpRequestOverride's context. Perhaps it's a race condition? Maybe we are not awaiting things correctly. 

My suspicion would be that if `createResponse()` is called in the case of a _request error_ (which still fires the `loadend` listener), maybe we're not propagating the error's status onto XMLHttpRequestOverride instance and it remains 0? 🤔

smo043: @kettanaito - I'll give your suggested solution a shot. I used yalc to integrate this package into my project and came up with the solution described in this PR.

kettanaito: Let me know once you come to it. I can help you rebase against the latest main, there've been some changes concerning XHR interception.

smo043: Sure, I'll work this weekend

kettanaito:
I can help you with rebasing this so don't worry. Just ping me once you think I can look at this. Thanks!

kettanaito:
@smo043, there's been quite fundamental changes to the XHR interceptor but what you're proposing here still makes sense. I've switched to a Proxy-based interception and I see you've already rebased. Thanks.

kettanaito:
I can confirm that if _request fails_ then `xhr.status` will be rightfully 0. I think we still attempt to construct a Response in that scenario? That'd be incorrect. 

This is how we can emulate request failure:

```js
interceptor.on('request', () => {
  throw new Error('Network error')
})
```

This will halt the request (readyState 4) and its status will remain at 0, which is correct. That indicates that request never took off. 

I'm curious if we still attempt to create a Fetch Response in that case 🤔

smo043:
@kettanaito - Amazing I'll test it today and update you

smo043:
@kettanaito - I tried with the latest version. I blocked the url in the browser's network tab, and this time there is no response with a status code.

I have also noticed that I can't import presets or ClientRequestInterceptor as per documentation.

Error:
- package path ./ClientRequest is not exported from package @mswjs/interceptors (see exports field in @mswjs/interceptors/package.json
- browserInterceptors preset: Cannot find module @mswjs/interceptors/presets/browser

kettanaito:
## Released: v0.22.10 🎉

This has been released in v0.22.10!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.10)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.10)

---

Issue #350, opened 2023-02-27
SyntaxError: Named export 'debug' not found


thebuilder:
After the upgrade to ESM and browser bundle, I'm getting this error when importing the library into `.mjs` files:

```
file:///node_modules/.pnpm/@mswjs+interceptors@0.22.1/node_modules/@mswjs/interceptors/lib/node/chunk-37CATPNG.mjs:13
import { debug } from "debug";
         ^^^^^
SyntaxError: Named export 'debug' not found. The requested module 'debug' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'debug';
const { debug } = pkg;
```

The `debug` library is exported as `commonjs`.

kettanaito:
Hi, @thebuilder. Thanks for raising this. 

I believe we're using an older version of `debug` that's CJS-only. Can you please try updating `debug` in your fork of this repository and see if this resolves the issue? Perhaps it'd be a better solution to migrate to the newest `debug` instead of accounting for its CJS-only exports.

thebuilder:
I checked, and https://github.com/debug-js/debug/ hasn't been updated - You are using the latest version.

kettanaito:
Thanks for checking that! In that case, what you're proposing in the pull request is the way to go. Will have a closer look at it later today. Thank you for contributing.

kettanaito:
## Released: v0.22.5 🎉

This has been released in v0.22.5!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.5)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.5)

Make sure to always update to the latest version (`npm i @mswjs/interceptors@latest`) to get the newest features and bug fixes.

---

_Predictable release automation by [@ossjs/release](https://github.com/ossjs/release)_.

PR #351, opened 2023-02-27
fix: correctly imports commonjs "debug" package

thebuilder:
Fix for #350 

Changes the imports, so they handle the CommonJS exports from the "debug" package.

### DIFF
--- a/src/Interceptor.ts
+++ b/src/Interceptor.ts
@@ -1,10 +1,13 @@
-import { Debugger, debug } from 'debug'
+import type { Debugger } from 'debug'
+import pkg from 'debug'
 import { Listener } from 'strict-event-emitter'
 import { AsyncEventEmitter } from './utils/AsyncEventEmitter'
 import { nextTick } from './utils/nextTick'
 
 export type InterceptorEventMap = Record<string, any>
 export type InterceptorSubscription = () => void
+
+const debug = pkg
 
 export function getGlobalSymbol<V>(symbol: Symbol): V | undefined {
   return (
--- a/src/Interceptor.ts
+++ b/src/Interceptor.ts
@@ -1,13 +1,10 @@
-import type { Debugger } from 'debug'
-import pkg from 'debug'
 import { Listener } from 'strict-event-emitter'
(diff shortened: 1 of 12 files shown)


src/Interceptor.ts
```diff
@@ -1,11 +1,14 @@
-import { Debugger, debug } from 'debug'
+import type { Debugger } from 'debug'
+import pkg from 'debug'
```
kettanaito, line 2: Perhaps let's name it something else than a generic `pkg`? In fact, maybe we should depict the reason for this kind of import in the name itself:

```js
import debugCjsImport from 'debug'

const debug = debugCjsImport
```

What do you think about this?

kettanaito (commented):
The changes are solid, I've but one comment regarding the import name. Would like to know what you think.

src/Interceptor.ts
thebuilder: Sure - Makes sense.

thebuilder:
Looking through the code, I see there's a lot more references to `debug` in the code.

Think it might make more sense to make a simple "wrapper" around around it, that essentially re-exports it as ESM. Otherwise we'll end up doing this import dance in a lot of places.

kettanaito:
That's true, we use `debug` quite extensively. Having some util makes sense. I was thinking about something like

```js
// src/utils/debug.ts
import debugCjsImport from 'debug'

export const debug = debugCjsImport
```

What do you have in mind?

thebuilder:
Yeah, exactly like that 👍

kettanaito:
Let me know if this is ready for review 👀

thebuilder:
Should be! Replaced all the debug imports with the one from utils

kettanaito (approved):
Thank you, @thebuilder! Welcome to contributors 🎉

kettanaito:
## Released: v0.22.5 🎉

This has been released in v0.22.5!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.5)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.5)

---

PR #354, opened 2023-03-01
fix: improve node.js host resolution

### LINKED ISSUE #353: "getHostByRequestOptions" assumes "host" and "hostname" are the same thing
That's incorrect and given a _host_ (hostname + port), it then concludes that it's an IPv6 address and tries to wrap it in brackets:

```js
new URL('/resource', 'http://[127.0.0.1:1234]')
``` 

Which results in an invalid URL being constructed. 

The root cause for this is here: 

https://github.com/mswjs/interceptors/blob/beb1561770403e6175e24e7948da53418cf50ddc/src/utils/getUrlByRequestOptions.ts#L60-L62

_host_ must be removed from here.

kettanaito:
- Fixes #353 

This should also make the URL resolution a bit faster in the case when `options.uri` is explicitly provided because we won't attempt to figure out URL parts from the options then.

### DIFF
--- a/src/utils/getUrlByRequestOptions.test.ts
+++ b/src/utils/getUrlByRequestOptions.test.ts
@@ -116,3 +116,20 @@ it('supports "hostname" instead of "host" and "port"', () => {
     'https://127.0.0.1:1234/resource'
   )
 })
+
+it('handles IPv6 hostnames', () => {
+  expect(
+    getUrlByRequestOptions({
+      host: '::1',
+      path: '/resource',
+    }).href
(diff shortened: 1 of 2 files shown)


kettanaito:
## Released: v0.22.3 🎉

This has been released in v0.22.3!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.3)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.3)

---

PR #355, opened 2023-03-01
fix(XMLHttpRequest): resolve the "Illegal invocation" exception (conditional setter)

### LINKED ISSUE #339: XMLHttpRequest: "TypeError: Illegal invocation" when setting custom timeout
I was trying to use this with axios, and didn't get it to work. Upon further inspection, I figured the issue is with `XMLHttpRequest.timeout`.

The following case ilustrates the issue:

```js
const req = new XMLHttpRequest();
req.timeout = 10
req.open("GET", "http://www.example.org/example.txt");
req.send();
```

I get this debug log and error

```
common.js:113 xhr constructed new XMLHttpRequest +8m
createProxy.ts:46 Uncaught TypeError: Illegal invocation
    at Reflect.set (<anonymous>)
    at next (createProxy.ts:46:34)
    at XMLHttpRequest.setProperty (XMLHttpRequestController.ts:61:16)
    at Object.handler.set (createProxy.ts:47:26)
    at <anonymous>:2:13
```
Removing the `req.timeout = 10` line makes que request go all the way.

kettanaito:
- Fixes #339
- Fixes an issue where setting XHR callbacks like `ontimeout` did nothing because we've treated them as _method calls_ and not property setters. 

## Root cause for `Illegal invocation` on setters

Turns out, having a `handler.set` in the Proxy handlers added _conditionally_ is quite an issue. If the setter function is always added to the proxy, there are no illegal invocations anymore. 

## Root cause for `xhr.responseType` not being respected

JSDOM sets a bunch of property descriptors on both XMLHttpRequest implementation _and_ its prototype. We were not forwarding any of the prototype descriptors in our proxy, and that's why setting `xhr.responseType` didn't propagate to JSDOM and didn't affect the returned response. 

> This is fixed on the XHR controller level as this is something specific to XHR in JSDOM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
   "devDependencies": {
     "@commitlint/cli": "^16.0.2",
     "@commitlint/config-conventional": "^16.0.0",
-    "@open-draft/test-server": "^0.4.2",
+    "@open-draft/test-server": "^0.5.1",
     "@ossjs/release": "^0.4.0",
     "@playwright/test": "^1.31.1",
     "@types/cors": "^2.8.12",
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts
@@ -46,19 +46,6 @@ export class XMLHttpRequestController {
     this.responseBuffer = new Uint8Array()
 
     this.request = createProxy(initialRequest, {
-      setProperty: ([propertyName, nextValue], invoke) => {
-        /**
-         * @note Setting the "withCredentials" property on the XHR proxy
-         * causes the "TypeError: Illegal invocation" error. Instead,
-         * define the property on the original XHR instance itself.
-         */
-        if (propertyName === 'withCredentials') {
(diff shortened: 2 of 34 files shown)


package.json
```diff
@@ -85,7 +85,7 @@
   "devDependencies": {
     "@commitlint/cli": "^16.0.2",
     "@commitlint/config-conventional": "^16.0.0",
-    "@open-draft/test-server": "^0.4.2",
+    "@open-draft/test-server": "^0.5.1",
```
kettanaito, line 88: Updating this so that custom handlers are applied _before_ the built-in CORS middleware of `test-server`. This way we can override CORS options, like `Access-Control-Allow-Origin` to more specific values (required when using `withCredentials = true` in XHR).

kettanaito:
## Released: v0.22.4 🎉

This has been released in v0.22.4!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.4)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.4)

---

PR #356, opened 2023-03-02
fix(fetch): treat middleware exceptions as TypeError

kettanaito:
- Fixes https://github.com/mswjs/msw/issues/1557

Since we now ship the fetch interceptor in the Node preset by default, we are mocking the `globalThis.fetch` function calls directly instead of tapping into the underlying `http.ClientRequest` as before. Previously, erroring the underlying client request bubbled to whichever fetch implementation the user was using, and that implementation faithfully handled those errors as `TypeError`.

Now, we become that fetch implementation (in the case of middleware exceptions) so we need to forward middleware exceptions as `TypeError` to the consumer. 

- [ ] Backport this to 0.17.x so the fix is there.


kettanaito:
## Released: v0.22.4 🎉

This has been released in v0.22.4!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.4)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.4)

---

PR #364, opened 2023-03-16
fix(ClientRequest): set null body for GET/HEAD requests

gfloyd:
Creating a GET or HEAD fetch Request with a non-null/undefined body throws a `TypeError`. Some (possibly misbehaving) clients (notably the Stripe Node library) write an empty string to the request buffer even for GET/HEAD, causing a failure when intercepting any of those requests.

A test has been added to `createRequest.test.ts` that fails without this change.


kettanaito (approved):
Thanks for fixing this, @gfloyd!

PR #369, opened 2023-04-05
fix(fetch): respect AbortSignal during response patching

### LINKED ISSUE #368: AbortSignal is ignored when using response patching
When aborting a request with an `AbortController`, fetch rejects with an `AbortError`.
This does not happen when we are using response patching at the moment.

```ts

interceptor.apply()

interceptor.on('request', async (request) => {
  if (request.url.endsWith("/blocked")) {
    request.respondWith(new Response("Blocked", { status: 403 }))
  }
})

// Rejects with an AbortError as expected
const controller1 = new AbortController();
fetch("/trusted", {signal: controller1.signal })
  .then(console.log, console.error);
controller1.abort();

// Fulfills with the patched response
const controller2 = new AbortController();
fetch("/blocked", { signal: controller2.signal })
  .then(console.log, console.error);
controller2.abort();
```

To keep things consistent with the fetch API, I would suggest to throw the same error when we are patching the response.

hpohlmeyer:
Fixes #368

---

Makes aborting fetch requests consistent between patched and unpatched responses.

### DIFF
--- a/src/interceptors/fetch/index.ts
+++ b/src/interceptors/fetch/index.ts
@@ -68,16 +68,8 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
 
       const mockedResponse = resolverResult.data
 
-      if (mockedResponse) {
+      if (mockedResponse && !request.signal?.aborted) {
         this.log('received mocked response:', mockedResponse)
-        
-        if (request.signal.aborted) {
-          this.log('request aborted, ignoring mocked response');
-
-          // Throws an "AbortError" DOMException
-          await pureFetch(request)
-        }
-
         const responseCloine = mockedResponse.clone()
 
         this.emitter.emit(
--- a/src/interceptors/fetch/index.ts
+++ b/src/interceptors/fetch/index.ts
@@ -70,6 +70,14 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
 
       if (mockedResponse) {
         this.log('received mocked response:', mockedResponse)
+        
+        if (request.signal.aborted) {
+          this.log('request aborted, ignoring mocked response');
+
+          // Throws an "AbortError" DOMException
+          await pureFetch(request)
+        }
+
         const responseCloine = mockedResponse.clone()
 
         this.emitter.emit(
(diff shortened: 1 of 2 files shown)


src/interceptors/fetch/index.ts
```diff
@@ -70,6 +70,14 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
 
       if (mockedResponse) {
         this.log('received mocked response:', mockedResponse)
+        
+        if (request.signal.aborted) {
+          this.log('request aborted, ignoring mocked response');
+
+          // Throws an "AbortError" DOMException
+          await pureFetch(request)
```
hpohlmeyer, line 78: We could throw a `DOMException` manually here, but I wanted the error message to have the same wording as the original. Calling `fetch` here throws that error.

kettanaito: I think it's better to rely on whichever `fetch` the consumer is using here. For example, I doubt that global fetch in Node will throw a `DOMException` 😉 So we remain consistent in this regard. 

My only concern with this is that we're creating a special clause for the aborted requests within the mocked response handling. I think it's confusing. If I understood it right, we will know about `request.signal.aborted` by the time we receive a mocked response (or don't receive any). With that in mind, would this produce the same result:

```js
if (mockedResponse && !request.signal?.aborted) {
  // ...this closure has an early return
}

// This one handled all the other use cases,
// which includes the lack of mocked response or
// aborted request regardless of the mocked response state.
pureFetch(...) // the same as now.
```

Basically, instead of treating aborted requests as a special case, we instead instruct the interceptor to only handle mocked responses if the request hasn't been aborted. This makes sense semantically because when the client aborts a request, it mustn't expect the mocked response to be sent. 

What do you think about this?

hpohlmeyer: Yes, this is way better. 

> Also, isn't request.signal potentially undefined?

According to the typings it is not. I have tested Safari, Chrome, Firefox and Deno and you always get an `AbortSignal`. The spec says ["A Request object has an associated signal (null or an AbortSignal object), initially null."](https://fetch.spec.whatwg.org/#request-signal) though, so I think assuming it might be null might be the safest bet.

kettanaito (approved):
Love it! Thank you for fixing this, @hpohlmeyer. Welcome to contributors 🎉

src/interceptors/fetch/index.ts
kettanaito: Thanks for double-checking this.

kettanaito:
## Released: v0.22.11 🎉

This has been released in v0.22.11!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.11)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.11)

---

PR #373, opened 2023-05-03
fix(ClientRequest): stream response chunks as they come in

kettanaito:
- Reported in https://github.com/mswjs/msw/discussions/1464#discussioncomment-5789415

## Changes

Sets the `res` on `ClientRequest` instance _immediately_ (previously, only when the response stream has finished reading). 

Instead of pushing to the `IncomingMessage`, emit its `data` events for each read chunk of the response stream. This is necessary because with a simple `this.response.push(value)`, the end readable stream only contained _the first chunk_. For some reason, any other chunks were ignored with the previous implementation (they never emitted the `data` event on `IncomingMessage`). 

> This is potentially dangerous but let's see what the tests tell us.

### DIFF
--- a/src/interceptors/ClientRequest/NodeClientRequest.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.ts
@@ -401,6 +401,7 @@ export class NodeClientRequest extends ClientRequest {
       // and shouldn't be written to anymore.
       this.response.push(null)
       this.response.complete = true
+      this.response.emit('end')
 
       isResponseStreamRead.resolve()
       this.logger.info('closed response stream!')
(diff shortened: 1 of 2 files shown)


kettanaito:
It may be a good idea to add an events compatibility test to `http` and `https` to ensure we aren't emitting more response events than expected (speaking of `data` and `end` response events).

kettanaito:
## Released: v0.22.13 🎉

This has been released in v0.22.13!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.13)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.13)

---

Issue #374, opened 2023-05-06
Support mocking SMTP


kentcdodds:
I'm circling back on https://github.com/mswjs/msw/discussions/751

I would like to make my `sendEmail` function easier to migrate between different email providers and using the SMTP standard is the easiest way to do this. But then I have to resort to doing weird things to mock it (nodemailer specifically) out during development and testing.

Any chance MSW could support mocking these kinds of requests?

kettanaito:
Hey, @kentcdodds. 

I took a look at how Nodemailer works, and as I suspected, they tap into a raw socket connection via the `net` module of Node.js. The Interceptors currently operate on a higher level `http.ClientRequest` class to provision HTTP interception since `net` is generally too low-level to gain much control over the request instances. At that level, Node is operating with socket messages that don't have any context of the sender (was it an HTTP request? Was it an SMTP message?) or don't have enough context to lower our implementation logic to the sockets.

Here's a rough representation of those network layers:

```
http
  http.ClientRequest
    IncomingMessage / OutgoingMessage
      Socket
        write / on('data')
```

The Socket layer seems to be the highest layer we can capture requests over SMTP and other non-HTTP protocols in Node.js. That's also [how Nodemailer](https://github.com/nodemailer/nodemailer/blob/e3cc93a9c20939b209c804857c75aea0d3305913/lib/smtp-connection/index.js#L354) and other implementations do that since Node doesn't ship with a standard API to make SMTP requests easier as it does for HTTP requests.

My approach to interception is that it should be unified, as in I'm not eager to ship logic that's specific to certain libraries or even protocols. A good solution here would be to lower _all_ the interception logic to the Socket layer, including the interception of HTTP requests. But, as I've mentioned, that can prove problematic in some use cases. 

An incremental solution would be to try implementing something like a `SocketInterceptor` that is limited by design and may not support full HTTP interception like the `ClientRequestInterceptor` does. In turn, it could be utilized to implement SMTP interception since it _is_ implemented on that layer and the consumer won't lose any context when tapping into it (it is dealing with raw sent/received messages). 

I'm currently busy with the Fetch API primitives adoption in MSW and I'm excited it's reaching the finish line. I won't have any time to look into this feature in the nearest future, I'm afraid. This can be a great improvement for the upcoming releases though!

kentcdodds:
Thanks for explaining all of that! If this becomes a high enough priority for me then I'll signal in here that I've started looking into it myself. Cheers!

[...]

PR #375, opened 2023-05-07
feat: add net.Socket-based interceptor

kettanaito:
> Before continuing, read this https://github.com/mswjs/interceptors/issues/399

- Related to #374 

A proof of concept of a Socket-based interceptor.

## Things I learned

- `net.createConnection()` accepts a `lookup` function that we can use to resolve non-existing hosts to whatever we want. It won't automatically give us a successful connection (because that's done by `socket._handler.connect()`) but it will prevent DNS lookup issues.
- `net.creaetConnection()` accepts a custom `handler` class instance. It's an internal API responsible for provisioning the connection to a resolved address, reading/writing socket chunks, and terminating the socket. 

## Links

- https://github.com/nodejs/node/blob/main/lib/net.js#L1169
- https://github.com/nodejs/node/blob/main/lib/net.js#L1524

## Roadmap

- [ ] Socket interceptor is sensitive to import order. If you import `./socket.impl` _after_ you import `http`, it stops working. 
- [ ] Should establish the actual socket connection instead of force-emitting `connect` and `ready` events. This will improve the underlying code's integrity and allow us to replay the connection errors (very similar to what we already do in `ClientRequestInterceptor`).
- [ ] Support SMTP and other protocols (currently getting `Unsupported fd type: UNKNOWN` from `createHandle`).
- [ ] Wrap this whole thing in the `Interceptor` (`class SocketInterceptor extends Interceptor`).

### DIFF
--- a/test/modules/http/socket.impl.ts
+++ b/test/modules/http/socket.impl.ts
@@ -1,0 +1,228 @@
+import net from 'net'
+import { Interceptor } from '../../../src'
+
+// net.createConnection()
+// Socket.connect()
+// lookup
+// emitLookup
+// socket._handle.connect()
+// -> afterComplete
+// socket.push(null), socket.read(), socket.emit('connect'), socket.emit('ready')
+
+export interface SocketEventMap {
+  connection: () => void
+}
+
+export class SocketInterceptor extends Interceptor<SocketEventMap> {
+  static interceptorSymbol = Symbol('SocketInterceptor')
+
+  constructor() {
+    super(SocketInterceptor.interceptorSymbol)
+  }
+
+  protected setup(): void {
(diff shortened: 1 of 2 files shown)


kettanaito:
Since this implementation is specific to the import order, I couldn't create the interceptor out of it. The interceptor is applied by calling `interceptor.apply()` in the `beforeAll()` test hook. At that point, the test has already imported `http`, which renders the current interceptor obsolete (again, import order sensitivity).

kettanaito:
Closing in favor of https://github.com/mswjs/interceptors/pull/515.

kettanaito:
## Released: v0.32.0 🎉

---

PR #376, opened 2023-05-08
fix(Interceptor): set interceptor state on the current event loop cycle

kettanaito:
- `APPLIED` and `DISPOSED` interceptor state is now transitioned to _immediately_ (in the current event loop). 

I find it confusing that the interceptor is still showing `APPLYING` state when it's already applied. The patching logic affects the runtime as soon as it's invoked, so there is little reason to defer the interceptor's state transition to the next tick. If anything, it only results in confusing logs. 

This isn't a breaking change because nobody should rely on the internal ready state of the interceptors.

### DIFF
--- a/src/Interceptor.test.ts
+++ b/src/Interceptor.test.ts
@@ -52,21 +52,15 @@ describe('readyState', () => {
     interceptor.apply()
 
     expect(interceptor.readyState).toBe(InterceptorReadyState.INACTIVE)
-
-    await nextTickAsync(() => {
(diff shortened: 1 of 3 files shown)


kettanaito:
## Released: v0.22.14 🎉

This has been released in v0.22.14!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.14)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.14)

---

PR #382, opened 2023-06-13
fix: upgrade to strict-event-emitter@0.4.3

kettanaito:
- Fixes https://github.com/mswjs/msw/issues/1539

Upgrading `strict-event-emitter` to the same version range as in the latest MSW (1.x). Since MSW 2.x, the `strict-event-emmiter` is on 0.5.

### DIFF
--- a/src/utils/AsyncEventEmitter.ts
+++ b/src/utils/AsyncEventEmitter.ts
@@ -1,6 +1,13 @@
 import { Debugger, debug } from 'debug'
-import { StrictEventEmitter, EventMapType } from 'strict-event-emitter'
+import { Emitter } from 'strict-event-emitter'
 import { nextTick } from './nextTick'
+
+export type FunctionEventMap = Record<string, (...args: Array<any>) => any>
+export type FunctionEventMapToStrictEventEmitterMap<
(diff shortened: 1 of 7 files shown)


src/utils/AsyncEventEmitter.ts
```diff
@@ -1,7 +1,14 @@
 import { Debugger, debug } from 'debug'
-import { StrictEventEmitter, EventMapType } from 'strict-event-emitter'
+import { Emitter } from 'strict-event-emitter'
 import { nextTick } from './nextTick'
 
+export type FunctionEventMap = Record<string, (...args: Array<any>) => any>
+export type FunctionEventMapToStrictEventEmitterMap<
+  Functions extends FunctionEventMap
+> = {
+  [K in keyof Functions]: Parameters<Functions[K]>
```
kettanaito, line 9: `EventMapType` and `EventMap` types are different across strict-event-emitter 0.2 and 0.5. Adding this transformer type to account for that difference since the events contract in Interceptors uses `Record<string, Function>` index signature to allow specifying the expected return type of the events (since those can be awaited, etc).

PR #387, opened 2023-06-28
fix(xhr): fix onloadend not being called

hpohlmeyer:
The `onloadend` event handler is not being called if we do not send mocked response from a request interceptor:

```ts
import { XMLHttpRequestInterceptor } from "@mswjs/interceptors/XMLHttpRequest";
import { BatchInterceptor } from "@mswjs/interceptors";

  name: "my-interceptor",
  interceptors: [new XMLHttpRequestInterceptor()],
});

interceptor.on("request", (req) => {
  console.log("new request", req);
})

interceptor.apply();

function handleResponse(event: any) {
  console.log(`[Response] Type: ${event.type}`);
  console.log(`[Response] Content: ${xhr.responseText}`);
}

const xhr = new XMLHttpRequest();
xhr.open("GET", <some-url>);

// This works as expected
xhr.addEventListener("loadend", handleResponse);

// This will never be called
xhr.onloadend = handleResponse;

xhr.send(null);
```

This is caused by `onloadend` not being located on `XMLHttpRequest`, but on `XMLHttpRequestEventTarget` which is part of its prototype chain (`XMLHttpRequest` <- `XMLHttpRequestPrototype` <- `XMLHttpRequestEventTargetPrototype`).

Because of that the property descriptor for `onloadend` is not found and we patch it on the `XMLHttpRequest`. When the browser calls the setter for `onloadend`, it does that on `XMLHttpRequestEventTarget`, which is why our proxy does not get notified.

This PR fixes the issue by walking the prototype chain, until it finds the property it is looking for. Redefining the property or looking for a property descriptor is done on the same level where the property is defined.

The vitest tests in `xhr-event-handlers.test.ts` do not complain about the issue, I am sure in which environment they are running, but it is an issue in browsers. I have added a playwright test because of that.

---

As a sidenote: While looking for a fix, I found https://github.com/mswjs/interceptors/blob/v0.22.15/src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts#L51-L64, which looks like it could be caused by a similar issue. I have removed the highlighted part and ran the tests. They run without errors, so if there is a test covering that part it might be fixed.

### DIFF
--- a/src/utils/findPropertySource.test.ts
+++ b/src/utils/findPropertySource.test.ts
@@ -1,0 +1,29 @@
+import { describe, it, expect } from 'vitest'
+import { findPropertySource } from './findPropertySource'
+
+describe('findPropertySource', () => {
+  it('does return the source for objects without prototypes', () => {
+    const obj = Object.create(null)
+    obj.test = undefined
+    const source = findPropertySource(obj, 'test')
+    expect(source).toBe(obj)
+  })
+
+  it('does return the source for objects with prototypes', () => {
+    const prototype = Object.create(null)
+    prototype.test = undefined
+
+    const obj = Object.create(prototype)
+
+    const source = findPropertySource(obj, 'test')
+    expect(source).toBe(prototype)
+  })
+
+  it('does return null if the prototype chain does not contain the property', () => {
+    const prototype = Object.create(null)
+    const obj = Object.create(prototype)
+
+    const source = findPropertySource(obj, 'test')
+    expect(source).toBeNull()
+  })
+})
--- a/src/utils/findPropertySource.ts
+++ b/src/utils/findPropertySource.ts
@@ -1,0 +1,12 @@
+export function findPropertySource(
+  target: object,
+  propertyName: string | symbol
+): object | null {
+  if (propertyName in target === false) return null
+
+  const hasProperty = Object.prototype.hasOwnProperty.call(target, propertyName)
+  if (hasProperty) return target
+
+  const prototype = Reflect.getPrototypeOf(target)
+  return prototype ? findPropertySource(prototype, propertyName) : null
+}
--- a/test/modules/XMLHttpRequest/compliance/xhr-event-handlers.browser.test.ts
+++ b/test/modules/XMLHttpRequest/compliance/xhr-event-handlers.browser.test.ts
@@ -17,7 +17,7 @@ test.afterAll(async () => {
   await httpServer.close()
 })
 
-test.only('onloadend handler is called', async ({ page, loadExample }) => {
+test('onloadend handler is called', async ({ page, loadExample }) => {
   await loadExample(require.resolve('./xhr-event-handlers.browser.runtime.js'))
 
   const { request, calls } = await page.evaluate(async (url) => {
--- a/test/modules/XMLHttpRequest/compliance/xhr-event-handlers.browser.test.ts
+++ b/test/modules/XMLHttpRequest/compliance/xhr-event-handlers.browser.test.ts
@@ -1,0 +1,55 @@
+import { HttpServer } from '@open-draft/test-server/http'
+import { test, expect } from '../../../playwright.extend'
+import { useCors } from '../../../helpers'
+
+const httpServer = new HttpServer((app) => {
+  app.use(useCors)
+  app.get('/resource', (req, res) => {
+    res.status(200).send('hello')
+  })
+})
+
+test.beforeAll(async () => {
+  await httpServer.listen()
+})
+
+test.afterAll(async () => {
+  await httpServer.close()
+})
+
+test.only('onloadend handler is called', async ({ page, loadExample }) => {
+  await loadExample(require.resolve('./xhr-event-handlers.browser.runtime.js'))
+
+  const { request, calls } = await page.evaluate(async (url) => {
+    const calls = {
+      loadEndHandler: 0,
+      loadEndListener: 0,
+    }
+
+    const xhr = new XMLHttpRequest()
+    xhr.open('GET', url)
+    xhr.onloadend = () => calls.loadEndHandler++
+    xhr.addEventListener('loadend', () => calls.loadEndListener++)
+
+    await Promise.all([
+      new Promise((resolve) => {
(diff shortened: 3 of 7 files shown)


hpohlmeyer:
Closing this, because I did some more digging. I’ll open an issue instead to discuss how a proper fix could look like.

hpohlmeyer:
Reopening this with an updated description and fix, because I found the root cause of the issue and a proper fix for it.

kettanaito:
Thanks for looking into this, @hpohlmeyer! 

To double-check, have you encountered this issue while working with the latest version of this library? Since it's not the version most people install when installing `msw` today, I wanted to know whether we need to fix this as a backport to 0.17.x.

test/modules/XMLHttpRequest/compliance/xhr-event-handlers.browser.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    xhr.onloadend = () => calls.loadEndHandler++
+    xhr.addEventListener('loadend', () => calls.loadEndListener++)
+
+    await Promise.all([
+      new Promise((resolve) => {
+        const resolveDelayed = () => setTimeout(resolve, 1000)
+        xhr.addEventListener('error', resolveDelayed)
+        xhr.addEventListener('load', resolveDelayed)
+      }),
+      xhr.send(null),
```
kettanaito, line 40: Since `xhr.send` doesn't return a Promise, is there a reason to put these two in a single `Promise.all`?

test/modules/XMLHttpRequest/compliance/xhr-event-handlers.browser.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+test.beforeAll(async () => {
+  await httpServer.listen()
+})
+
+test.afterAll(async () => {
+  await httpServer.close()
+})
+
+test('onloadend handler is called', async ({ page, loadExample }) => {
```
kettanaito, line 20: Maybe let's specify it a bit more explicitly in the test name that this concerns the scenario when _not_ returning a mocked response?

src/utils/findPropertySource.test.ts
```diff
@@ -0,0 +1,29 @@
+import { describe, it, expect } from 'vitest'
+import { findPropertySource } from './findPropertySource'
+
+describe('findPropertySource', () => {
```
kettanaito, line 4: I think we generally avoid describe blocks so you can drop it and use individual `it` blocks as test cases for the `findPropertySource` function.

src/utils/findPropertySource.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  it('does return the source for objects without prototypes', () => {
+    const obj = Object.create(null)
+    obj.test = undefined
+    const source = findPropertySource(obj, 'test')
+    expect(source).toBe(obj)
+  })
+
+  it('does return the source for objects with prototypes', () => {
```
kettanaito, line 12: Please drop the `does` and use `returns` instead ;) It's a nitpick, I know, but I prefer using third-person "returns"/"prevents"/etc for positive test cases and "does not X" for negative. Consistency matters.

src/utils/findPropertySource.ts
```diff
@@ -0,0 +1,12 @@
+export function findPropertySource(
+  target: object,
+  propertyName: string | symbol
+): object | null {
+  if (propertyName in target === false) return null
```
kettanaito, line 5:
```suggestion
  if (!(propertyName in target)) return null
```

src/utils/findPropertySource.test.ts
hpohlmeyer: I was looking at `getCleanUrl` for reference, which seems to be the only one using a `describe` block. 😅

test/modules/XMLHttpRequest/compliance/xhr-event-handlers.browser.test.ts
hpohlmeyer: Good point. Do you think that’s good enough?

test/modules/XMLHttpRequest/compliance/xhr-event-handlers.browser.test.ts
hpohlmeyer: I wanted to avoid adding the event listeners after the `send` call.

```ts
xhr.send(null)

await new Promise((resolve) => {
  const resolveDelayed = () => setTimeout(resolve, 1000)
  xhr.addEventListener('error', resolveDelayed)
  xhr.addEventListener('load', resolveDelayed)
})
```

seems to work fine though. Do think that’s fine or would you rather see a more specific version like:

```ts
xhr.send(null)

const finishedPromise = new Promise((resolve) => {
  const resolveDelayed = () => setTimeout(resolve, 1000)
  xhr.addEventListener('error', resolveDelayed)
  xhr.addEventListener('load', resolveDelayed)
})

xhr.send(null)
await finishedPromise
```

hpohlmeyer:
> To double-check, have you encountered this issue while working with the latest version of this library? Since it's not the version most people install when installing msw today, I wanted to know whether we need to fix this as a backport to 0.17.x.

Yes, I am using `@mswjs/interceptors` on the latest version directly, not through `msw`.

src/utils/findPropertySource.ts
```diff
@@ -0,0 +1,12 @@
+export function findPropertySource(
```
kettanaito, line 1: Could you please add a short description of the intention of this function? So it's clear what it's intended to do as time goes on. Thanks!

kettanaito (approved):
This is a fantastic change! Thank you, @hpohlmeyer. Looks ready to me!

src/utils/findPropertySource.test.ts
kettanaito: My bad then. I'm still normalizing the tests across the ecosystem. Thanks for adjusting this!

test/modules/XMLHttpRequest/compliance/xhr-event-handlers.browser.test.ts
kettanaito: I like it. Thanks!

PR #393, opened 2023-07-25
Abort all requests the interceptor is disposed

JesusTheHun:
This PR aims to solve issue https://github.com/mswjs/msw/issues/778.

I took the following approach : 

- Decorate the global AbortController class
- Hold weak references to all controllers created
- Once a request is intercepted, hold a strong reference to its controller
- Once a request ends, free the strong reference
- On dispose, abort the requests using their controllers 

Since this only affects Node.js, I implemented the logic into `FetchInterceptor` and `ClientRequestInterceptor`

### DIFF
--- a/src/interceptors/ClientRequest/NodeClientRequest.test.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.test.ts
@@ -27,7 +27,6 @@ const httpServer = new HttpServer((app) => {
 })
 
 const logger = new Logger('test')
-const registerSignal = () => {}
 
 beforeAll(async () => {
   await httpServer.listen()
@@ -46,7 +45,6 @@ it('gracefully finishes the request when it has a mocked response', async () => 
     {
       emitter,
       logger,
-      registerSignal
     }
   )
 
@@ -96,7 +94,6 @@ it('responds with a mocked response when requesting an existing hostname', async
     {
       emitter,
       logger,
-      registerSignal
     }
   )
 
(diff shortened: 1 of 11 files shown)


kettanaito:
Let's have a discussion on this one, I think I know how we can approach it in an elegant manner. Will gather my thoughts and post here once I have a minute.

JesusTheHun:
@kettanaito do you think you can find some time in the coming weeks to handle this PR ? I think I can allocate an afternoon next week if you want to discuss this.

howagain:
Wanted to voice my support for this. As being able to test abort controllers is an important part of unit testing. Hoping this PR solves https://github.com/mswjs/msw/discussions/1646

jd-carroll:
Copying my 2cents here as well...

I think this change really should be in the `RequestHandler` not the interceptors.

Check out mswjs/msw#778 for a more full discussion.

JesusTheHun:
> Wanted to voice my support for this. As being able to test abort controllers is an important part of unit testing. Hoping this PR solves https://github.com/mswjs/msw/discussions/1646 

@howagain MSW 2.0 includes my first PR that fixes AbortSignal. So if you manually abort the request, you should be able to test that.
If not, this PR won't be enough because what this PR does is to abort all requests during MSW shutdown. It gives you no control while the tests are running.

This could be a nice improvement to build on top of this PR. Although, given that this PR has not been merged after 6 months, I will not write anything else before it is merged. 

Maintainers have a limited bandwidth. I understand it can be frustrating but we are spending hundreds of hours of our free time for the benefits of strangers on the internet.

Be kind ❤️

PR #394, opened 2023-07-25
fix(fetch): respect "abort" event on the request signal

JesusTheHun:
This [discussion](https://github.com/mswjs/msw/discussions/1646) pointed out that fetch request are not aborted if the user abort them using an `AbortController`.
This PR fixes it.


src/interceptors/fetch/index.ts
```diff
@@ -46,13 +47,21 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
 
       this.logger.info('awaiting for the mocked response...')
 
+      const signal = interactiveRequest.signal
+      const rejectWhenRequestAborted = new DeferredPromise<string>()
+
+      signal.addEventListener('abort', () => rejectWhenRequestAborted.reject(signal.reason))
```
kettanaito, line 53: I think we should implement it the way that user-caused aborts are treated as `AbortError`, as per [spec](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort). 

If I'm reading this right, if the user aborts the request, we don't have to do anything about that except stop the request listener promise. If that's the case, then we should instead have a race between two promises that _resolve_. 

But this may be more complicated than that. I'm trying to account for user-caused abort events while working on #398 and I could certainly use your opinion on what's the best way to approach this.

kettanaito: We can also use a built-in rejection mechanism `.throwIfAborted` on the signal ([spec](https://dom.spec.whatwg.org/#aborting-ongoing-activities)). 

Perhaps it'd be more semantically correct to do that and handle the thrown `AbortError` as a special case below.

```js
const resolverResult = await until(async () => {
  request.signal.throwIfAborted()

  // the rest of the request listener promises.
})

if (resolverResult.error) {
  if (resolverResult.error instanceof AbortError) {
    // We know request listener lookup halted due to
    // the operation being aborted.
    handleIfNeeded()
    return
  }

  // Handle other exceptions from the request listeners.
}
```

JesusTheHun: I will address your points in the near future but I wanted to let you know that I already have submitted a PR for #398

JesusTheHun: @kettanaito Outside of `msw` realm, when a request is aborted, the request/promise is rejected with a `DOMException` that have `reason` set to `AbortError`. Is there any reason to change this behavior ?

JesusTheHun: I have updated the PR to use the native error instead of a wrapped error.

src/interceptors/fetch/index.test.ts
```diff
@@ -0,0 +1,96 @@
+import { DeferredPromise } from '@open-draft/deferred-promise'
```
kettanaito, line 1: I think we should rename this test file to `abort-controller.test.ts` and move it under `test/modules/fetch/compliance` where we store all integration tests.

src/interceptors/fetch/index.ts
```diff
@@ -46,13 +47,21 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
 
       this.logger.info('awaiting for the mocked response...')
 
+      const signal = interactiveRequest.signal
```
kettanaito, line 50: Will the original request's abort controller propagate to the interactive request instance too? 

I'm wondering if we should account for `request.signal` directly in case the user aborts it instead of aborting a request in the `request` listener (where they have access to the interactive request). 

I will take a closer look at the tests, I think you've answered this question in there.

src/interceptors/fetch/index.ts
```diff
@@ -62,10 +71,7 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
       })
 
       if (resolverResult.error) {
-        const error = Object.assign(new TypeError('Failed to fetch'), {
-          cause: resolverResult.error,
-        })
-        return Promise.reject(error)
+        return Promise.reject(resolverResult.error)
```
kettanaito, line 74: I believe this is incorrect. I copied the previous behavior from Undici and we must comply by it. Note that the FetchInterceptor is primarily meant for Node, and I trust Undici implement the spec rather faithfully.

kettanaito: I'd suggest we revert this particular change for now because it's not related to the abort controller support. We can discuss it as a separate improvement point, what do you think?

src/interceptors/fetch/index.ts
JesusTheHun: @kettanaito it turns out they do : https://github.com/nodejs/undici/blob/main/test/fetch/abort.js#L78

src/interceptors/fetch/index.test.ts
JesusTheHun: same goes for the test added inside `src/interceptors/ClientRequest/index.test.ts` I assume ?

src/interceptors/fetch/index.ts
kettanaito: Okay, got it now. There's no way to abort the request from the request listener because you have to do that via the abort controller you set on the original request. 

I wonder if this isn't a design flaw on our part since allowing the user to abort a request in the request listener may be a good feature to have. Regardless, irrelevant in the context of this fix.

src/interceptors/fetch/index.test.ts
kettanaito: We can leave that one be, it doesn't concern itself with request handling but focuses on how the `.respondWith()` works in the context of the ClientRequest. I think it's fine.

kettanaito:
I've updated the tests a little, but what you wrote was really good. The abort controller tests now pass but I think we need to take a look at the `test/modules/fetch/fetch-exception.test.ts` to see if I got the original intention there wrong. 

During my tests with Undici, network errors would result in a `TypeError: Failed to fetch` where the original thrown exception is set in the `cause` by Undici (I don't believe that happens in the browser). In other aspects, a network error also results in `TypeError` in the browser. 

Here's the reference: https://github.com/nodejs/undici/blob/111fd2388094270259814ca7ea5da7d07b2126f6/lib/fetch/index.js#L227

Note that Unidici treats response (?) abort and network errors differently:

https://github.com/nodejs/undici/blob/111fd2388094270259814ca7ea5da7d07b2126f6/lib/fetch/index.js#L211-L221

Maybe we should do the same in the `resolverResult.error` handling.

src/interceptors/fetch/index.ts
JesusTheHun: Yes because on [line 63](https://github.com/mswjs/interceptors/pull/394/files/e0ca09dc464e815ef00f9ccc292cd523774c1734#diff-4593f50c24857397e849d8d6618757914a847fa2642a22bfaa3b703e15cb4969R63) we `Promise.race` the resolution and the rejection.

src/interceptors/fetch/index.ts
```diff
@@ -46,13 +47,21 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
 
       this.logger.info('awaiting for the mocked response...')
 
+      const signal = interactiveRequest.signal
+      const requestAbortRejection = new DeferredPromise<string>()
+
+      signal.addEventListener('abort', () => requestAbortRejection.reject())
```
kettanaito, line 53: Should we forward the abort information exposed to this listener, like abort reason and such? It does work without it so I assume the abort controller already exposes that information in the abort error, is that correct? 

As in, why this is not needed?

```js
signal.addEventListener('abort', () => {
  requestAbortRejection.reject(signal.reason)
})
```

🤔

src/interceptors/fetch/index.ts
JesusTheHun: I checked the Undici implementation and there are only two causes of error : abortion and network issue.
Since the abortion error is well defined (specific class and specific error code), we can check the error type and reject accordingly. 
If we don't do that, we deviate from production behavior.

src/interceptors/fetch/index.test.ts
JesusTheHun: I mean I've added a test abort abortion in this file

src/interceptors/fetch/index.ts
kettanaito: I agree, we should do it the same way: handle the two error scenarios separately:

- Keep what you've introduced for abort errors.
- Revert what was there previously (the `TypeError`) to handle all the other errors (effectively, network errors).

kettanaito: Thanks so much for addressing it so quickly! Will give it the last round of review and let's get this published.

src/interceptors/fetch/index.ts
kettanaito: I understand how it affects the promise flow, us using the abort reason as the rejection value and that it may produce false negatives, but my question was about whether we have to respect that reason in any way? As in, forward it directly. We can always reject with an object, or a different data structure to prevent false positives.

src/interceptors/fetch/index.ts
```diff
@@ -70,10 +70,17 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
         return mockedResponse
       })
 
-      if (resolverResult.error || requestAbortRejection.state === 'rejected') {
+      if (requestAbortRejection.state === 'rejected') {
```
kettanaito, line 73: Do you think we should conceptually treat this promise as resolved instead? 

If we imagine promise as an intention, then the intention here is to know when the request has been aborted by another actor. In that light, an aborted request would mean a resolved `requestAborted` promise. 

It would make this check read much nicer as well. What are your thoughts on this? 

I'm also a bit hesitant to base this condition on the request abort promise itself. I'd rather we used that promise in the `Promise.race()` but handled the `resolverResult.error` differently. This implies that `requestAbortRejection` can still throw but it does so with a custom error/object that is later checked on the `resolverResult.error`:

```js
signal.addEventListener('abort', () => {
  requestAbortRejection.reject({ type: 'aborted', reason: signal.reason })
})

if (resolverResult.error) {
  if (resolverResult.error.type === 'aborted') {
    return Promise.reject(resolverResult.error.reason)
  }

  return /* create the TypeError, set "cause" */
}
```

JesusTheHun: It is true that the promise encapsulate the expectation of an aborted request.
Resolving it would mean that the expectation has been reached, which is logically correct.
However, in our case, we use a promise as a poor man's event : the promise rejection carries no expectation and therefore no logical meaning. We just want to be notified when the abortion happens.
From there, this promise is a mere async tool.
Resolving it, then check the race winner, then throw a custom object, check the resolverResult resolution value, all of that to ultimately throw the error we had in the first place, seems unnecessary to me.
On a side note, in the end, I'm not the maintainer of the project, if you feel more comfortable with this extra layer, do it, this is your call :)

kettanaito: I ended up leaving the explicit rejection promise check but instead of coupling the request rejection promise and the resolver error, I reject the request promise with the abort rejection reason. Also, that is hell of a sentence to write.

kettanaito:
@JesusTheHun, hi! The changes look great from my side, is there anything else we should consider before publishing this?

src/interceptors/fetch/index.ts
```diff
@@ -46,13 +47,27 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
...
 
+      const signal = interactiveRequest.signal
+      const requestAborted = new DeferredPromise()
+
+      signal.addEventListener(
+        'abort',
+        () => {
+          requestAborted.reject(signal.reason)
+        },
+        { once: true }
```
kettanaito, line 58: I've made sure we only react to the abort event once to make sure there's no event emitter memory leak when cloning requests (each request clone will also preserve previously attached listeners, I believe; I recall an issue about this in the past).

kettanaito (approved):
This is awesome. Thank you for your contribution, @JesusTheHun! 👏

JesusTheHun:
LGTM 👌 Now #398 😁💪

kettanaito:
There's an issue with the automated releases in this repo, I will look into it and see if I can figure it out before releasing this. If I notice it taking much time, I will release manually but I'd pretty much love to solve the automation issue.

kettanaito:
## Released: v0.24.1 🎉

This has been released in v0.24.1!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.24.1)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.24.1)

---

src/interceptors/fetch/index.ts
sag-tobias-frey: This line of code seems to break for me when I'm not setting an AbortController because signal is `null` in that case.

JesusTheHun: > This line of code seems to break for me when I'm not setting an AbortController because signal is `null` in that case.

When no abort signal is has been provided to the intercepted request, one is injected automatically, so the signal cannot be null.
If you are facing any issue, please open one.

PR #401, opened 2023-08-21
set x-request-id header in XMLHttpRequestController only for node js

avivasyuta:


avivasyuta:
@kettanaito @msutkowski @marcosvega91 @timdeschryver could you please review this

Issue #345, opened 2023-02-22
Interceptor presets are no longer accessible for import


cdimitroulas:
See this comment: https://github.com/mswjs/interceptors/pull/313#issuecomment-1436419369

It seems that when we added bundling for ESM/CJS, we didn't consider some of the imports that are documented in the project's documentation that say that presets should be imported like so:
```ts
import browserInterceptors from '@mswjs/interceptors/lib/presets/browser'
```

kettanaito:
Hi, @cdimitroulas. This makes sense, I think we forgot the "/presets" exports. Is this only the documentation problem though?

cdimitroulas:
I haven't reviewed the documentation to see if other things are invalid, I was just opening this issue based on the comment I linked so that it doesn't get lost since the comment was on a closed PR.

blaenk:
Yeah thank you @cdimitroulas I was about to create a tracking issue as well. Subscribed!

kettanaito:
@blaenk, thanks for discovering this. Please, would you be interested in opening a pull request? This should be a matter of adding that path to those presets in "exports" in package.json. I can help you with the code review and getting this fix released.

blaenk:
Apologies @kettanaito, this slipped, didn't mean to leave you hanging. I had responded to a similar request here https://github.com/mswjs/interceptors/pull/313#issuecomment-1437730466 and didn't see this one.

Thank you so much for working on it!

kettanaito:
## Released: v0.22.5 🎉

This has been released in v0.22.5!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.22.5)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.22.5)

Make sure to always update to the latest version (`npm i @mswjs/interceptors@latest`) to get the newest features and bug fixes.

---

_Predictable release automation by [@ossjs/release](https://github.com/ossjs/release)_.

kettanaito:
Hey, @blaenk, no worries! I think presents should be correctly exported now. Thanks for keeping an eye on these changes. I feel really good about the exports we've got now!

blaenk:
Hey there, it's been a while and I just recently upgraded this package from the version I pinned from before this issue, and I noticed that unfortunately there is still an issue with the exports.

Running `tsc` directly also leads to this issue:

```
error TS2307: Cannot find module '@mswjs/interceptors/presets/browser' or its corresponding type declarations.

[...]

PR #421, opened 2023-09-10
fix: add root-level "browser" export

### LINKED ISSUE #476: Error using BatchInterceptor with browserInterceptors in browser environment.
Still getting the issue from here: https://github.com/mswjs/interceptors/pull/361#issuecomment-1455223410

```
Type 'readonly [FetchInterceptor, XMLHttpRequestInterceptor]' is not assignable to type 'readonly Interceptor<any>[]'.
  Type 'FetchInterceptor | XMLHttpRequestInterceptor' is not assignable to type 'Interceptor<any>'.
    Type 'FetchInterceptor' is not assignable to type 'Interceptor<any>'.
      Types have separate declarations of a private property 'symbol'.ts(2322)
```

```typescript
import { BatchInterceptor } from '@mswjs/interceptors';
import browserInterceptors from '@mswjs/interceptors/presets/browser';

  name: 'interceptor',
  interceptors: browserInterceptors,
});
```

This looks to be because the exports in package.json for "." does not include a browser config.
It should be something like this:
```json
{
 // ...
  "exports": {
    ".": {
      "browser": {
        "types": "./lib/browser/index.d.ts",
        "require": "./lib/browser/index.js",
        "import": "./lib/browser/index.mjs",
        "default": "./lib/browser/index.js"
      },
      "types": "./lib/node/index.d.ts",
      "require": "./lib/node/index.js",
      "import": "./lib/node/index.mjs",
      "default": "./lib/node/index.js"
    },
    // ...
  },
  // ...
}

### PR DESCRIPTION
**Pull request by blaenk**
- See https://github.com/mswjs/interceptors/issues/345#issuecomment-1712962587
- Fixes #476 

This, in combination with configuring my `tsconfig.ts` to 'activate' the `browser` condition when evaluating this package's `exports` field via [`customConditions`](https://www.typescriptlang.org/tsconfig#customConditions), allows typescript to automatically (well, via `exports`) pick all of the right types and compile cleanly!

Here is the [`customConditions`](https://www.typescriptlang.org/tsconfig#customConditions) field in my `tsconfig.ts`:

```json
"customConditions": ["browser"]
```

I think having the user enable this is not onerous at all and probably ideal if they are indeed targeting the browser. Compare that to requiring the user to explicitly override paths via `paths` and then requiring the bundler to know about that (via some tsconfig-paths plugin).

This way, it will ignore the root level `"types"` if `browser` is activated/true, in which case it instead uses the browser ones, solving the issue where otherwise the root typings file was used, so the `browserInterceptors` would not match with the `BatchInterceptor` since the one being used was the node one.

I'm not sure what other effects this would have, but I don't see why it would have any, and it appears to be consistent with everything else we're doing with the `browser` condition.


blaenk:
@kettanaito any chance you can look at this? I would appreciate it!

Thank you for working on this project!

package.json
```diff
@@ -7,6 +7,12 @@
   "types": "./lib/node/index.d.ts",
   "exports": {
     ".": {
+      "browser": {
+        "types": "./lib/browser/index.d.ts",
+        "require": "./lib/browser/index.js",
+        "import": "./lib/browser/index.mjs",
+        "default": "./lib/browser/index.js"
+      },
       "types": "./lib/node/index.d.ts",
```
kettanaito, line 16: `types` field must _always be the first_.

kettanaito (approved):
This looks fantastic! Thanks for fixing this, @blaenk.

kettanaito:
## Released: v0.25.12 🎉

This has been released in v0.25.12!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.25.12)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.25.12)

---

steve-zhu-sh:
Woo didn't realize this had already been addressed, thanks for the fix!

package.json
steve-zhu-sh: @kettanaito Why does the types field have to be the first here? It seems like it stops browser.types from being resolved if it's first.

blaenk: https://nodejs.org/api/packages.html#community-conditions-definitions

> "types" - can be used by typing systems to resolve the typing file for the given export. This condition should always be included first.

Doesn't explain why. I haven't tested myself but you might be able to fix your situation if you use something like what I posted in the OP:

```typescript
"customConditions": ["browser"]
```

That is specific to webpack, but if you don't use webpack you can read the [documentation](https://www.typescriptlang.org/tsconfig#customConditions) for it to see if there is an equivalent in whatever you might be using.

To be clear, I haven't tested if this library works in my particular case after the change Artem made, I'm simply pointing out that the solution for me had two parts: this change + the custom condition.

steve-zhu-sh: I see, I'm assuming that it actually means that types needs to come first within a group of exports, but since browser is separate it can come first. It's just that types needs to come first with the `browser` export config.

Yeah I was using the exact config that you posted `"customConditions": ["browser"]`. There's an issue there because it's resolving types first, when it should be resolving browser.types instead.

steve-zhu-sh: Yeah I think your original fix here https://github.com/mswjs/interceptors/issues/345#issuecomment-1712962587 is the correct fix. And I think it correctly interprets the line in the docs. From what I know about `customConditions` in tsconfig, is that it'll cause the listed custom conditions to be considered for export resolution.
Reference: https://www.typescriptlang.org/tsconfig#customConditions

Because of that we're supposed to resolve `browser` -> `types` here first.
```json
    ".": {
      "browser": {
        "types": "./lib/browser/index.d.ts",
        "require": "./lib/browser/index.js",
        "import": "./lib/browser/index.mjs",
        "default": "./lib/browser/index.js"
      },
      "types": "./lib/node/index.d.ts",
      "require": "./lib/node/index.js",
      "import": "./lib/node/index.mjs",
      "default": "./lib/node/index.js"
    },
```

Instead with this config, it resolves `types` first which is incorrect.

```json
    ".": {
      "types": "./lib/node/index.d.ts",
      "browser": {
        "types": "./lib/browser/index.d.ts",
        "require": "./lib/browser/index.js",
        "import": "./lib/browser/index.mjs",
        "default": "./lib/browser/index.js"
      },
      "require": "./lib/node/index.js",
      "import": "./lib/node/index.mjs",
      "default": "./lib/node/index.js"
    },
```

PR #430, opened 2023-09-18
fix(Interceptor): wrap the "once" event listener, ditch "untilIdle" Promise

kettanaito:
- Follow-up to #427 

## Changes

> These are internal changes and have no effect on the public API. 

This change drops the internal implementation of `AsyncEventEmitter` used to await all `request` listeners from being done. Instead, it introduces a `RequestConrtoller` that just holds a response Promise to whichever surface calling `.respondWith()`. 

The request listener await logic is achieved by:

- Adding a `once` request listener that always responds to the intercepted request with `undefined. using the request controller. This leverages the native listeners execution order to provide a fall-safe listener in case the consumer hasn't handled the request in any way.
- Adding a new `emitAsync` utility to account for _async_ request listeners. `emitAsync` executes all the listeners for the given events _awaiting them_ (i.e. in order).


src/utils/AsyncEventEmitter.ts
```diff
@@ -79,6 +55,16 @@ export class AsyncEventEmitter<
     return super.on(eventName, wrappedListener)
   }
 
+  public once<EventName extends keyof Events>(
+    eventName: EventName,
+    listener: Listener<any>
+  ): this {
+    const wrappedListener = this.wrapListener(eventName, listener)
```
kettanaito, line 62: `once` listener has to be wrapped the same as the `on` listener to open the event queue once the respective event is emitted.

kettanaito:
## Released: v0.25.2 🎉

This has been released in v0.25.2!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.25.2)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.25.2)

---

PR #432, opened 2023-09-18
fix(ClientRequest): ignore request destroyed state when responding with a mocked response

### LINKED ISSUE #426: Responses with delay hang forever with "got"
I'm not sure if it's a `got` error or `mswjs/interceptors` one.
This is working with `axios` and native `fetch`

```js
const { BatchInterceptor } = require('@mswjs/interceptors')
const { default: nodeInterceptors } = require('@mswjs/interceptors/presets/node')
const got = require('got')
const axios = require('axios')

    name: 'my-interceptor',
    interceptors: nodeInterceptors,
  })
interceptor.apply();
interceptor.on('request', function ({ request, requestId }) {
  return new Promise(resolve => {
    setTimeout(() => {
      console.log(request.url);
      request.respondWith(new Response('', { status: 200 }))
       resolve()
     },
    2000)
  })
})

(await got('http://example.test')).statusCode // stuck
(await axios.get('http://example.test')).status // work
```

kettanaito:
- Fixes #426


kettanaito:
Hey, @mikicho 👋 I've tried reproducing your issue from #426 in an integration test but it passes reliably. Do you spot any difference that might be causing this?

mikicho:
@kettanaito 
Add after the sleep
```
  interceptor.on('request', async () => {
    await sleep(750)
    request.respondWith(new Response('', { status: 200 }))
  })
```

example:

```
const got = require('got')

  name: 'my-interceptor',
})
interceptor.apply();
interceptor.on('request', async ({request}) => {
  await sleep(750)
  console.log(request.url)
  request.respondWith(new Response('', { status: 200 }))
})

got('http://example.test')
  .then(console.log)
  .catch(console.error)
  .finally(console.trace)

function sleep(duration) {
  return new Promise((resolve) => {
    setTimeout(resolve, duration)
  })
}
```

mikicho:
Another thing I found: you may want to try an address that does not exist like: nowhere.com
https://github.com/mswjs/interceptors/issues/426#issuecomment-1725813242

kettanaito:
I found it to be related to the `sleep` in the request listener. If I observe how the listeners get called (after #427), I can see that the library doesn't wait for the `sleep` to finish before calling the next listener. That's. . . odd.

mikicho:
Where it doesn't await?

Also, You can pass `retry: 0` to got

kettanaito:
Talking about this `sleep`:

https://github.com/mswjs/interceptors/pull/432/files#diff-d18cec8d1031bb154cb3fea3a0fefcf1c5ff3ee1f1db6c6346286e266832330bR46

kettanaito:
## Root cause

I found the root cause for this. This line in `strict-event-emitter` implements `once` listeners by wrapping them in a higher function:

https://github.com/open-draft/strict-event-emitter/blob/b5da3bffe6b7913fe5ea27e3abe0285741765424/src/Emitter.ts#L81-L84

Because that function is _not_ `async`, awaiting one-time listeners with `await listener(...data)` will resolve _immediately_, even if the wrapped listener is an async function with a `sleep()`. This is also the reason why this issue isn't reproducible with `.on()`, where the listener isn't wrapped and its reference is preserved.

kettanaito:
Alas, the issue with `got` remains even with the fix to `strict-event-emitter`. There must be something else in play.

kettanaito:
## Looking into `got`

When comparing the passing test (without delay) and a failing one, I've noticed one difference in how Got handles it. When there's _any_ delay in the request listener, `response.request.aborted` in Got becomes `true` right here:

https://github.com/sindresorhus/got/blob/0da732f4650c398f3b2fea672f8916e6c7004c8f/source/as-promise/index.ts#L47-L50

This prevents Got from reading the response body initiated here:

https://github.com/sindresorhus/got/blob/0da732f4650c398f3b2fea672f8916e6c7004c8f/source/as-promise/index.ts#L55

And then forwarded to [here](https://github.com/sindresorhus/got/blob/0da732f4650c398f3b2fea672f8916e6c7004c8f/source/core/index.ts#L2150-L2154) and [here](https://github.com/sindresorhus/got/blob/0da732f4650c398f3b2fea672f8916e6c7004c8f/source/core/index.ts#L2602). Without this response body reading, Got request pends forever despite receiving the response (the `response` event gets emitted correctly). 

I'm trying to see what aborts the request in case of delay but can't find anything. I've added a setter on `set aborted() {}` in `NodeClientRequest` to no avail. I can confirm that Got uses the correctly patched `http.get`/`http.request` functions, which means patched `NodeClientRequest`. Perhaps the native Node.js logic sets `aborted` on the request?

mikicho:
@kettanaito 
`aborted` is a getter: 
https://github.com/sindresorhus/got/blob/0da732f4650c398f3b2fea672f8916e6c7004c8f/source/core/index.ts#L2745

When debugging it, I saw that the `this[kRequest]?.destroyed` was true. Is there any chance we don't mock something related to the socket that causes the NodeRequest to fail? ([more context](https://github.com/mswjs/interceptors/issues/426#issuecomment-1725813242))

kettanaito:
@mikicho, would you have a minute to review these changes? I could use your eyes on this 🙏

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -229,7 +237,17 @@ export class NodeClientRequest extends ClientRequest {
...
+          'received mocked response:',
+          mockedResponse.status,
+          mockedResponse.statusText
+        )
+
+        /**
+         * @note Ignore this request being destroyed by TLS in Node.js
+         * due to connection errors.
+         */
+        this.destroyed = false
```
mikicho, line 250: Instead of ignoring the failed request, can we handle the error by monkey-patching whatever is needed? This may not be urgent, and the issue may be resolved when/if we implement mocking at the socket level.
(https://github.com/mswjs/interceptors/pull/375)

kettanaito: I'm not sure what we could monkey-patch in this case though. 

This assignment happens only for mocked responses. When you have a mock response returned, it shouldn't matter if there's been connection errors or not. This way we can keep the original connection logic from Node (handy for when you don't have mocked responses) but ignore any errors that occurred because they are irrelevant when providing a mock response. 

Note that we still do the following:

1. Capture any errors occurring during the request, which includes connection errors;
2. If the consumer hasn't responded to the request, we _replay those errors_ so they are never ignored when the request was performed as-is.

mikicho: > This assignment happens only for mocked responses...

I'm not an expert, but what if you want mock connection-related behavior, for example, Nock lets you [mock delay in the connection](https://github.com/nock/nock/#delay-the-connection) and mock delay in the response. 

Anyway, this is another feature or something anyway, so LGTM.

kettanaito: Having more fine-grained delays would certainly be a feature worth exploring. For now, whichever delays you specify in the request listener are treated as respond delays (they trigger _after_ the connection has been made and the request has been written).

mikicho: Thanks! 
We will probably need to to keep parity with what Nock have now if we want to move Nock to `mswjs/interceptors`

kettanaito: Let's raise this as a new feature request and discuss it in more detail.

kettanaito:
## Released: v0.25.3 🎉

This has been released in v0.25.3!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.25.3)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.25.3)

---

PR #441, opened 2023-09-21
fix(ClientRequest): respect "options.path" over "url.pathname"

mikicho:
Currently, we ignore the path from the second argument.

@kettanaito 
When I do: 
```js
url.pathname = "/user?id=123"
```
It's automatically encodes it. Are we ok with this? see the `exepct` request.url line

### DIFF
--- a/src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.test.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.test.ts
@@ -212,6 +212,40 @@ it('handles [URL, RequestOptions, callback] input', () => {
   expect(callback?.name).toEqual('cb')
 })
 
+it('handles [URL, RequestOptions] where options override the URL', () => {
+  const firstResult = normalizeClientRequestArgs(
+    'http:',
+    new URL('http://example.com/path-from-url'),
+    {
+      hostname: 'host-from-options.com',
+    }
+  )
+  expect(firstResult[0].href).toBe('http://host-from-options.com/path-from-url')
+  expect(firstResult[1]).toMatchObject({
+    host: 'host-from-options.com',
+    path: '/path-from-url',
+  })
+
+  const secondResult = normalizeClientRequestArgs(
+    'http:',
+    new URL('http://example.com/path-from-url?a=b&c=d'),
+    {
+      hostname: 'host-from-options.com',
+      path: '/path-from-options',
+      port: 1234,
+    }
+  )
+  // Must remove the query string since it's not specified in "options.path"
+  expect(secondResult[0].href).toBe(
+    'http://host-from-options.com:1234/path-from-options'
+  )
+  expect(secondResult[1]).toMatchObject({
+    host: 'host-from-options.com:1234',
+    path: '/path-from-options',
+    port: 1234,
+  })
+})
+
 it('handles [RequestOptions, callback] input', () => {
   const initialOptions = {
     method: 'POST',
@@ -397,8 +431,8 @@ it('respects custom "options.path" over URL path with query string', () => {
     }
   )
 
-  // Must replace the path but preserve the query string.
-  expect(url.href).toBe('http://example.com/path-from-options?a=b&c=d')
+  // Must replace both the path and the query string.
+  expect(url.href).toBe('http://example.com/path-from-options')
   expect(options.protocol).toBe('http:')
   expect(options.host).toBe('example.com')
   expect(options.hostname).toBe('example.com')
--- a/src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.test.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.test.ts
@@ -212,21 +212,23 @@ it('handles [URL, RequestOptions, callback] input', () => {
   expect(callback?.name).toEqual('cb')
 })
 
-it('handles [URL, RequestOptions] where options override the URL', () => {
-  const firstResult = normalizeClientRequestArgs(
+it('handles [URL, RequestOptions] where options have custom "hostname"', () => {
+  const [url, options] = normalizeClientRequestArgs(
     'http:',
     new URL('http://example.com/path-from-url'),
     {
       hostname: 'host-from-options.com',
     }
   )
-  expect(firstResult[0].href).toBe('http://host-from-options.com/path-from-url')
-  expect(firstResult[1]).toMatchObject({
+  expect(url.href).toBe('http://host-from-options.com/path-from-url')
+  expect(options).toMatchObject({
     host: 'host-from-options.com',
     path: '/path-from-url',
   })
+})
 
-  const secondResult = normalizeClientRequestArgs(
+it('handles [URL, RequestOptions] where options contain "host" and "path" and "port"', () => {
+  const [url, options] = normalizeClientRequestArgs(
     'http:',
     new URL('http://example.com/path-from-url?a=b&c=d'),
     {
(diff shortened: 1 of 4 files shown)


src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.ts
```diff
@@ -110,11 +110,13 @@ export function normalizeClientRequestArgs(
   // and derive request options from it.
   else if (args[0] instanceof URL) {
     url = args[0]
-    logger.info('first argument is a URL:', url)
 
     options = resolveRequestOptions(args, url)
     logger.info('derived request options:', options)
 
+    url.pathname = options.path || ''
```
kettanaito, line 117: This should be redundant as in this clause the request has been constructed given a `URL` instance instead of the options. You can see above we call `resolveRequestOptions` to derive request options _from_ the given URL. This means that `options.path` should already be derived from `url.pathname` in that function. 

Is there a particular issue that surrounds this? We may start with a test to confirm it.

mikicho: Given this request:
```js
http.get(new url.URL('http://example.test'), { path: '/hello' })
```
The `options.path` is derived correctly, but we may want to populate the `url.pathname` with the correct value because we pass this to the interceptor handler.

> We may start with a test to confirm it.

See the test in the PR.

kettanaito: Then I guess it should be 

```ts
url.pathname = options.path || url.pathname
```

We need to check which value takes precedence: the explicit `url.pathname` or `options.path`.

```ts
http.get(new URL('https://example.com/one'), { path: '/two' })
// what would be the actual request URL?
```

mikicho: Great point! 
image
I guess the `path` option wins

kettanaito: Got it. But that fallback value still makes sense otherwise we'll lose the original path set on the URL.

test/modules/http/intercept/http.request.test.ts
```diff
@@ -229,3 +229,27 @@ it('intercepts an http.request given RequestOptions without a protocol', async (
 
   expect(requestId).toMatch(UUID_REGEXP)
 })
+
+it('intercepts an http.request with three argument form: URL, options, and callback', async () => {
```
kettanaito, line 233: Would be also great to add that ambiguous scenario when both URL and options include `path` (the one we talked about in the comments). Just so we reflect that behavior correctly as well.

src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.ts
mikicho: I updated the PR 👍

src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.ts
```diff
@@ -110,11 +110,13 @@ export function normalizeClientRequestArgs(
   // and derive request options from it.
   else if (args[0] instanceof URL) {
     url = args[0]
-    logger.info('first argument is a URL:', url)
 
     options = resolveRequestOptions(args, url)
     logger.info('derived request options:', options)
 
+    url.pathname = options.path || url.pathname
```
kettanaito, line 117: Maybe it's a bit nitpicky, but what do you think if we handle this case rather explicitly, including logging?

```ts
if (options.path) {
  logger.info('found explicit "path" option ("%s"), overriding URL pathname ("%s")', options.path, url.pathname)

  url.pathname = path
}
```

This would help in debugging any related issues a lot.

mikicho: Good idea!
I'll add it when I get back home

mikicho: Done

test/modules/http/intercept/http.request.test.ts
mikicho: done

src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.ts
```diff
@@ -110,11 +110,18 @@ export function normalizeClientRequestArgs(
   // and derive request options from it.
   else if (args[0] instanceof URL) {
     url = args[0]
-    logger.info('first argument is a URL:', url)
 
     options = resolveRequestOptions(args, url)
     logger.info('derived request options:', options)
 
+    if (options.path) {
```
kettanaito, line 117: We seem to have a regression here: since `options` is derived from the `url` a few lines above, what ends up happening is this:

```js
'http://example.com/resource?a=b' // url
{ path: '/resource?a=b' } // options

url.pathname = options.path
```

This will duplicate the query string on the URL since `options.path` rightfully includes query string. 

I will fix this.

src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.ts
```diff
@@ -112,6 +112,27 @@ export function normalizeClientRequestArgs(
     url = args[0]
     logger.info('first argument is a URL:', url)
 
+    // Check if the second provided argument is RequestOptions.
+    // If it is, check if "options.path" was set and rewrite it
+    // on the input URL.
+    // Do this before resolving options from the URL below
+    // to prevent query string from being duplicated in the path.
+    if (
```
kettanaito, line 120: I've added this check to only respect `options.path` if the second argument to the request is request options. Previously, we checked this on _resolved_ options, which meant options that have already inherited properties from the `URL` instance, potentially.

test/modules/http/intercept/http.request.test.ts
```diff
@@ -230,6 +230,30 @@ it('intercepts an http.request given RequestOptions without a protocol', async (
   expect(requestId).toMatch(UUID_REGEXP)
 })
 
+it('intercepts an http.request path in url and options', async () => {
```
kettanaito, line 233: I've left only one of these actual request tests since they are quite verbose and assert many things apart from the correct request options. I've moved the majority of tests to the unit test for `normalizeClientRequestArgs`, which also allowed us to cover more use cases.

src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.test.ts
```diff
@@ -371,3 +371,50 @@ it('merges URL-based RequestOptions with the custom RequestOptions', () => {
...
+  expect(options.path).toBe('/path-from-options')
+})
+
+it('preserves URL query string', () => {
+  const [url, options] = normalizeClientRequestArgs(
+    'http:',
+    new URL('http://example.com/resource?a=b&c=d')
+  )
+
+  expect(url.href).toBe('http://example.com/resource?a=b&c=d')
```
kettanaito, line 414: I believe this is the correct behavior. I think we also need to add a test if the `options.path` provides a query string. I _think_ in that case, it must rewrite the query string on the URL.

kettanaito: No, it looks like I'm wrong about this:

```js
http.get(new URL('http://example.com/resource?a=b&c=d'), { path: 'foo' })
```

This will produce a request where `path = 'foo'` without the query string! Oh boy.

kettanaito: I've pushed a fix for this, the unit tests are passing.

kettanaito:
There are more ways to construct `http.ClientRequest` than there are stars in the sky.

kettanaito:
@mikicho, everything looks good from my side. Let me know if we haven't missed anything, and we can merge then. Thanks!

src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.test.ts
mikicho: So many details! great catch!

mikicho:
LGTM

kettanaito:
## Released: v0.25.5 🎉

This has been released in v0.25.5!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.25.5)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.25.5)

---

PR #442, opened 2023-09-22
feat(ClientRequest): implement "flushHeaders()"

### LINKED ISSUE #439: Support http.request flushHeaders
`flushHeaders` should kickstart the request (same as `end`): https://nodejs.org/api/http.html#requestflushheaders
Nock test: https://github.com/nock/nock/blob/main/tests/test_intercept.js#L964

kettanaito:
- Fixes #439

### DIFF
--- a/src/interceptors/ClientRequest/NodeClientRequest.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.ts
@@ -312,6 +312,17 @@ export class NodeClientRequest extends ClientRequest {
     return this
   }
 
+  flushHeaders(): void {
+    /**
+     * @note `.flusHeaders()` isn't strictly equivalent to `.end()`.
+     * Flushing headers writes request headers but not the request body,
+     * and we likely cannot replace it with `.end()`.
+     *
+     * @see https://github.com/nodejs/node/blob/c2cd74453e7d2794ad81cab63e68371e08bad04f/lib/_http_outgoing.js#L1161
+     */
+    this.end('')
+  }
+
   emit(event: string, ...data: any[]) {
     this.logger.info('emit: %s', event)
 
--- a/src/interceptors/ClientRequest/NodeClientRequest.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.ts
@@ -1,7 +1,7 @@
 import { ClientRequest, IncomingMessage } from 'http'
 import { Readable } from 'stream'
+import { invariant } from 'outvariant'
 import type { Logger } from '@open-draft/logger'
-import { until } from '@open-draft/until'
 import { DeferredPromise } from '@open-draft/deferred-promise'
 import type { ClientRequestEmitter } from '.'
 import {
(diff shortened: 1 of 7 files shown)


src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -312,6 +312,17 @@ export class NodeClientRequest extends ClientRequest {
...
 
+  flushHeaders(): void {
+    /**
+     * @note `.flusHeaders()` isn't strictly equivalent to `.end()`.
+     * Flushing headers writes request headers but not the request body,
+     * and we likely cannot replace it with `.end()`.
+     *
+     * @see https://github.com/nodejs/node/blob/c2cd74453e7d2794ad81cab63e68371e08bad04f/lib/_http_outgoing.js#L1161
+     */
+    this.end('')
```
kettanaito, line 323: I don't think this is correct (see the comment above).

We don't distinguish between writing headers and body in the interceptor, we only tap into the `write()` and `end()` methods.

kettanaito:
I've moved forward with the implementation for this. Right now, the main challenge is to distinguish between these two scenarios:

1. `.end()` + request without any body. In this case, `request.body` must be `null` in the listener.
2. `.flushHeaders()` + `.write()` + `.end()` (flushed request _with_ body). In this case, `request.body` should be a `ReadableStream` eventually populated by the `.write()`/`.end()` calls.

Since the `request` event is now emitted in `.flusHeaders()`, we somehow need to know if this request is going to have any body or not so its Fetch API representation has the correct `body` property set. That proves to be a bit problematic. 

@mikicho, maybe you've got some thoughts on this?

mikicho:
This is interesting. I didn't know about the `flushHeaders` function either, so I'm unfamiliar with its usage patterns.

From what I see, [Nock runs the interceptors logic] (https://github.com/nock/nock/blob/feaa66fa64d24f95937ef759cdd5a7ca07646f1c/lib/intercepted_request_router.js#L221), and don't wait for further `write` requests.

After reading your insights and Node code. I think Nock implemented it incorrectly, `flushHeaders` shouldn't kickstart the request, it just flush headers. From my understanding, in the context of a mocking library, we don't care about it. we just need wait to until the `end` event and then flush all data as usual.

So, I think we need to do nothing on `flushHeaders`. maybe patch the function and save the headers if the original function removes it from the buffer.

WDYT?

kettanaito:
> flushHeaders shouldn't kickstart the request, it just flush headers.

That's the same thing! Node.js buffers the request body _entirely_ before sending it to the server with `.end()`. If you imagine sending a really large request to the server, you may want to start streaming it to the server _early_, while the request body is still being written. That's precisely what `.flushHeaders()` does. 

> Also an important notice that "headers" here refers to the HTTP message headers, which is this:

```
GET /url HTTP/1.0
content-type: text/plain;charset=UTF-8
some-other-header: value
```

> Basically, the start of the HTTP message + the actual headers. This is why `flushHeaders()` means starting a request: that first part of the HTTP message is enough to kickstart a request.

mikicho:
```js
const req = http.request('http://httpstat.us/200')
req.setHeader('my-header', 'tes')
req.flushHeaders()

req.on('response', res => {
  const emitSpyRes = sinon.spy(res, 'emit')
  res.on('data', () => {})
  res.on('end', () => {
    console.log(emitSpy.getCalls().map(e => e.firstArg))
    console.log(emitSpyRes.getCalls().map(e => e.firstArg))
    done()
  })
  req.end()
})
```
This prints:
> [ 'socket', 'response', 'prefinish', 'finish' ]
[ 'resume', 'data', 'readable', 'end' ]

If I remove the `req.end()` the process "hang" (which is normal) and it prints:
> [ 'socket', 'response' ]
[ 'resume', 'data', 'readable', 'end' ]

So it seems like we want to run the interceptor (request) and return mockedResponse if it exists.

kettanaito:
Closing in favor of https://github.com/mswjs/interceptors/pull/515. Relying on the Socket will automatically support `.flushHeaders()`.

kettanaito:
## Released: v0.32.0 🎉

---

PR #447, opened 2023-09-24
fix(ClientRequest): support empty "username" and "password" basic auth values

mikicho:
According to [this test](https://github.com/nock/nock/blob/feaa66fa64d24f95937ef759cdd5a7ca07646f1c/tests/test_basic_auth.js#L34), passwords are not mandatory.

I couldn't find any specific spec for this, but I guess that if Nock has a test for it someone needed this.

### DIFF
--- a/src/interceptors/ClientRequest/utils/createRequest.test.ts
+++ b/src/interceptors/ClientRequest/utils/createRequest.test.ts
@@ -83,3 +83,43 @@ it('creates a fetch Request with an empty string body', async () => {
   expect(request.url).toBe('https://api.github.com/')
   expect(request.body).toBe(null)
 })
+
+it('creates a fetch Request with an empty username', async () => {
+  const clientRequest = new NodeClientRequest(
+    [
+      new URL('https://api.github.com'),
+      { auth: 'username:' },
+      () => {},
+    ],
+    {
+      emitter,
+      logger,
+    }
+  )
+  clientRequest.write('')
+
+  const request = createRequest(clientRequest)
+
+  expect(request.headers.get('Authorization')).toBe(`Basic ${btoa('username:')}`)
+  expect(request.url).toBe('https://api.github.com/')
+})
+
+it('creates a fetch Request with an empty password', async () => {
+  const clientRequest = new NodeClientRequest(
+    [
+      new URL('https://api.github.com'),
+      { auth: ':password' },
+      () => {},
+    ],
+    {
+      emitter,
+      logger,
+    }
+  )
+  clientRequest.write('')
+
+  const request = createRequest(clientRequest)
+
+  expect(request.headers.get('Authorization')).toBe(`Basic ${btoa(':password')}`)
+  expect(request.url).toBe('https://api.github.com/')
+})
--- a/src/interceptors/ClientRequest/utils/createRequest.test.ts
+++ b/src/interceptors/ClientRequest/utils/createRequest.test.ts
@@ -84,7 +84,7 @@ it('creates a fetch Request with an empty string body', async () => {
   expect(request.body).toBe(null)
 })
 
-it('creates a fetch Request with an empty username', async () => {
+it('creates a fetch Request with an empty password', async () => {
   const clientRequest = new NodeClientRequest(
     [
       new URL('https://api.github.com'),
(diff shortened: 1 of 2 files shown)


kettanaito (approved):
I think you're right. It seems to be possible to construct a URL with `username` alone:

```js
let u = new URL('https://example.com')
u.username = 'user'
u.href
// https://user@example.com/
```

src/interceptors/ClientRequest/utils/createRequest.ts
```diff
@@ -25,7 +25,7 @@ export function createRequest(clientRequest: NodeClientRequest): Request {
    * the request "Authorization" header.
    * @see https://github.com/mswjs/interceptors/issues/438
    */
-  if (clientRequest.url.username && clientRequest.url.password) {
+  if (clientRequest.url.username) {
     const auth = `${clientRequest.url.username}:${clientRequest.url.password}`
```
kettanaito, line 29: Let's then have a fallback for `clientRequest.url.password` to an empty string. If we don't, when the password is not provided this `auth` string will result in 

```
${username}:undefined
```

And produce a malformed encoded auth string.

kettanaito (changes requested):
This looks great but let's add a fallback to the password value to make sure it's an empty string if it's not set.

mikicho:
image
Seems like password without a user is ok too 😅

src/interceptors/ClientRequest/utils/createRequest.ts
```diff
@@ -25,7 +25,7 @@ export function createRequest(clientRequest: NodeClientRequest): Request {
...
     headers.set('Authorization', `Basic ${btoa(auth)}`)
```
kettanaito, line 30: Also, it's a good idea to add two more tests covering this:

1. username + no password set. Should result in the encoded `Basic USERNAME:` header.
2. user + password set explicitly to `""`. Should result in the same encoded header.

mikicho:
> add a fallback to the password value to make sure it's an empty string if it's not set.

I thought about it, but isn't it the job of the normalization function?

src/interceptors/ClientRequest/utils/createRequest.ts
mikicho: I'll add them tomorrow

kettanaito:
@mikicho, not really. The normalization happens between the variable nature of the `ClientRequest` options, but we still treat the given values as-is. In this case, if the `URL` has been given, we will operate with its `url.username` and `url.password` values as they were provided to prevent any confusion. 

Regarding the password being optional too, I suggest then making this check OR-OR and composing the `auth` string taking the optional nature of the two values into account. 

```js
if (username != null || password != null) {
  const resolvedUsername = username || ''
  const resolvedPassword = password || ''
  const auth = `${resolvedUsername}:${resolvedPassword}`
}
```

src/interceptors/ClientRequest/utils/createRequest.ts
kettanaito: Thank you so much!

mikicho:
@kettanaito PTAL.
Feel free to update the formatting :)

src/interceptors/ClientRequest/utils/createRequest.test.ts
```diff
@@ -83,3 +83,43 @@ it('creates a fetch Request with an empty string body', async () => {
   expect(request.url).toBe('https://api.github.com/')
   expect(request.body).toBe(null)
 })
+
+it('creates a fetch Request with an empty username', async () => {
```
kettanaito, line 87: Looks like the names of these test suites are mixed up: this one is about empty password, and the bottom one is about empty username, isn't it?

mikicho: oopsie. fixed 😅

kettanaito (approved):
Looks great to me. Yet another superb contribution, @mikicho 🚀

src/interceptors/ClientRequest/utils/createRequest.test.ts
kettanaito: Thank you!

kettanaito:
## Released: v0.25.5 🎉

This has been released in v0.25.5!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.25.5)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.25.5)

---

PR #453, opened 2023-09-28
fix(ClientRequest): support throwing custom errors in request listener

### LINKED ISSUE #452: Don't supress interceptors errors
Currently, we [suppress all errors for specific types](https://github.com/mswjs/interceptors/blob/v0.25.6/src/interceptors/ClientRequest/NodeClientRequest.ts#L357), which prevent the user to throw these error within the interceptors handler:
```js
const http = require('http');
const { inherits } = require('util');

  name: 'my-interceptor',
})
interceptor.apply();
interceptor.on('request', ({request}) => {
  throw new NetConnectNotAllowedError('example.test', '/')
});

http.get('http://nowhere.com/', res => {
  res.on('data', () => {})
  res.on('end', () => {
    console.log(1) 
  })
})
.on('error', e => {
  console.log('custom error') // never get called!!!
})

function NetConnectNotAllowedError(host, path) {
  Error.call(this)

  this.name = 'NetConnectNotAllowedError'
  this.code = 'ENETUNREACH'
  this.message = `Nock: Disallowed net connect for "${host}${path}"`

  Error.captureStackTrace(this, this.constructor)
}

inherits(NetConnectNotAllowedError, Error)
```

As I see it, we can do one of two things:

1. have a flag that indicates that the error is from the user and shouldn't be suppressed.
2. Set the default `resourceSource` to `pass` or to `undefined`

mikicho:
@kettanaito 

1. I throw a simple object with `code` property. Do we have a more solid `Error` object for tests?
2. I mainly used trial and error to determine that `this.afterConnectionPhase = true` belongs there. Please confirm if this is correct.

closes #452


src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -48,6 +48,7 @@ export class NodeClientRequest extends ClientRequest {
   }> = []
   private responseSource: 'mock' | 'bypass' = 'mock'
   private capturedError?: NodeJS.ErrnoException
+  private afterConnectionPhase: boolean = false
```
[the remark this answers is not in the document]
mikicho: Why not change the default to `pass`?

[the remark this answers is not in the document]
kettanaito: Because that'd mean for the class that it should treat this request as bypassed. Frankly, it should be neither until the request listener promise has been settled. So introducing something like `unknown` is a great internal improvement.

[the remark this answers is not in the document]
mikicho: Fixed

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -229,6 +227,7 @@ export class NodeClientRequest extends ClientRequest {
           'encountered resolver exception, aborting request...',
           resolverResult.error
         )
+        this.responseSource = 'mock'
```
kettanaito, line 229: This is incorrect. At this point, we only know that the listeners have finished running but there's no certainty whether they have returned anything. We should remove this particular line.

mikicho: IIUC, if we reach this line, the handler throws an error, and we want to send it back to the user.
I couldn't put it outside the `if` because of the reason you mentioned, so I put it twice inside each `if`.

mikicho: > If the response source remains unknown, and there's an error in the request listener, we will never forward it to the user, is that correct?

Yes, one of the tests failed because of this.

> Perhaps for error suppression we should have this.suppressErrors boolean flag?

Like the  `afterConnectionPhase` solution?

kettanaito: It looks like we can reuse the same `this.requestSent` new internal flag I talk about in [here](https://github.com/mswjs/interceptors/pull/462#discussion_r1343809809) for this purpose too! 

I believe once the request is sent (`.end()` is called), any other errors that arise should never be suppressed. Those errors are either:

1. Errors originating from the user-defined request listeners. 
2. Errors originating from Node when performing a bypassed request.

mikicho:
@kettanaito Fixed. PTAL

kettanaito:
Okay, I've pushed some refactoring to the error suppression logic so it behaves as we expect and account for the support of throwing network connection errors from the request listener. Explaining the changes below...

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -23,6 +23,18 @@ import { getRawFetchHeaders } from '../../utils/getRawFetchHeaders'
 
 export type Protocol = 'http' | 'https'
 
+enum HttpClientInternalState {
```
kettanaito, line 26: First, I'm adding a new `state` internal flag that represents the internal state of the `ClientRequest`. I take advantage of the numeric values of the enums in TypeScript to filter out all the states before a certain states like so:

```ts
if (this.state < HttpClientInternalState.MockLookupEnd) {
  // This is for states 0 - 3
}
```

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -146,15 +169,7 @@ export class NodeClientRequest extends ClientRequest {
...
-    this.isRequestSent = true
-
-    const requestId = uuidv4()
-
-    const [chunk, encoding, callback] = normalizeClientRequestEndArgs(...args)
-    this.logger.info('normalized arguments:', { chunk, encoding, callback })
-
-    // Write the last request body chunk passed to the "end()" method.
-    this.writeRequestBodyChunk(chunk, encoding || undefined)
+    this.state = HttpClientInternalState.Sent
```
kettanaito, line 172: I'm replacing `isRequestSent` boolean flag with the `Sent` state of the request.

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -367,20 +384,28 @@ export class NodeClientRequest extends ClientRequest {
...
-        this.responseSource === 'mock' &&
-        NodeClientRequest.suppressErrorCodes.includes(errorCode)
-      ) {
-        // Capture the first emitted error in order to replay
-        // it later if this request won't have any mocked response.
-        if (!this.capturedError) {
-          this.capturedError = error
-          this.logger.info('captured the first error:', this.capturedError)
+      // Suppress only specific Node.js connection errors.
+      if (NodeClientRequest.suppressErrorCodes.includes(errorCode)) {
```
kettanaito, line 388: This is the biggest change I've added. The previous condition was a bit misleading, so I rewrote it to correctly reflect when the connection errors must be suppressed. 

1. We are only interested in connection errors (those listed in the static `suppressErrorCodes` field on the class. I start the condition from that.
2. Until we've executed all the request listeners, we must suppress all the connection errors and capture the first one to replay it later just in case the listeners won't yield any mocked response and the request will be performed as-is.
3. If we know we are in the mocked response phase, simply ignore all the _connection_ errors (taking advantage of wrapping all these conditions in a higher-level condition). Those aren't relevant. 

The existing test suite has shown that connection errors can occur even in the mocked response phase. I think that's because mocks are applied generally faster than the original responses, and we have to account for that.

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -398,9 +423,8 @@ export class NodeClientRequest extends ClientRequest {
     encoding?: BufferEncoding | null,
     callback?: ClientRequestEndCallback | null
   ): this {
-    // Set the response source to "bypass".
-    // Any errors emitted past this point are not suppressed.
-    this.responseSource = 'bypass'
+    this.state = HttpClientInternalState.ResponseReceived
+    this.responseType = 'passthrough'
```
kettanaito, line 427: I kept the `responseSource`, just renamed it to `responseType`. The `ResponseReceived` phase alone doesn't imply the response type, and that's correct. That way request phases can be filtered out like I've shown in the first comment. But the `responseType` is an additional information that's important in certain scenarios, such as error suppression.

test/modules/http/compliance/http-errors.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+beforeAll(() => {
+  interceptor.apply()
+})
+
+afterAll(() => {
+  interceptor.dispose()
+})
+
+it('suppresses ECONNREFUSED error given a mocked response', async () => {
```
kettanaito, line 26: I've also removed the error suppression tests from the unit tests of `NodeClientRequest` and added them as compliance tests. 

In addition to that, I've added a test that ensures that it's possible to throw a connection error from the request listener and it won't be suppressed.

kettanaito (approved):
Thank you so much for all the awesome contribution, @mikicho! 👏 I think these changes are good to go now.

kettanaito:
I suspect that the failing CI has to do with the network on GitHub Actions:

```
 FAIL  modules/http/compliance/http-errors.test.ts > forwards ENETUNREACH error for a bypassed request
AssertionError: expected 'ENETUNREACH' to be 'EHOSTUNREACH' // Object.is equality
 ❯ modules/http/compliance/http-errors.test.ts:131:29
    129|   const requestError = await errorPromise
    130| 
    131|   expect(requestError.code).toBe('EHOSTUNREACH')
       |                             ^
    132|   expect(requestError.syscall).toBe('connect')
    133|   expect(requestError.address).toBe('2607:f0d0:1002:51::4')

  - Expected   "EHOSTUNREACH"
  + Received   "ENETUNREACH"
```

This same test produces the correct error locally. Will look into it...

mikicho:
As long as the state doesn't become too complicated, I think it's a great solution! 
Thanks!! LGTM

kettanaito:
> As long as the state doesn't become too complicated, I think it's a great solution!

I will keep an eye on this part of the code to make sure it doesn't grow out of hand. I think we should be covered now.

mikicho:
I'm working with WSL, and I get these errors:
image

kettanaito:
@mikicho, which version of Node.js is that? Also, I think the second error is related to the OS you're using.

mikicho:
Perhaps we should attempt to replicate this behavior using a more dependable error code? (e.g. ECONREFUSED)

kettanaito:
> Perhaps we should attempt to replicate this behavior using a more dependable error code? (e.g. ECONREFUSED)

Not sure what you mean here. Can you give an example of how we can achieve this substitute? 

I'm not proficient with network error codes but they seem to stand for quite different things. `ECONREFUSED` means the connection was OK DNS-wise but was refused by the server. It appears to be a different error than `EHOSTUNREACH`, where the requested host is outside of the current network (but the host itself and the network are OK).

kettanaito:
I've just ran the test on Node.js v20.6.1 and saw that it groups `ECONNREFUSED` into an `AggregateError` that doesn't have the `syscall` or `address` properties. I've adjusted those assertions accordingly. All tests are passing locally on that version of Node.

@mikicho, I'd still highly recommend `nvm` and use v18.14.x because we don't officially support Node.js v20 in Interceptor just yet.

mikicho:
This is with 18.14.2. The OS acts differently (I have WSL Debian)
image

> Not sure what you mean here. Can you give an example of how we can achieve this substitute?

I mean, we want to make sure we can throw suppressed errors from the `request` handler, and capture them before the request is sent. maybe we can use a more reliable error, which acts the same across OS

kettanaito:
@mikicho, I fixed that assertion to pass on Ubuntu. Also, looks like your branch has to be pulled to resolve the second failing assertion. Let me know how it goes!

mikicho:
image

kettanaito:
## Released: v0.25.7 🎉

---

PR #454, opened 2023-09-28
fix(ClientRequest): suppress "ENETUNREACH" socket error

### LINKED ISSUE #451: Support IPv6
Maybe I'm missing something but:

```js
const http = require('http')

  name: 'my-interceptor',
})
interceptor.apply();
interceptor.on('request', ({request}) => {
  request.respondWith(new Response('hey'))
});

http.get('http://[2607:f0d0:1002:51::4]:8080/', res => {
  res.on('data', () => {})
  res.on('end', () => {
    console.log(1) 
  })
})
```

Prints:
> RequestError: connect ENETUNREACH 2607:f0d0:1002:51::4:8080 - Local (:::0)

Logs:

```
02:24:25:826 [http] constructing the interceptor...
02:24:25:843 [http:apply] applying the interceptor...
02:24:25:843 [http] retrieved global instance: undefined
02:24:25:843 [http:apply] no running instance found, setting up a new instance...
02:24:25:844 [http:setup] native "http" module patched!
02:24:25:844 [http:setup] native "https" module patched!
02:24:25:844 [http] set global instance! http
02:24:25:844 [http:on] adding "request" event listener:
02:24:25:845 [http normalizeClientRequestArgs] arguments ["http://[2607:f0d0:1002:51::4]:8080/",null]
02:24:25:845 [http normalizeClientRequestArgs] using default protocol: http:
02:24:25:845 [http normalizeClientRequestArgs] first argument is a location string: http://[2607:f0d0:1002:51::4]:8080/
02:24:25:845 [http normalizeClientRequestArgs] created a url: "http://[2607:f0d0:1002:51::4]:8080/"
02:24:25:846 [http normalizeClientRequestArgs] request options from url: {"method":"GET","protocol":"http:","hostname":"2607:f0d0:1002:51::4","host":"[2607:f0d0:1002:51::4]:8080","path":"/","port":8080}
02:24:25:846 [http normalizeClientRequestArgs] request options not provided, deriving from the url "http://[2607:f0d0:1002:51::4]:8080/"
02:24:25:846 [http normalizeClientRequestArgs] resolved request options: {"method":"GET","protocol":"http:","hostname":"2607:f0d0:1002:51::4","host":"[2607:f0d0:1002:51::4]:8080","path":"/","port":8080}
02:24:25:847 [http normalizeClientRequestArgs] resolved fallback agent: {"_events":{},"_eventsCount":2,"defaultPort":80,"protocol":"http:","options":{"noDelay":true,"path":null},"requests":{},"sockets":{},"freeSockets":{},"keepAliveMsecs":1000,"keepAlive":false,"maxSockets":null,"maxFreeSockets":256,"scheduling":"lifo","maxTotalSockets":null,"totalSocketCount":0}
02:24:25:847 [http normalizeClientRequestArgs] has no default agent, setting the default agent for "http:"
02:24:25:847 [http normalizeClientRequestArgs] successfully resolved url: http://[2607:f0d0:1002:51::4]:8080/
02:24:25:847 [http normalizeClientRequestArgs] successfully resolved options: {"method":"GET","protocol":"http:","hostname":"2607:f0d0:1002:51::4","host":"[2607:f0d0:1002:51::4]:8080","path":"/","port":8080,"agent":{"_events":{},"_eventsCount":2,"defaultPort":80,"protocol":"http:","options":{"noDelay":true,"path":null},"requests":{},"sockets":{},"freeSockets":{},"keepAliveMsecs":1000,"keepAlive":false,"maxSockets":null,"maxFreeSockets":256,"scheduling":"lifo","maxTotalSockets":null,"totalSocketCount":0},"_defaultAgent":{"_events":{},"_eventsCount":2,"defaultPort":80,"protocol":"http:","options":{"keepAlive":true,"scheduling":"lifo","timeout":5000,"noDelay":true,"path":null},"requests":{},"sockets":{},"freeSockets":{},"keepAliveMsecs":1000,"keepAlive":true,"maxSockets":null,"maxFreeSockets":256,"scheduling":"lifo","maxTotalSockets":null,"totalSocketCount":0}}
02:24:25:847 [http normalizeClientRequestArgs] successfully resolved callback: res => {
  res.on('data', () => {})
  res.on('end', () => {
    console.log(1)
  })
}
02:24:25:851 [http:request GET http://[2607:f0d0:1002:51::4]:8080/] constructing ClientRequest using options: {"url":"http://[2607:f0d0:1002:51::4]:8080/","requestOptions":{"method":"GET","protocol":"http:","hostname":"2607:f0d0:1002:51::4","host":"[2607:f0d0:1002:51::4]:8080","path":"/","port":8080,"agent":{"_events":{},"_eventsCount":2,"defaultPort":80,"protocol":"http:","options":{"noDelay":true,"path":null},"requests":{},"sockets":{"2607:f0d0:1002:51::4:8080:":[{"connecting":true,"_hadError":false,"_parent":null,"_host":null,"_closeAfterHandlingError":false,"_readableState":{"objectMode":false,"highWaterMark":16384,"buffer":{"head":null,"tail":null,"length":0},"length":0,"pipes":[],"flowing":null,"ended":false,"endEmitted":false,"reading":false,"constructed":true,"sync":true,"needReadable":false,"emittedReadable":false,"readableListening":false,"resumeScheduled":false,"errorEmitted":false,"emitClose":false,"autoDestroy":true,"destroyed":false,"errored":null,"closed":false,"closeEmitted":false,"defaultEncoding":"utf8","awaitDrainWriters":null,"multiAwaitDrain":false,"readingMore":false,"dataEmitted":false,"decoder":null,"encoding":null},"_events":{},"_eventsCount":6,"_writableState":{"objectMode":false,"highWaterMark":16384,"finalCalled":false,"needDrain":false,"ending":false,"ended":false,"finished":false,"destroyed":false,"decodeStrings":false,"defaultEncoding":"utf8","length":0,"writing":false,"corked":0,"sync":true,"bufferProcessing":false,"writecb":null,"writelen":0,"afterWriteTickInfo":null,"buffered":[],"bufferedIndex":0,"allBuffers":true,"allNoop":true,"pendingcb":0,"constructed":true,"prefinished":false,"errorEmitted":false,"emitClose":false,"autoDestroy":true,"errored":null,"closed":false,"closeEmitted":false},"allowHalfOpen":false,"_sockname":null,"_pendingData":null,"_pendingEncoding":"","server":null,"_server":null}]},"freeSockets":{},"keepAliveMsecs":1000,"keepAlive":false,"maxSockets":null,"maxFreeSockets":256,"scheduling":"lifo","maxTotalSockets":null,"totalSocketCount":1},"_defaultAgent":{"_events":{},"_eventsCount":2,"defaultPort":80,"protocol":"http:","options":{"keepAlive":true,"scheduling":"lifo","timeout":5000,"noDelay":true,"path":null},"requests":{},"sockets":{},"freeSockets":{},"keepAliveMsecs":1000,"keepAlive":true,"maxSockets":null,"maxFreeSockets":256,"scheduling":"lifo","maxTotalSockets":null,"totalSocketCount":0}}}
02:24:25:852 [http:request GET http://[2607:f0d0:1002:51::4]:8080/] end []
02:24:25:852 [utils getUrlByRequestOptions] arguments []
02:24:25:853 [utils getUrlByRequestOptions] normalized args [null,null,null]
02:24:25:853 [http:request GET http://[2607:f0d0:1002:51::4]:8080/] normalized arguments: {"chunk":null,"encoding":null,"callback":null}
02:24:25:867 [http:request GET http://[2607:f0d0:1002:51::4]:8080/] emitting the "request" event for 2 listener(s)...
02:24:25:871 [http:request GET http://[2607:f0d0:1002:51::4]:8080/] emit: socket
02:24:25:871 [http:request GET http://[2607:f0d0:1002:51::4]:8080/] emit: error
02:24:25:871 [http:request GET http://[2607:f0d0:1002:51::4]:8080/] error:
 {"errno":-101,"code":"ENETUNREACH","syscall":"connect","address":"2607:f0d0:1002:51::4","port":8080}
node:events:492
      throw er; // Unhandled 'error' event
      ^

Error: connect ENETUNREACH 2607:f0d0:1002:51::4:8080 - Local (:::0)
    at internalConnect (node:net:1090:16)
    at defaultTriggerAsyncIdScope (node:internal/async_hooks:464:18)
    at node:net:1315:9
    at process.processTicksAndRejections (node:internal/process/task_queues:77:11)
Emitted 'error' event on _NodeClientRequest instance at:
    at _NodeClientRequest.emit (/.../nock/node_modules/@mswjs/interceptors/lib/node/chunk-44QGFZIT.js:353:18)
    at Socket.socketErrorListener (node:_http_client:495:9)
    at Socket.emit (node:events:514:28)
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -101,
  code: 'ENETUNREACH',
  syscall: 'connect',
  address: '2607:f0d0:1002:51::4',
  port: 8080
}

Node.js v20.6.1
```

mikicho:
I didn't add a test because we already tested this logic in the following:
1. emits the ENOTFOUND error connecting to a non-existing hostname given no mocked response
2. emits the ECONNREFUSED error connecting to an inactive server given no mocked response

closes #451

### DIFF
--- a/src/interceptors/ClientRequest/NodeClientRequest.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.ts
@@ -37,6 +37,7 @@ export class NodeClientRequest extends ClientRequest {
     'ECONNREFUSED',
     'ECONNRESET',
     'EAI_AGAIN',
+    'ENETUNREACH',
   ]
 
   private response: IncomingMessage
--- a/src/interceptors/ClientRequest/NodeClientRequest.test.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.test.ts
@@ -252,6 +252,35 @@ it('does not emit ECONNREFUSED error connecting to an inactive server given mock
   expect(response.statusMessage).toBe('Works')
 })
 
+it('does not emit ENETUNREACH error connecting to an inactive server given mocked response', async () => {
+  const emitter = new Emitter<HttpRequestEventMap>()
+  const handleError = vi.fn()
+  const request = new NodeClientRequest(
+    normalizeClientRequestArgs('http:', 'http://[2607:f0d0:1002:51::4]'),
+    { emitter, logger }
+  )
+
+  emitter.on('request', async ({ request }) => {
+    await sleep(250)
+    request.respondWith(
+      new Response(null, { status: 200, statusText: 'Works' })
+    )
+  })
+
+  request.on('error', handleError)
+  request.end()
+
+  const responseReceived = new DeferredPromise<IncomingMessage>()
+  request.on('response', (response) => {
+    responseReceived.resolve(response)
+  })
+  const response = await responseReceived
+
+  expect(handleError).not.toHaveBeenCalled()
+  expect(response.statusCode).toBe(200)
+  expect(response.statusMessage).toBe('Works')
+})
+
 it('sends the request body to the server given no mocked response', async () => {
   const emitter = new Emitter<HttpRequestEventMap>()
   const request = new NodeClientRequest(


src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -37,6 +37,7 @@ export class NodeClientRequest extends ClientRequest {
     'ECONNREFUSED',
     'ECONNRESET',
     'EAI_AGAIN',
+    'ENETUNREACH',
```
kettanaito, line 40: May I please ask you to add an integration test for this? You can grab your use case from the issue and put it `test/modules/http/regressions/http-socket-enetunreach.test.ts`.

mikicho:
@kettanaito I encountered something interesting.
I think there is a minor confusion between `hostname` and `host` in the `getUrlByRequestOptions` function,

AFAIK, the `host` is the one with the port, and the `hostname` is only the.. well.. the hostname.

image
image

I went ahead and made the required changes, but before I fix all the tests, (~7), please review the changes and let me know if I missed big time :)

src/utils/getUrlByRequestOptions.ts
```diff
@@ -75,14 +75,7 @@ function getPortByRequestOptions(
...
 
-function getHostByRequestOptions(options: ResolvedRequestOptions): string {
-  const { hostname, host } = options
-
-  // If the hostname is specified, resolve the host from the "host:port" string.
-  if (hostname != null) {
-    return hostname.replace(/:\d+$/, '')
-  }
-
+function getHostByRequestOptions(host: RequestOptions['host']): string {
```
kettanaito, line 78: Isn't `host` just `string | undefined`? I think we can annotate that manually, no need to reference the type from `RequestOptions`.

test/modules/http/regressions/http-socket-enetunreach.test.ts
```diff
@@ -0,0 +1,14 @@
+import { it, expect } from 'vitest'
+import { httpGet } from '../../../helpers'
+import { ClientRequestInterceptor } from '../../../../src/interceptors/ClientRequest'
+
+it('does not leave the test process hanging due to the custom socket timeout', async () => {
+  const interceptor = new ClientRequestInterceptor()
+  interceptor.apply();
```
kettanaito, line 7: May I please ask you to move the `ClientRequestInterceptor` instance outside of the test and enable/dispose of it in the `beforeAll`/`afterAll` hooks imported from `vitest`? Resource cleanup is important 😃

src/utils/getUrlByRequestOptions.ts
mikicho: I can. Just FYI, this is: `string | null | undefined`

kettanaito: I think explicit type is better here. This way if type defs for Node slip in a bug we won't just blindly forward it to the consumer.

```ts
host?: string | null
```

test/modules/http/regressions/http-socket-enetunreach.test.ts
mikicho: Fixed

src/utils/getUrlByRequestOptions.ts
```diff
@@ -166,7 +136,7 @@ export function getUrlByRequestOptions(options: ResolvedRequestOptions): URL {
     : ''
   logger.info('auth string:', authString)
 
-  const url = new URL(`${protocol}//${hostname}${path}`)
+  const url = new URL(`${protocol}//${host}${path}`)
```
kettanaito, line 139: You are correct regarding this change though. We should use `host` here, not `hostname`.

src/utils/getUrlByRequestOptions.ts
```diff
@@ -100,31 +93,8 @@ function getAuthByRequestOptions(
...
-  if (isRawIPv6Address(host)) {
-    return `[${host}]${portString}`
-  }
-
-  if (typeof port === 'undefined') {
-    return host
-  }
-
-  return `${host}${portString}`
+function getHostname(host: string): string {
```
kettanaito, line 96: Let's revert this back, there's most history behind this than first catches the eye. 

A bit of context: it's possible for the `ClientRequest` options to have all sorts of combinations of `host`, `hostname` and `port`. Any can be undefined and in some cases must be inferred from the other value so the interceptor would compose the right URL. 

There are some libraries out there can use `ClientRequest` rather. . . unconventionally. We must still support those usages, thus such a bullet-proof resolution for host/hostname.

test/modules/http/regressions/http-socket-enetunreach.test.ts
```diff
@@ -0,0 +1,14 @@
+import { it, expect } from 'vitest'
+import { httpGet } from '../../../helpers'
+import { ClientRequestInterceptor } from '../../../../src/interceptors/ClientRequest'
+
+it('does not leave the test process hanging due to the custom socket timeout', async () => {
```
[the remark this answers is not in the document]
mikicho: > I may be wrong on this but isn't the test case about IPv6 support?

Yes

Don't we suppress the error? so I make sure that there are no unhandled errors being thrown.
What assertion do you think we are missing?

src/utils/getUrlByRequestOptions.ts
mikicho: Fixed

test/modules/http/regressions/http-socket-enetunreach.test.ts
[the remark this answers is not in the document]
kettanaito: Apologies for the confusion. The test/assertions themselves are great! What I meant is the test's name. 

The error suppression is an implementation detail (thus, it's an implicit assertion in this test: no error — the positive assertion passes). Maybe we can phrase the test like something `supports requests with IPv6 request url`.

test/modules/http/regressions/http-socket-enetunreach.test.ts
```diff
@@ -0,0 +1,23 @@
...
+})
+
+afterAll(() => {
+  interceptor.dispose()
+})
+
+it('does not leave the test process hanging due to the custom socket timeout', async () => {
+  interceptor.apply();
+  interceptor.on('request', ({ request }) => {
+    request.respondWith(new Response('test'))
```
kettanaito, line 18: May also be a good idea to create a deferred promise for the `request.url` value here in the request listener to make sure we are providing the correct IPv6 URL in the listener. What do you think?

```ts
const listenerUrlPromise = new DeferredPromise<string)

  listenerUrlPromise.resolve(request.url)
  request.respondWith(new Response('test'))
})

// ...
const listenerRequestUrl = await listenerUrlPromise
expect(listenerRequestUrl).toBe('IPv6_URL_HERE')
```

Would have to move the `interceptor.on()` listener into the test to make this work.

src/utils/getUrlByRequestOptions.ts
mikicho: Sure, can you please be more specific?
What do we need to revert?

If the `host` is undefined, should we set the `hostname` as `host`?

kettanaito: We should keep those two helper functions as they were before: `getHostFromRequestOptions` and `getHostname`. I don't think we should change them, unless there's an issue in what they return.

test/modules/http/regressions/http-socket-enetunreach.test.ts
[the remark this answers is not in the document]
mikicho: OH I forgot to change the test name after copying the test file 😅 
Fixed.

Maybe should we change the test file as well to: `http-ipv6`?

test/modules/http/regressions/http-socket-enetunreach.test.ts
mikicho: Absolutely agree!
Fixed.

src/utils/getUrlByRequestOptions.ts
mikicho: > unless there's an issue in what they return

```js
if (hostname != null) {
    return hostname.replace(/:\d+$/, '')
  }
```
This deletes the last part of the ipv6: `2607:f0d0:1002:51::4` become to `2607:f0d0:1002:51:`

kettanaito: Nice! So there's a bug in the regexp, let's fix that 👀 

Would it be beneficial if we prioritized `host`, and if it exists, don't even bother to infer it from `hostname`?

mikicho: > Nice! So there's a bug in the regexp, let's fix that 👀

I think we don't need the regex at all. The hostname is a "subset" of the host. we don't need to remove the port from the hostname, because it doesn't have one.

I think:

1. if host exists: return host
2. if hostname exists: return hostname
3. otherwise: DEFAULT

kettanaito: 😆 I think that line meant `host`. Uhh, I need to re-read all of this... 

I think we could keep both utilities so we don't change much and can refactor this separately. For now, the intention behind those functions is the following:

### `getHostByRequestOptions`

```js
if (options.host) {
  return options.host
}

if (options.hostname) {
  if (options.port) {
    return `${options.hostname}:${options.port}`
  }

  return options.hostname
}

return null
```

### `getHostname`

```js
if (options.hostname) {
  return options.hostname
}

if (options.host) {
  // Check the presence of the port, and if it's present,
  // remove it from the host, returning a hostname.
  return hostWithoutPort(options.host)
}
```

test/modules/http/regressions/http-socket-enetunreach.test.ts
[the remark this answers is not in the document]
kettanaito: Sure. I'd say it's not a regression then, maybe `compliance/http-request-ipv6.test.ts`?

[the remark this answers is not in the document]
mikicho: Fixed

src/utils/getUrlByRequestOptions.ts
mikicho: I pushed a commit.

I fixed another confusion in `getPortByRequestOptions`. PTAL. I'm not sure if it's ipv6-proof.
Also, do we want `getHostByRequestOptions` and `getHostname` to return `null` or `DEFAULT_HOST`?

src/utils/getUrlByRequestOptions.ts
```diff
@@ -75,8 +75,20 @@ function getPortByRequestOptions(
...
-function getHostByRequestOptions(host?: string | null): string {
-  return host || DEFAULT_HOST
+function getHostByRequestOptions(options: ResolvedRequestOptions): string {
+  if (options.host) {
+    return options.host
+  }
+
+  if (options.hostname) {
+    if (options.port) {
+      return `${options.hostname}:${options.port}`
```
kettanaito, line 85: Just to be sure, would this work for IPv6 hosts as well?

mikicho: AFAIK:
If set, the hostname should be: `[ipv6]`
If set, the port should be: `8080`

In conjunction: `[ipv6]:8080` which is a valid `host

kettanaito: Thanks for clarifying!

src/utils/getUrlByRequestOptions.ts
```diff
@@ -166,7 +158,7 @@ export function getUrlByRequestOptions(options: ResolvedRequestOptions): URL {
     : ''
   logger.info('auth string:', authString)
 
-  const url = new URL(`${protocol}//${hostname}${path}`)
+  const url = new URL(`${protocol}//${host}${path}`)
```
kettanaito, line 161: I think we need to add this:

```
url.port = port
```

The port may not be a part of the `options.host` if we trust this test:

https://github.com/mswjs/interceptors/blob/2e98a714bc4364fb201313c1bcd326b3b2767a61/src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.test.ts#L313-L315

There's all sorts of shenanigans in how people construct requests but I trust that test suite had a real use case behind it so I'd keep it.

src/utils/getUrlByRequestOptions.ts
kettanaito: I'd say they should return `undefined` if there was no value set in the options. Then, the actual usage surface of those functions can have fallbacks to `DEFAULT_*`, if needed.

src/utils/getUrlByRequestOptions.ts
kettanaito: This will also fix the currently failing tests.

mikicho: What is the priority? 

```js
 port: '50176', 
 path: '/resource', 
 host: '127.0.0.1:8080', 
```

What's the port of this request?

kettanaito: > What is the priority?

I'd say `options.port` is the priority. We are threading on broken userland here but I think if port has been set explicitly we must always respect it. 

> Maybe we should do:

Yes, I think you're right!

mikicho:
@kettanaito 
About the `undefined`, there is a test that expects the `host` to be `localhost`.
How would you like to handle this?

```
12:43:31:468 [utils getUrlByRequestOptions] request options {}
12:43:31:469 [utils getUrlByRequestOptions] figuring out url from request options...
12:43:31:470 [utils getUrlByRequestOptions] protocol http:
12:43:31:470 [utils getUrlByRequestOptions] host undefined
12:43:31:470 [utils getUrlByRequestOptions] port undefined
12:43:31:470 [utils getUrlByRequestOptions] hostname undefined
12:43:31:470 [utils getUrlByRequestOptions] path /
12:43:31:470 [utils getUrlByRequestOptions] credentials undefined
12:43:31:470 [utils getUrlByRequestOptions] auth string:
12:43:31:470 [utils getUrlByRequestOptions] created url: "http://undefined/"
 ❯ src/utils/getUrlByRequestOptions.test.ts (12)
   ↓ returns a URL based on the basic RequestOptions [skipped]
   ↓ inherits protocol and port from http.Agent, if set [skipped]
   ↓ inherits protocol and port from https.Agent, if set [skipped]
   ↓ resolves protocol to "http" given no explicit protocol and no certificate [skipped]
   ↓ resolves protocol to "https" given no explicit protocol, but certificate [skipped]
   ↓ resolves protocol to "https" given no explicit protocol, but port is 443 [skipped]
   ↓ resolves protocol to "https" given no explicit protocol, but agent port is 443 [skipped]
   ↓ respects explicitly provided port [skipped]
   ↓ inherits "username" and "password" [skipped]
   × resolves host to localhost if none provided
   ↓ supports "host" instead of "hostname" and "port" [skipped]
   ↓ handles IPv6 host [skipped]

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/utils/getUrlByRequestOptions.test.ts > resolves host to localhost if none provided
AssertionError: expected 'undefined' to be 'localhost' // Object.is equality
 ❯ src/utils/getUrlByRequestOptions.test.ts:105:43
    103|
    104| it.only('resolves host to localhost if none provided', () => {
    105|   expect(getUrlByRequestOptions({}).host).toBe('localhost')
       |                                           ^
    106| })
    107|

  - Expected   "localhost"
  + Received   "undefined"

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

 Test Files  1 failed (1)
      Tests  1 failed | 11 skipped (12)
   Start at  12:43:30
   Duration  591ms (transform 80ms, setup 0ms, collect 80ms, tests 12ms)
```

mikicho:
I went through all [URL properties in the MDN](https://developer.mozilla.org/en-US/docs/Web/API/URL) site and tried to figure out which heuristic to use while parsing `RequestOptions` to `URL` using [Node.js docs](https://nodejs.org/api/http.html#httprequesturl-options-callback) and [source code](https://github.com/nodejs/node/blob/main/lib/_http_client).


## TL;DR

```js
const portString = typeof port !== 'undefined' ? `:${port}` : ''
const url = new URL(`${protocol}//${hostname}${portString}${path}`)
url.username = credentials?.username || ''
url.password = credentials?.password || ''
```

## Heuristic

### hash
No such concept in `RequestOptions`. It's part of the `path`

### host
It's important to notice that "URL hostname" != "URL host", but in Node.js they are the same:
> [Node docs](https://nodejs.org/api/http.html#httprequesturl-options-callback) - "Alias for host. To support url.parse(), **hostname will be used if both host and hostname are specified**."

URL host = URL hostname + ":" + URL port

### hostname
hostname = RequestOptions.hostname || RequestOptions.host || 'localhost' ([source](https://github.com/nodejs/node/blob/main/lib/_http_client.js#L190))

### port

`RequestOptions` host can not contain `port`. We shouldn't (can't) extract the `port` from the `host`

port = options.port || defaultPort || agent.defaultPort ([source](https://github.com/nodejs/node/blob/main/lib/_http_client.js#L189))

### auth:
Authorization header || RequestOptions.auth ([source](https://github.com/nodejs/node/blob/main/lib/_http_client.js#L309))

### protocol
options.protocol || defaultProtocol ([source](https://github.com/nodejs/node/blob/main/lib/_http_client.js#L169))

-----

P.S: Assuming `options` always takes precedence over URL. Why do we figure out the `URL` from the `options` instead of the other way around?
This way, `NodeClientRequest` would take only `options`  and `callback`, and `getUrlByRequestOptions` would be redundant because we pass the options almost as-is.

@kettanaito WDYT?

kettanaito:
@mikicho, thanks for looking into that! As long as the unit tests pass we can rewrite the options parsing logic but I would be extremely careful when considering changing tests. There's a convoluted edge case behind every test case and I'd rather we keep tests as-is, even if it seems like they make no sense (there are all sorts of usages in real world, some even incorrect according to Node.js docs but they are there and we must handle them). 

> Why do we figure out the URL from the options instead of the other way around?

Because there are third-party packages that use `ClientRequest` in every imaginable way. I've seen scenarios with partial URL _object_ + `RequestOptions` that compliment each other. In those scenarios, neither can be taken as the complete source of truth as Node itself merges the two to produce the complete request options. That's why we try to take URL as a primary input (the first argument) but still check if there are any options that'd modify it.

kettanaito:
If we can reproduce the `ENETUNREACH` error with a non-IPv6 address, then I highly recommend we extract the actual fix from this pull request and open a new one designated for that. I hope we don't have to refactor the options parsing logic to have that network error suppressed. 

We can keep the current pull request aimed at improving IPv6 support specifically and any related refactoring to options parsing.

mikicho:
@kettanaito 
I agree. 
IIUC, the test now should be in the `NodeClientRequest.test.ts`, so we don't need to bother with the URL options now.

The commit I overridden was: `f0ada9663ce02c27d98d50f3ad20bd2ce1f52dac`

kettanaito:
@mikicho, simply fantastic! Giving this a quick review and will merge if there's no comments from my side.

kettanaito (approved):
🚀

kettanaito:
## Released: v0.25.7 🎉

---

PR #463, opened 2023-09-30
fix(ClientRequest): preserve headers casing on mocked responses

### LINKED ISSUE #448: How to return the rawHeaders?
The `respondWith` function expects `Response` object that (AFAIK) doesn't support `rawHeaders`, but it seems like it's possible to test `rawHeaders` [in the code](https://github.com/mswjs/interceptors/blob/3ae4b4ccd2cc5d7bc6248238cc41bac3bf427e22/src/interceptors/ClientRequest/NodeClientRequest.ts#L458)

image

mikicho:
There are a few typing issues, but I would like to hear what you think about this solution first.
Do you know how can we find the Symbol more elegantly? I'm not very familiar with Symbols and `Symbol.for('headers list')` returns `undefined`

closes #448 

----

Some thoughts:

1. While this solution probably will work for good, it relies on an "unofficial" behavior of Node.
2. Perhaps we could add a function that allows users to set the 'rawHeaders', although it may be a bit clumsy.
3. We can also solve it from Nock's side by sending a tempered version of the `Response` object which does not return lowercase headers. but IMO the solution should be in this lib.

### DIFF
--- a/src/interceptors/ClientRequest/NodeClientRequest.test.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.test.ts
@@ -315,36 +315,3 @@ it('does not send request body to the original server given mocked response', as
   const text = await getIncomingMessageBody(response)
   expect(text).toBe('mock created!')
 })
-
-it('preserves the original mocked response headers casing in "rawHeaders"', async () => {
-  const emitter = new Emitter<HttpRequestEventMap>()
-  const request = new NodeClientRequest(
-    normalizeClientRequestArgs('http:', 'http://any.thing'),
-    {
-      emitter,
-      logger,
-    }
-  )
-
-  emitter.on('request', ({ request }) => {
-    request.respondWith(
-      new Response(null, {
-        headers: {
-          'X-CustoM-HeadeR': 'Yes',
-        },
-      })
-    )
-  })
-
-  request.end()
-
-  const responseReceived = new DeferredPromise<IncomingMessage>()
-
-  request.on('response', async (response) => {
-    responseReceived.resolve(response)
-  })
-  const response = await responseReceived
-
-  expect(response.rawHeaders).toStrictEqual(['X-CustoM-HeadeR', 'Yes'])
-  expect(response.headers).toStrictEqual({ 'x-custom-header': 'Yes' })
-})
--- a/src/interceptors/ClientRequest/NodeClientRequest.test.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.test.ts
@@ -315,3 +315,36 @@ it('does not send request body to the original server given mocked response', as
   const text = await getIncomingMessageBody(response)
   expect(text).toBe('mock created!')
 })
+
+it.only('sets the correct rawHeaders', async () => {
+  const emitter = new Emitter<HttpRequestEventMap>()
+  const request = new NodeClientRequest(
+    normalizeClientRequestArgs('http:', 'http://any.thing'),
+    {
+      emitter,
+      logger,
+    }
+  )
+
+  emitter.on('request', ({ request }) => {
+    request.respondWith(
+      new Response(null, {
+        headers: {
+          'X-Custom-Header': 'Yes',
+        },
+      })
+    )
+  })
+
+  request.end()
+
+  const responseReceived = new DeferredPromise<IncomingMessage>()
+
+  request.on('response', async (response) => {
+    responseReceived.resolve(response)
+  })
+  const response = await responseReceived
+
+  expect(response.rawHeaders).toStrictEqual([ 'X-Custom-Header', 'Yes' ])
+  expect(response.headers).toStrictEqual({ 'x-custom-header': 'Yes' })
+})
--- a/src/interceptors/ClientRequest/NodeClientRequest.test.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.test.ts
@@ -316,7 +316,7 @@ it('does not send request body to the original server given mocked response', as
   expect(text).toBe('mock created!')
 })
 
-it.only('does not lowercase the rawHeaders', async () => {
+it('does not lowercase the rawHeaders', async () => {
   const emitter = new Emitter<HttpRequestEventMap>()
   const request = new NodeClientRequest(
     normalizeClientRequestArgs('http:', 'http://any.thing'),
--- a/src/interceptors/ClientRequest/NodeClientRequest.test.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.test.ts
@@ -316,7 +316,7 @@ it('does not send request body to the original server given mocked response', as
   expect(text).toBe('mock created!')
 })
 
-it('does not lowercase the rawHeaders', async () => {
+it('preserves the original mocked response headers casing in "rawHeaders"', async () => {
   const emitter = new Emitter<HttpRequestEventMap>()
   const request = new NodeClientRequest(
     normalizeClientRequestArgs('http:', 'http://any.thing'),
@@ -330,7 +330,7 @@ it('does not lowercase the rawHeaders', async () => {
     request.respondWith(
       new Response(null, {
         headers: {
-          'X-Custom-Header': 'Yes',
+          'X-CustoM-HeadeR': 'Yes',
         },
       })
     )
@@ -345,6 +345,6 @@ it('does not lowercase the rawHeaders', async () => {
   })
   const response = await responseReceived
 
-  expect(response.rawHeaders).toStrictEqual([ 'X-Custom-Header', 'Yes' ])
+  expect(response.rawHeaders).toStrictEqual(['X-CustoM-HeadeR', 'Yes'])
   expect(response.headers).toStrictEqual({ 'x-custom-header': 'Yes' })
 })
--- a/src/interceptors/ClientRequest/NodeClientRequest.test.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.test.ts
@@ -316,7 +316,7 @@ it('does not send request body to the original server given mocked response', as
   expect(text).toBe('mock created!')
 })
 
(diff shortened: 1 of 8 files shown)


src/interceptors/ClientRequest/NodeClientRequest.test.ts
```diff
@@ -315,3 +315,36 @@ it('does not send request body to the original server given mocked response', as
   const text = await getIncomingMessageBody(response)
   expect(text).toBe('mock created!')
 })
+
+it.only('does not lowercase the rawHeaders', async () => {
```
kettanaito, line 319: This test should be unmarked as `.only`.

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -20,6 +20,10 @@ import { toInteractiveRequest } from '../../utils/toInteractiveRequest'
 import { uuidv4 } from '../../utils/uuid'
 import { emitAsync } from '../../utils/emitAsync'
 
+const kHeadersList = Object.getOwnPropertySymbols(new Headers()).find(
```
kettanaito, line 23: I recommend highly against tapping into Node's internal symbols. They are tedious to work with intentionally—you shouldn't access them unless you own them. They can be swapped at any point of time in Node, and more so across different versions, making them a big maintenance issue.

kettanaito (changes requested):
While this approach may work, I recommend against it (see my reasoning in the comments). 

Instead, I propose we find some middle-ground in exposing the original response to the consumer (the response listener). Historically, Interceptors used to expose the original request instance in the request listener, which looked like this:

```ts
type RequestListener = (
  request: Request,
  ref: ClientRequest | XMLHttpRequest
) => void
```

This was to give the consumer the means to access and operate on that original instance, need be. I've then moved away from this call signature, finding it too dynamic and confusing for the consumer. 

But I think we can bring it back, although partially. I dislike that the `ref` was a completely different instance based on the request-issuing module. I think Interceptors must expose requests/responses in a unified way. I think the next unified way to represent requests/responses is HTTP message. 

What do you say if we create and expose an HTTP message based on the request/response in the respective listeners?

```js
interceptor.on('response', ({ response, httpMessage }) => {
  httpMessage.headers // raw headers
})
```

The `httpMessage` will be a custom abstraction respecting the HTTP message specification and describing the request/response as it was sent/received. This would be your primary way to access raw headers, as an example. 

I'm currently working on something very closely related, and I can abstract the `HttpMessage` class into a separate package and we can consume it in Interceptors then. 

Having access to the HTTP message also provides more data to the consumer, as they are now able to do thing like calculating the headers/body size.

mikicho:
> I dislike that the ref was a completely different instance

Agree!

> This would be your primary way to access raw headers, as an example.

Do you mean that the `response` hook no longer would be read-only?
If the response remains read-only, I believe that including this in the `request` hook would be more appropriate. This is because we must ensure that the headers are set correctly in the response instance that the user receives in the callback.

kettanaito:
The `response` event remains read-only. What I meant is to expose _another_ property in the listener to that event called `httpMessage`, which will be an abstraction over the originally sent HTTP message. That way the consumer can operate with the response using its Fetch API `Response` representation as well as the "raw" HTTP message representation (your use case, to access the raw response headers). Neither allow modifying the response.

kettanaito:
@mikicho, no because the HTTP message will be created from the original `IncomingMessage` _not_ its fetch representation. That way, the HTTP message will contain the raw headers as they came, which seems to be the expected behavior here. That way we have a more low-level representation of responses (and even requests, if we want to) without exposing a variadic request instance reference, which would be different for each request client.

mikicho:
What about mocked responses?
I think I'm missing something.

The current flow is:
1. someone calls the req.end() function
2. Call to all `request` event listeners
3. Check if someone called to `responseWith` in one of the listeners
4. If yes, use the mocked response which is a `Response` instance with `lowercased` header names.

What do I miss?

kettanaito:
@mikicho, since mock responses are declared using the Fetch API `Response` class, there is no concept of `raw` headers there. The response headers will be available by their raw name and by their normalized name (lowercase). Generally, response headers will be case-insensitive. 

This will make your test pass in both cases:

```js
// Assert normalized response headers.
const normalizedHeaders = Object.fromEntries(response.headers.entries())
expect(normalizedHeaders).toEqual(...)

// Assert raw response headers.
expect(response.headers.get('X-Custom-Header')).toBe('Expected Value')
```

But I think I understand the issue here. Since it's `Response` -> `IncomingMessage` translation, you also want to make sure that `ClientRequest` receives the `IncomingMessage` object with the right headers, which _are_ case-sensitive (speaking of `rawHeaders`). 

I will think about it and reply on this tomorrow.

kettanaito:
Apologies for not understanding the requirements here sooner. I think I've finally wrapped my head around them. Yes, I think your original proposal to tap into `Headers`'s symbols may be our best bet here given the constraints of `Response`. 

Let me show you how we can access that symbol a bit better...

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -448,10 +449,14 @@ export class NodeClientRequest extends ClientRequest {
     this.response.statusCode = status
     this.response.statusMessage = statusText
 
-    if (headers) {
+    // Try extracting the raw headers from the headers instance.
+    // If not possible, fallback to the headers instance as-is.
+    const rawHeaders = getRawFetchHeaders(headers) || headers
```
kettanaito, line 454: We will try to read the raw headers from the internal `Headers` symbols but if those are not present for any reason, fallback to the `headers` instance as-is.

src/utils/getRawFetchHeaders.ts
```diff
@@ -0,0 +1,38 @@
...
+): RawHeadersMap | undefined {
+  const headersList = getValueBySymbol<object>('headers list', headers)
+
+  if (!headersList) {
+    return
+  }
+
+  const headersMap = getValueBySymbol<
+    Map<string, { name: string; value: string }>
+  >('headers map', headersList)
```
kettanaito, line 25: I've noticed the raw headers map is buried a bit deeper in the symbols:

```
Headers
 - Symbol(headers list)
   - Symbol(headers map)
```

src/utils/getValueBySymbol.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Returns the value behind the symbol with the given name.
+ */
+export function getValueBySymbol<T>(
+  symbolName: string,
+  source: object
+): T | undefined {
+  const ownSymbols = Object.getOwnPropertySymbols(source)
+
+  const symbol = ownSymbols.find((symbol) => {
+    return symbol.description === symbolName
```
kettanaito, line 11: I took your original lookup logic just wrapped it in a utility function.

test/modules/http/compliance/http-res-raw-headers.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+beforeAll(() => {
+  interceptor.apply()
+})
+
+afterAll(() => {
+  interceptor.dispose()
+})
+
+it('preserves the original mocked response headers casing in "rawHeaders"', async () => {
```
kettanaito, line 16: I've also moved the raw headers test from `NodeClientRequest.test.ts` to a new compliance test. I think it makes sense to be here.

src/utils/getRawFetchHeaders.ts
kettanaito: This seems to be specific to a Node.js version. This is where the raw headers map is for me on 18.14.0 but, apparently, it wasn't there on 18.8.0 used in the CI. I've updated the Node version on CI, let's see how it fares.

kettanaito: I confirm that's due to the Node.js difference in versions:

- On 18.14.0, `headers map` is `Map<normalizedName, { name: rawName, value: string }>`
- On 18.8.0, the same `headers map` symbol is `Map<normalizedName, value>`. Looks like older versions of Node doesn't even expose the raw header name in any way.

mikicho: They do in the `Symbol(headers list).entires` (not the function `entries`)

kettanaito: I fixed this by checking what values `headersMap` actually has. If it's not an object with a `name` property, return undefined.

kettanaito: @mikicho, unfortunately, `.entries()` on `headers list` contains already normalized header names:

```js
const h = new Headers([['aBc, 'value']])
getValueBySymbol('headers list', h).entries()
// [ ['abc', 'value'] ]
```

kettanaito (approved):
This looks good from my side now. @mikicho, let me know if we should merge this.

mikicho:
@kettanaito LGTM! Thanks! ❤️ 
I still need to verify the support limit with Nock's maintainer(s?), but I think I can make a good workaround until 18.14 is widely used and even deprecated.

Just a quick note, I see the value of sticking with a common and familiar object like `Response`, but it can also restrict us in unforeseen ways. It might be beneficial to consider a solution that can help us avoid these limitations moving forward.

kettanaito:
> It might be beneficial to consider a solution that can help us avoid these limitations moving forward.

You may be right. But the way I see it—the Fetch API is extremely powerful and is the future of JavaScript. It's limitations are imposed only by our familiarity with, frankly, archaic ways of dealing with network requests in Node.js (looking at you, `http.IncomingMessage`). Things like header sensitivity have little practice use since most specs declare headers as case-insensitive. I'd even go as far as to say that writing tests for `rawHeaders` is testing Node's implementation details but I understand that the line between those is rather blurred (has been blurred for years when using Node).

kettanaito:
## Released: v0.25.7 🎉

---

PR #464, opened 2023-10-01
fix(ClientRequest): immediate abort

### LINKED ISSUE #444: Should we run interceptor handler for destroyed(/aborted) requests?
In Nock, we have this test:

```js
  it.only('Emits the expected event sequence when `end` is called on an aborted request', done => {
    const scope = nock('http://example.test').get('/').reply()

    const req = http.request('http://example.test')
    req.abort()
    req.end()

    setTimeout(() => {
      expect(emitSpy).to.have.been.calledTwice
      expect(emitSpy.firstCall).to.have.been.calledWith('close')
      expect(emitSpy.secondCall).to.have.been.calledWith('abort')
      expect(scope.isDone()).to.be.false()
      done()
    }, 10)
  })
```

This test fails on this line:
```js
expect(scope.isDone()).to.be.false()
```
because we handle the `end` event as usual even for aborted requests.

mikicho:
I added failing tests.

1. Should emit close and then abort events.
2. Should not call to the request listeners if defined.

Currently, when we do return a mocked response, we get the following error:

> TypeError: Cannot set property writableFinished of #<OutgoingMessage> which has only a getter

if merged, should fix #444  

@kettanaito Let me know WDYT about the desired behavior, and any guidance is welcome. :)

### DIFF
--- a/src/interceptors/ClientRequest/NodeClientRequest.test.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.test.ts
@@ -1,5 +1,6 @@
 import { vi, it, expect, beforeAll, afterAll } from 'vitest'
 import express from 'express'
+import { setImmediate } from 'node:timers/promises';
 import { IncomingMessage } from 'http'
 import { Emitter } from 'strict-event-emitter'
 import { Logger } from '@open-draft/logger'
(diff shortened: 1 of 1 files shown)


kettanaito:
I've shared some thoughts on this in [here](https://github.com/mswjs/interceptors/issues/444#issuecomment-1744616122). Would like to know what you think.

PR #468, opened 2023-10-07
fix(ClientRequest): support "http.request()" and "http.get()" without any arguments

### LINKED ISSUE #461: Handle http.request() with no options
From nock's [source code](https://github.com/nock/nock/blob/feaa66fa64d24f95937ef759cdd5a7ca07646f1c/lib/intercept.js#L383-L390):
>    // As weird as it is, it's possible to call `http.request` without
      // options, and it makes a request to localhost or somesuch. We should
      // support it too, for parity. However it doesn't work today, and fixing
      // it seems low priority. Giving an explicit error is nicer than
      // crashing with a weird stack trace. `new ClientRequest()`, nock's
      // other client-facing entry point, makes a similar check.
      // https://github.com/nock/nock/pull/1386
      // https://github.com/nock/nock/pull/1440

I think we should follow and just throw error. currently we throw:
> Cannot use 'in' operator to search for 'hash' in undefined

kettanaito:
- Fixes #461

### DIFF
--- a/src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.ts
+++ b/src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.ts
@@ -23,6 +23,8 @@ const logger = new Logger('http normalizeClientRequestArgs')
 export type HttpRequestCallback = (response: IncomingMessage) => void
 
 export type ClientRequestArgs =
+  // Request without any arguments is also possible.
+  | []
   | [string | URL | LegacyURL, HttpRequestCallback?]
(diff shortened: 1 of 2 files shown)


kettanaito:
Hey, @mikicho. I'm adding support for `http.request()` without any arguments as per your suggestion. I've also spotted that `http.get()` can be called without any arguments too. We use the same normalization function for both, so both will be supported now.

kettanaito:
## Released: v0.25.7 🎉

---

Issue #470, opened 2023-10-10
Replace "uuid" with "crypto.randomUUID()"


kettanaito:
Node.js ships with the built-in way to generate UUID via the `node:crypto` module. We should remove the `uuid` dependency and use `node:crypto` instead.

hpohlmeyer:
Since #470 is closed now, I am adding the comment here as well:

Using `node:crypto` would break browser compatibility. As far as I know msw uses this package to defer requests until the service worker has been registered, so browser compatibility is still needed.

Using `crypto.randomUUID()` work  on https-sites, but cannot be used on http-sites, because `crypto` is `undefined` there.

kettanaito:
@hpohlmeyer, thanks for the context around the pull request. 

> Using node:crypto would break browser compatibility.

True. However, both Node.js and browsers ships a global `crypto` object that has the `.randomUIUD()` method. This code runs the same in the browser and in Node.js:

```js
crypto.randomUUID()
```

We don't need to import `node:crypto` at all! 

> Using crypto.randomUUID() work on https-sites, but cannot be used on http-sites, because crypto is undefined there.

This is out of the scope of this package. None of the interceptors are meant to be deployed to production. I don't treat this as a blocker.

kettanaito:
## Released: v0.26.12 🎉

This has been released in v0.26.12!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.26.12)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.26.12)

Make sure to always update to the latest version (`npm i @mswjs/interceptors@latest`) to get the newest features and bug fixes.

---

_Predictable release automation by [@ossjs/release](https://github.com/ossjs/release)_.

PR #471, opened 2023-10-12
feat: replaced "uuid" with "crypto.randomUUID()" (#470)

biplobsd:
I have deleted all uuid and used crypto.randomUUID(). I also deleted uuid file for is being unused function.

Fixed : #470

### DIFF
--- a/src/interceptors/ClientRequest/NodeClientRequest.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.ts
@@ -19,6 +19,7 @@ import { createRequest } from './utils/createRequest'
 import { toInteractiveRequest } from '../../utils/toInteractiveRequest'
 import { emitAsync } from '../../utils/emitAsync'
 import { getRawFetchHeaders } from '../../utils/getRawFetchHeaders'
+import crypto from "node:crypto"
 
 export type Protocol = 'http' | 'https'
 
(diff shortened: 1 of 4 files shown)


src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -150,7 +149,7 @@ export class NodeClientRequest extends ClientRequest {
   end(...args: any): this {
     this.logger.info('end', args)
 
-    const requestId = uuidv4()
+    const requestId = crypto.randomUUID()
```
kettanaito, line 152: Can you please import `crypto` from `node:crypto`? Afaik, it's not a universally supported global yet.

biplobsd: Ok.

biplobsd:
Hello @kettanaito, is there any changes needed? 😐

hpohlmeyer:
Just saw this PR and wanted to note, that this makes @mswjs/interceptors incompatible with browsers. If I remember correctly, msw uses the interceptors library to defer request execution until the service worker has been registered.

`node:crypto` is not available in browsers, so this would break compatibility. But even using `crypto.randomUUID()` without the import breaks compatibility, if it is used on a http site, because `crypto` is only available on https sites.

biplobsd:
> Just saw this PR and wanted to note, that this makes @mswjs/interceptors incompatible with browsers. If I remember correctly, msw uses the interceptors library to defer request execution until the service worker has been registered.
> 
> `node:crypto` is not available in browsers, so this would break compatibility. But even using `crypto.randomUUID()` without the import breaks compatibility, if it is used on a http site, because `crypto` is only available on https sites.

Okay, I'm close this PR. @hpohlmeyer . This changes make more problem 😄

PR #472, opened 2023-10-13
fix(XMLHttpRequest): prevent invalid listener name access

hpohlmeyer:
When a XMLHttpRequest event handler is reset by setting it to `null`, an error will be thrown, that @mswjs/interceptors cannot access the property `name` on `null`. Here is an example of that error:

```ts
req.onload = function() { console.log(`Responded with status %i`, this.status) }
req.ontimeout = null
req.open("GET", "http://localhost:7003/test/status/200")
req.send()
```

This PR prevents the error by not accessing the `name` property if the listener is `undefined | null`;

Do you think it needs tests, or is it fine as-is?

### DIFF
--- a/src/Interceptor.ts
+++ b/src/Interceptor.ts
@@ -149,7 +149,7 @@ export class Interceptor<Events extends InterceptorEventMap> {
       return this
     }
 
-    logger.info('adding "%s" event listener:', event, listener?.name)
+    logger.info('adding "%s" event listener:', event, listener)
 
     this.emitter.on(event, listener)
     return this
--- a/src/Interceptor.ts
+++ b/src/Interceptor.ts
@@ -149,7 +149,7 @@ export class Interceptor<Events extends InterceptorEventMap> {
       return this
     }
 
-    logger.info('adding "%s" event listener:', event, listener.name)
+    logger.info('adding "%s" event listener:', event, listener?.name)
 
     this.emitter.on(event, listener)
     return this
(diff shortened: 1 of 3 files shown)


kettanaito:
This is a great find, @hpohlmeyer! Thank you for fixing this. 

Yes, adding an integration test that does exactly what you do in that code snippet would be essential. Would you have a chance to add it, please? You can put it in `test/modules/XMLHttpRequest/compliance/xhr-event-callback-null`. A better name for the test module is also welcome! 

Let me know if you need any help writing that test.

hpohlmeyer:
@kettanaito sorry for not responding sooner. I am on vacation right now. I will be back next week. 

I did write some tests already, but was not sure where to put it. Thanks for pointing me to the right file. I will probably be able to add the tests on Monday.

kettanaito:
@hpohlmeyer, I think it's enough to put your usage scenario from the description and assert that the request doesn't error. We shouldn't assert the `name` or anything since those are implementation details.

hpohlmeyer:
Okay, I have added some tests and I am just logging the event listener now.

kettanaito (approved):
This looks fantastic! Thank you for your work on this, @hpohlmeyer. 🚀

kettanaito:
## Released: v0.25.8 🎉

This has been released in v0.25.8!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.25.8)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.25.8)

---

PR #477, opened 2023-11-07
fix(xhr): allow setters for non-existing properties

kettanaito:
- Fixes https://github.com/mswjs/msw/issues/1816

## Investigation

This issue is caused by the Axios XHR adapter in Happy DOM only. Axios does:

```js
request.timeout = config.timeout
```

Where `request` is already a proxy established by Interceptors. It looks up the `setProperty` trap on the proxy and since `timeout` doesn't have a special case, it defaults to `invoke()`. 

Calling `invoke()` to set `timeout` on XMLHttpRequest proxy _returns false_ because property `timeout` does not exist on `target`:

https://github.com/mswjs/interceptors/blob/1f04d100bad190a393a5855fc6cc3a54f6117dfb/src/utils/findPropertySource.ts#L9-L11

The `timeout` property is not mandatory on XMLHttpRequest, so it's missing from the immediate XHR instance as well as its prototype. We return false from the setter and that causes the following error:

```
TypeError: 'set' on proxy: trap returned falsish for property 'timeout'
```

## Solution

I think us forbidding setters on unknown properties is a mistake. To fix this, I'm adding a fallback to `propertySource` to be the immediate `target` itself is `findPropertySource` returned `null`.

### DIFF
--- a/src/utils/createProxy.ts
+++ b/src/utils/createProxy.ts
@@ -44,21 +44,21 @@ function optionsToProxyHandler<T extends Record<string, any>>(
     }
   }
 
-  handler.set = function (target, propertyName, nextValue, receiver) {
+  handler.set = function (target, propertyName, nextValue) {
     const next = () => {
-      const propertySource = findPropertySource(target, propertyName)
-      if (propertySource === null) return false
-
+      const propertySource = findPropertySource(target, propertyName) || target
       const ownDescriptors = Reflect.getOwnPropertyDescriptor(
         propertySource,
         propertyName
       )
 
+      // Respect any custom setters present for this property.
       if (typeof ownDescriptors?.set !== 'undefined') {
         ownDescriptors.set.apply(target, [nextValue])
         return true
       }
 
+      // Otherwise, set the property on the source.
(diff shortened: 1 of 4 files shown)


src/utils/createProxy.ts
```diff
@@ -44,21 +44,21 @@ function optionsToProxyHandler<T extends Record<string, any>>(
     }
   }
 
-  handler.set = function (target, propertyName, nextValue, receiver) {
+  handler.set = function (target, propertyName, nextValue) {
     const next = () => {
-      const propertySource = findPropertySource(target, propertyName)
-      if (propertySource === null) return false
-
+      const propertySource = findPropertySource(target, propertyName) || target
```
kettanaito, line 49: The actual fix: `target` is now considered a fallback for `propertySource` if setting properties that don't exist on `target` or its prototype.

kettanaito:
## Released: v0.25.9 🎉

This has been released in v0.25.9!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.25.9)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.25.9)

---

PR #479, opened 2023-11-08
fix(ClientRequest, Fetch): support miniflare by checking "Response.type" access

### LINKED ISSUE #478: `Response.type` throws in miniflare-like environments
https://github.com/mswjs/interceptors/blob/3258f5cc62c472ba63f5c70e455d5b245c74a216/src/interceptors/fetch/index.ts#L106

In environments like miniflare, `Request.type` is defined as a getter that throws. 

We should write some tests to ensure that interceptors can work in miniflare (jest/vitest environment miniflare should support us here) and update to check something else or ignore this check entirely if the property access throws an error (like miniflare's does)

related to https://github.com/mswjs/msw/issues/1834

mattcosta7:
- Fixes #478

quick tests/updates to support usage in miniflare/cloudflare like environmetns. 

I'm not super familiar here, but this should fix and add some regression tests for miniflare environments.

The issue here was that `Response.type` is not implemented in miniflare, and accessing that property throws, so we need to guard against that and validate it can be accessed before attempting to access it. 

Do we have a different path to validating `Response.type === 'error'`?  would Response.status = 0 suffice, or something similar?   If we do we may need additional error tests here, but checking in progress for now

`Response.error()` is also not defined in miniflare, so I don't think we can create a Response with a type of 'error' or similar, outside of status code checks?

relates to https://github.com/mswjs/msw/issues/1834


test/third-party/miniflare.test.ts
```diff
@@ -0,0 +1,54 @@
...
+import { ClientRequestInterceptor } from '../../src/interceptors/ClientRequest'
+import { XMLHttpRequestInterceptor } from '../../src/interceptors/XMLHttpRequest'
+import { FetchInterceptor } from '../../src/interceptors/fetch'
+import { httpGet, httpsGet } from '../helpers'
+
+const interceptor = new BatchInterceptor({
+  name: 'setup-server',
+  interceptors: [
+    new ClientRequestInterceptor(),
+    new XMLHttpRequestInterceptor(),
```
mattcosta7, line 13: XMLHttp is not defined in miniflare, but `setupServer` in msw defaults to applying it.  it's a noop, but I've kept that here, even though it's unused - as a regression check that we can apply it

We may want to look into scoping an interceptor set for that environment?

test/third-party/miniflare.test.ts
```diff
@@ -0,0 +1,54 @@
+// @vitest-environment miniflare
```
mattcosta7, line 1: let me know if this test placement makes sense in `third-party`.  It's mostly an integration test for miniflare compat, which seems reasonable here

kettanaito: Yes, this is the perfect place to put this. Thanks!

kettanaito:
> The issue here was that Response.type is not implemented in miniflare, and accessing that property throws, so we need to guard against that and validate it can be accessed before attempting to access it.

This begins to sound like an overly harsh Not Implemented behavior in Miniflare forces us into uncertainty-driven engineering here. I see no reason not to implement the `"type"` property on `Response` but this isn't my area to reason about. 

> Do we have a different path to validating Response.type === 'error'? 

I wouldn't adjust a perfectly valid implementation to comply with environments that don't implement the Fetch API specification correctly. I don't mind a compromise of accessing the `type` property a bit more cautiously, but I think that's where our responsibility ends. 

> Response.error() is also not defined in miniflare, so I don't think we can create a Response with a type of 'error' or similar, outside of status code checks?

Then it looks like you cannot emulate network errors in Miniflare. This should be clearly documented on Miniflare's side so their users are aware of it.

test/third-party/miniflare.test.ts
```diff
@@ -0,0 +1,71 @@
...
+test('responds to https get', async () => {
+  const { resBody } = await httpsGet('https://example.com')
+  expect(resBody).toEqual('mocked-body')
+  expect(requestListener).toHaveBeenCalledTimes(1)
+})
+
+test('Some response properties/methods throw', async () => {
+  // If this test fails in the future, we can remove the property access
+  // safety checks in the interceptors.
+  expect(() => new Response('body').type).toThrow()
```
kettanaito, line 59: I don't think this is something we should be testing. These tests don't even involve Interceptors so they assert the behavior of a third-party (`vitest-environment-miniflare` in this case). This is incorrect. 

Instead, if we wish to assert what happens on `Response.type` access, create a request listener that responds with that call. It will throw, and we will receive a request error, I believe. Assert _that_.

test/third-party/miniflare.test.ts
```diff
@@ -0,0 +1,71 @@
...
+})
+
+/**
+ * We aren't testing this on purpose, even though we install the interceptor here.
+ * The reason is that we want to make sure we can install the XMLHttpRequest interceptor
+ * even when the global XMLHttpRequest is not available. So that we can avoid
+ * issues with miniflare.
+ */
+test('XMLHttpRequest is not available', () => {
+  expect(() => XMLHttpRequest).toThrow()
```
kettanaito, line 70: The same comment here: the fact that `XMLHttpRequest` is unavailable in Miniflare isn't related in any way to Interceptors. We should drop this test.

src/interceptors/fetch/index.ts
```diff
@@ -103,7 +104,10 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
         this.logger.info('received mocked response:', mockedResponse)
 
         // Reject the request Promise on mocked "Response.error" responses.
-        if (mockedResponse.type === 'error') {
+        if (
+          isPropertyAccessible(mockedResponse, 'type') &&
```
kettanaito, line 108: Can you please add a comment block that explains why we check this so, and include a link to the issue/PR? Thanks.

kettanaito:
@mattcosta7, thanks for working on this. I think I understand the issue with Miniflare better now. 

I've restructured the tests, moved the XHR test into its own thing so we could assert `interceptor.apply()` itself, added a comment regarding the `Response.type` access and also a new test for mocking network errors with `Response.error()`. Everything is passing now. 

Anything missing here? Should we merge once it's green?

kettanaito (approved):

kettanaito:
I will go and merge this, this looks self-contained. We can always iterate on this later.

kettanaito:
## Released: v0.25.10 🎉

This has been released in v0.25.10!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.25.10)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.25.10)

---

PR #487, opened 2023-11-24
Fix response raw headers

mikicho:
The `Response` merges the headers, so we don't get the original rawHeaders array.
image

@kettanaito WDYT?

P.S: I fixed a small error in headers, set-cookie is not an array. (easy fix)


mikicho:
After I thought about it, I think we should add an extra argument or a way to pass the raw headers more straightforwardly.
I think that the `rawHeaders` case is an edge case; most users don't care about them very much and use the `headers` object.
So, instead of utilizing an internal implementation of Node.js, which may break at some point without notice, we should expose a proper API for this.

I don't think we should "dirty" the `withResponse` function for this edge case, but maybe add another function or another solution.

kettanaito:
Closing in favor of #598.

PR #488, opened 2023-11-27
fix: export a "isResponseWithoutBody" utility

kettanaito:
- Related to https://github.com/mswjs/msw/pull/1894

### DIFF
--- a/src/utils/responseUtils.ts
+++ b/src/utils/responseUtils.ts
@@ -2,12 +2,14 @@
  * Response status codes for responses that cannot have body.
  * @see https://fetch.spec.whatwg.org/#statuses
  */
-export const RESPONSE_STATUS_CODES_WITHOUT_BODY = [101, 103, 204, 205, 304]
+export const RESPONSE_STATUS_CODES_WITHOUT_BODY = new Set([
+  101, 103, 204, 205, 304,
+])
 
 /**
  * Returns a boolean indicating whether the given response status
  * code represents a response that cannot have a body.
  */
 export function isResponseWithoutBody(status: number): boolean {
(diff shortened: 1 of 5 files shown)


src/utils/responseUtils.ts
```diff
@@ -2,4 +2,12 @@
  * Response status codes for responses that cannot have body.
  * @see https://fetch.spec.whatwg.org/#statuses
  */
-export const responseStatusCodesWithoutBody = [204, 205, 304]
+export const RESPONSE_STATUS_CODES_WITHOUT_BODY = [101, 103, 204, 205, 304]
```
mattcosta7, line 5: super duper nitty:

Should we make this a set instead of an array? super minimal but a bit more semantic

mattcosta7 (approved):

src/utils/responseUtils.ts
mattcosta7: CI seems to suggest we throw on 101/103, which I think makes sense.

101 is switching protocols (changing connection to http/2 or sockets)
103 is early hints

I'm not sure how interceptors handle those now if we received them from somewhere. If we ignore them  or what, but maybe we only need to remove them from here as they're not standard request/response style headers that would happen here?

kettanaito: @mattcosta7, made the status codes a set. Adjusted the test.

kettanaito:
## Released: v0.25.13 🎉

This has been released in v0.25.13!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.25.13)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.25.13)

---

PR #494, opened 2023-12-22
Package exports should only contain valid environments

jd-carroll:
Fix package exports issue

See: [msw/#1809](https://github.com/mswjs/msw/issues/1809)


jd-carroll:
This is outdated, if updated patch works will submit new PR

PR #484, opened 2023-11-11
feat: intercept WebSocket in the browser

kettanaito:
A refreshed look at the WebSocket interception originally drafted in #236. 

## Roadmap

- [ ] Test the current behavior completely before moving to the next WebSocket transport. 
- [ ] Implement `connection.close(code)` to mock server connection close.
- [ ] Implement `connection.error(error)` to mock server connection errors.

PR #501, opened 2024-01-28
feat: add WebSocket class interceptor

kettanaito:
- Closes #236 
- Closes #484 
- **See this branch's `README.md` for the feature documentation.**

## Changes

- Adds a WebSocket interceptor based on the `globalThis.WebSocket` class. 
- Introduces the `client` and the `server` APIs to represent the incoming client connection and the original server connection respectively in the connection listener of the interceptor.
- Adds the required infrastructure of the interceptor/client/server/transport classes that describe the generic class and also the WebSocket class-relevant implementations of them, such as `WebSocketClassClient`, `WebSocketClassServer`, `WebSocketClassTransport`.

## Todo

- [x] **Tests only:** The `ws` module we use for testing doesn't pass the `instanceof `check on dispatched events in Node, causing some tests to fail (https://github.com/websockets/ws/issues/1818). 
  - [x] Once this is fixed, unskip the tests at `test/modules/WebSocket/WebSocketClass/exchange/websocket.server.connect.test.ts`
  - **This was caused by JSDOM**. It polyfills global `Event`. 
- [x] Write WebSocket compliance tests.
  - [x] Default state
  - [x] Events
  - [x] Methods (`.send()` and `.close()` and its input)
  - [x] Setters (`.onopen` and etc)
- [x] Some third-parties also [add `origin` on the `MessageEvent`](https://github.com/nodejs/undici/blob/f1d7ada7eb2292932065a7c9ab23569b8a6e1720/lib/websocket/util.js#L100) from the socket. Check if the interceptor should do that too (inspect the original WebSocket event).
- [x] Emit the `error` event correctly. It's emitted in the following scenarios:
  - [x] Connection was closed with a `> 1000 <= 1015` (https://www.rfc-editor.org/rfc/rfc6455#section-7.4.1)
  - [x] When sending data and being unable to buffer it anymore (the buffer is full, the connection is marked as full and closed). Such connection closures must "fire an event named error" (https://websockets.spec.whatwg.org/#dom-websocket-onerror).
    - I'll skip this; not sure we can implement it reliably.
- [x] Close the actual server connection whenever the original (mocked) client is closed.
- [x] Consider removing `WebSocketClient#emit`. It's non-standard. 
- [x] ~~Support custom encoder/decoder for (a) JSON transfer; (b) to support custom event format (like SocketIO)~~.
  - **This may be out of scope**. Consider this carefully. 
  - Use bindings instead. I've implemented the [socket.io-binding](https://github.com/mswjs/socket.io-binding).
- [x] Consider migrating to EventTarget-like public API: `client.addEventListener()`.

## Third-party support

- [x] SocketIO
  - Proposed to expose higher-level encoding/decoding utilities https://github.com/socketio/socket.io-parser/issues/129


src/interceptors/WebSocket/index.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export class WebSocketInterceptor extends BatchInterceptor<
+  [WebSocketClassInterceptor],
+  WebSocketEventsMap
+> {
+  constructor() {
+    super({
+      name: 'websocket',
+      interceptors: [new WebSocketClassInterceptor()],
```
kettanaito, line 28: The `WebSocketInterceptor` itself has started as a batched interceptor on a premise that in the future it will combine multiple, transport-related interceptors (WebSocket, XHR polling). 

I believe that's redundant now. I don't plan to support any non-standard WebSocket implementations, and this pull request implements the interceptor for the standard implementation already.

src/interceptors/WebSocket/implementations/WebSocketClass/WebSocketClassInterceptor.ts
```diff
@@ -0,0 +1,266 @@
...
+  static symbol = Symbol('websocket-class')
+
+  constructor() {
+    super(WebSocketClassInterceptor.symbol)
+  }
+
+  protected checkEnvironment(): boolean {
+    // Enable this interceptor in any environment
+    // that has a global WebSocket API.
+    return typeof globalThis.WebSocket !== 'undefined'
```
kettanaito, line 20: The WebSocket interceptor will be available anywhere where the global `WebSocket` class is present. 

We are setting that class in a custom Vitest environment, taking it from `undici` and making it a global class.

src/interceptors/WebSocket/implementations/WebSocketClass/WebSocketClassInterceptor.ts
```diff
@@ -0,0 +1,266 @@
...
+        const [url, protocols] = args
+
+        const createConnection = (): WebSocket => {
+          return Reflect.construct(target, args, newTarget)
+        }
+
+        // All WebSocket instances are mocked and don't forward
+        // any events to the original server (no connection established).
+        // To forward the events, the user must use the "server.send()" API.
+        const mockWs = new WebSocketClassOverride(url, protocols)
```
kettanaito, line 39: All constructed WebSocket instances will point to a "dummy" implementation of the WebSocket class (i.e. mock).

## Reason

`new WebSocket(url)` connections to non-existing addresses throw a special kind of error in the browser that cannot be caught by any means.

src/interceptors/WebSocket/WebSocketTransport.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export type WebSocketTransportOnIncomingCallback = (
+  event: MessageEvent<WebSocketSendData>
+) => void
+
+export type WebSocketTransportOnOutgoingCallback = (
+  data: WebSocketSendData
+) => void
+
+export abstract class WebSocketTransport {
```
kettanaito, line 15: Transport is an abstract class responsible for handling incoming and outgoing events from a WebSocket instance. 

At its core, the transport is the barebones implementation of how to apply `client.send()` events and how to listen to `client.on()` events.

Later, the transport is wrapped in the Connection instance to acquire a nicer public API.

src/interceptors/WebSocket/WebSocketClient.ts
```diff
@@ -0,0 +1,73 @@
...
+import { bindEvent } from './utils/bindEvent'
+
+const kEmitter = Symbol('kEmitter')
+
+/**
+ * The WebSocket client instance represents an incoming
+ * client connection. The user can control the connection,
+ * send and receive events.
+ */
+export class WebSocketClient {
```
kettanaito, line 22: This is a WebSocket client connection class. It represents a single WebSocket client connected to the "server" (the interceptor). 

This API is implementation-agnostic and provides the public methods to interact with the intercepted WebSocket client.

src/interceptors/WebSocket/WebSocketServer.ts
```diff
@@ -0,0 +1,33 @@
+import type { WebSocketSendData } from './WebSocketTransport'
+import type { WebSocketMessageListener } from './implementations/WebSocketClass/WebSocketClassInterceptor'
+
+/**
+ * The WebSocket server instance represents the actual production
+ * WebSocket server connection. It's idle by default but you can
+ * establish it by calling `server.connect()`.
+ */
+export class WebSocketServer {
```
kettanaito, line 9: Similar to the WebSocket client representation, this is the actual WebSocket server connection representation. It's used to:

- Establish the original connection to the server;
- Intercept the incoming server events (the real ones);
- Forward any (intercepted) outgoing client events to the actual server (if you wish a full passthrough interception).

kettanaito:
The API itself is complete. One thing remaining is proper test coverage, specifically addressing the bug in the `ws` module (the `Event` instance check) or migrating to a different server package to use for testing.

kettanaito:
## Released: v0.26.0 🎉

This has been released in v0.26.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.26.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.26.0)

---

PR #508, opened 2024-02-16
fix(WebSocket): add "client.id" property

kettanaito:
When implementing broadcasting, there has to be an identifier for each client connection. The `client.id` string serves as the identifier. 

## On broadcasting API in the Interceptors

I believe that the Interceptors library should remain low-level. This means that higher-level features such as event broadcasting have to be implemented by higher-level libraries like MSW.


kettanaito:
## Released: v0.26.4 🎉

This has been released in v0.26.4!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.26.4)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.26.4)

---

PR #509, opened 2024-02-18
fix(WebSocket): exports client connection protocol, transport, renames types

kettanaito:


src/interceptors/WebSocket/WebSocketClientConnection.ts
```diff
@@ -5,44 +5,53 @@
...
-import type { WebSocketRawData, WebSocketTransport } from './WebSocketTransport'
+import type { WebSocketData, WebSocketTransport } from './WebSocketTransport'
 import { WebSocketMessageListener } from './WebSocketOverride'
 import { bindEvent } from './utils/bindEvent'
 import { CloseEvent } from './utils/events'
 import { uuidv4 } from '../../utils/uuid'
 
 const kEmitter = Symbol('kEmitter')
 
+export interface WebSocketClientConnectionProtocol {
```
kettanaito, line 16: Exporting this protocol means that other consumers, like MSW, can implement custom client connection-like classes. For example, to support `.clients` in the browser, MSW creates virtual connection objects that use `BroadcastChannel` to implement methods like `send()` (signal other clients from other runtimes that they should now receive data).

src/interceptors/WebSocket/WebSocketClientConnection.ts
```diff
@@ -5,44 +5,53 @@
...
+  close(code?: number, reason?: string): void
+}
+
 /**
  * The WebSocket client instance represents an incoming
  * client connection. The user can control the connection,
  * send and receive events.
  */
-export class WebSocketClientConnection {
+export class WebSocketClientConnection
```
kettanaito, line 28: Extending this class itself to implement connection-like objects is inefficient. It requires to provide a bunch of irrelevant properties, like `kEmitter`, `socket`, and `transport`. Those are internal details of this class but extending it means providing those internal details also.

kettanaito:
## Released: v0.26.5 🎉

This has been released in v0.26.5!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.26.5)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.26.5)

---

PR #435, opened 2023-09-21
feat: add "FetchInterceptor" to "node" preset

kettanaito:
- Closes #429

Issue #429, opened 2023-09-18
Add "FetchInterceptor" to the "node" preset


kettanaito:
The current version of this library supports Node.js 18 as the minimal required version. On that version, global fetch already exists.

kettanaito:
## Released: v0.34.0 🎉

This has been released in v0.34.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.34.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.34.0)

Make sure to always update to the latest version (`npm i @mswjs/interceptors@latest`) to get the newest features and bug fixes.

---

_Predictable release automation by [@ossjs/release](https://github.com/ossjs/release)_.

PR #513, opened 2024-02-29
fix(RemoteHttpInterceptor): leaking event emitters

mxschmitt:
Motivation: The problem was that the event emitter for IPC messages was added over and over again and reached the max event listener limits.

Solution: Add it once.

### DIFF
--- a/src/RemoteHttpInterceptor.ts
+++ b/src/RemoteHttpInterceptor.ts
@@ -4,8 +4,9 @@ import { Interceptor } from './Interceptor'
 import { BatchInterceptor } from './BatchInterceptor'
 import { ClientRequestInterceptor } from './interceptors/ClientRequest'
 import { XMLHttpRequestInterceptor } from './interceptors/XMLHttpRequest'
-import { toInteractiveRequest } from './utils/toInteractiveRequest'
+import { InteractiveRequest, toInteractiveRequest } from './utils/toInteractiveRequest'
 import { emitAsync } from './utils/emitAsync'
+import { FetchInterceptor } from './interceptors/fetch'
 
 export interface SerializedRequest {
   id: string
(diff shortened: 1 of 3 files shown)


src/RemoteHttpInterceptor.ts
```diff
@@ -29,23 +30,23 @@ export interface SerializedResponse {
...
+  [ClientRequestInterceptor, XMLHttpRequestInterceptor, FetchInterceptor]
 > {
+  requestId2Response: Map<string, { request: InteractiveRequest, resolve: () => void }> = new Map();
   constructor() {
     super({
       name: 'remote-interceptor',
       interceptors: [
         new ClientRequestInterceptor(),
         new XMLHttpRequestInterceptor(),
+        new FetchInterceptor(),
```
mxschmitt, line 42: Happy to move this change out (if wanted) of the PR.

src/RemoteHttpInterceptor.ts
```diff
@@ -67,45 +68,52 @@ export class RemoteHttpInterceptor extends BatchInterceptor<
...
-            if (!serializedResponse) {
-              return resolve()
-            }
+    const handleParentMessage: NodeJS.MessageListener = (message) => {
+      if (typeof message !== 'string') {
+        return
+      }
+      const startIdPosition = 'response:'.length;
+      const endIdPosition = message.indexOf(':', startIdPosition);
+      const requestId = message.slice(startIdPosition, endIdPosition)
```
mxschmitt, line 82: Instead of having it serialised manually I recommend putting the structure straight into the IPC message, since Node.js does JSON internally by itself. I'd do something like: `{ type: 'request', payload: {...}`.

Would you be open for such an enhancement?

kettanaito:
Hi, @mxschmitt. Can you please elaborate on what made you close this pull request? Did you discover the root of the problem elsewhere? Please mention it here for posterity. Thanks.

src/RemoteHttpInterceptor.ts
kettanaito: Yeah, these are covered in #429 (#435). Fetch interceptor wasn't in the node preset because we used to support Node.js 16 where fetch wasn't global. That's not the case anymore so we should proceed with merging that change.

PR #514, opened 2024-02-29
fix(WebSocket): reassign global to restore WebSocket

kettanaito:
Calling `.revoke()` on a revocable Proxy _does not restore its impementation_. I was wrong to assume that. It simply blows up the proxy object so _any object operations that result in a trap **throw**_. 

Also added a test to ensure that the restored `WebSocket` class is fully operational after the interceptor is disposed.

### DIFF
--- a/src/interceptors/WebSocket/index.ts
+++ b/src/interceptors/WebSocket/index.ts
@@ -48,7 +48,9 @@ export class WebSocketInterceptor extends Interceptor<WebSocketEventMap> {
   }
 
   protected setup(): void {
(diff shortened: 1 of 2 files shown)


kettanaito:
## Released: v0.26.7 🎉

This has been released in v0.26.7!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.26.7)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.26.7)

---

PR #495, opened 2024-01-07
feat(wip): async hooks interceptor

kettanaito:
A proof of concept of using `async_hooks` for request interception. Quite interesting.

Issue #455, opened 2023-09-28
Socket: Respect "connect" and "setTimeout" behaviors


mikicho:
I unified this in one issue because we need to take all of these into account in the solution.

1. Socket authorized and encrypted are false for https requests
```js
https.get('https://example.test').on('socket', socket => {
    expect(socket.authorized).to.equal(true) // always false, test fails
    expect(socket.encrypted).to.equal(true) // always false, test fails
    done()
  })
```

2. never emit (socket) `connect` event for an invalid URL (e.g. `example.test`) because Node tries to connect to the URL, and we capture the `ENOTFOUND`.
3. We can't use `socket.setTimeout` for invalid URL (e.g. example.test`) because it emits error

PR #515, opened 2024-03-03
feat(ClientRequest): use net.Socket interceptor

### LINKED ISSUE #399: Docs: Socket-based interceptor
Socket Interception
#ideas

Over the years, I sat down to write Socket-based interceptor at least half a dozen times. It never ended fruitfully. This time, I decided to at least document my findings and struggles so I don’t spend so much time finding my way in the same water once more.

## Intention
The intention here is very simple: the network in Node.js goes through `net.Socket`, no matter who and how initiates it. `http.ClientRequest` constructs a Socket instance, `fetch` (Undici) constructs a Socket instance, and even bi-directional communication like WebSockets is also implemented through Socket. Tapping into `net.Socket` directly would mean even less request-specific code.

### `net.Socket` is omitted  sometimes
Contradictory to my statement in the Intention, the call to `net.Socket`, `net.connect()` or `net.createConnection()` is not always performed when a request is made in Node.js. For example, the very default `http.get()` call will not result in either of those functions being called. There’s quite a chunk of internal logic in Node.js around creating and managing sockets (hence the `Agent`), and I can only speculate that it doesn’t create a *new* socket but it depends on the already created one, somehow.

### Socket is too low-level
Socket handles Buffer sent/received over the network. We are talking about raw HTTP messages and raw `Readable` streams. Reading those isn’t a problem, *recreating those* is. 

Re-creating that, eventually, becomes a necessity because if you extend `net.Socket` and emulate a successful connection to a host (since the actual connection would error, the host is mocked), you end up managing the entire socket’s life-cycle by yourself.

```js
class Socket extends net.Socket {
  connnect(...args) {
    // This is enough to pretend that the Socket instance
    // has looked up, resolved, and connected to a host.
    this.connecting = true
    this.emit('lookup', '0.0.0.0', 'IPv6', 'hostname')
    this.emit('connect', false)
    this.connecting = false
    return this
  }
}
```

There are more parts in the socket that depend on its state. For instance, you have to have the `this._handle` set, which is a reference to a `TCP` or `Pipe` wraps responsible for piping chunks through the socket. 

> To make things worse, those handles are internal in Node.js and you cannot construct them by yourself. [Reference](https://github.com/nodejs/node/blob/d150316a8ecad1a9c20615ae62fcaf4f8d060dcc/lib/net.js#L1230-L1231).

You end up implementing things like `handle.writeLatin1String`, `handle.readStart()`, `handle.close()`, and so forth. Honestly, this is too much.

## Good things
On the bright side, or, rather, the side I’ve never got to in order to prove it’s even worse, reading/writing chunks is a matter of manipulating the socket’s stream. The worst part about it is having to translate the HTTP request message to some consumable format, like `Request`, and translate the mocked `Response` to the HTTP response message. None of those are deal-breaking.

> Node.js doesn’t expose its HTTP message `parser`. It’s present on `ClientRequest` and the socket instances but you cannot operate with it easily.

Another good thing is that socket *observer* is trivial. It can be achieved with revokable Proxies spying at `.write` and `.on(‘data')`  methods/events. But observer alone is not enough to fulfill the interception requirements.

## Where to go from here
Socket-based interception is, technically, possible. But it’s a big question whether it’s worth the price. Maintaining such an interceptor would require a deep knowledge of Node.js internals and keeping track of whenever those change across Node.js releases. 

We may require to revisit this in the light of WebSockets support, but given how there’s no standard protocol for WebSockets, and the most popular solutions implement it over HTTP/XHR polling anyway, perhaps Socket-based interception won’t be needed entirely.

### LINKED ISSUE #400: [XHR] Response interceptors executes after application receives the response
## Issue Title

XHR Response interceptors get executed after the application receives the response.

## Description

I expect to see the response interceptor functions execute before the application receives the response. This is not the case with `XMLHttpRequestInterceptor`. It is, however, the case with `FetchInterceptor`

## Context

I am working on a solution where we need to intercept all HTTP requests fired by any client (fetch, Angular, Axios, etc). There's no telling which underlying HTTP method is used to make the request, so it should cater to all. 

We came across this `@mswjs/interceptors` package that looked promising as it allows us to intercept all requests, but in the POC, we encountered some strange behaviour.

The expected flow would be:
```
HTTP request initiated

  - Request interceptor 1
  - Request interceptor 2
  - Request interceptor ...n

  - Response interceptor 1
  - Response interceptor 2
  - Response interceptor ...n

HTTP Response received by the application
```

This should be the same for all HTTP requests.

## Current Behavior

I've noticed that XHR requests it does not follow the above pattern. Instead, they do the following:

```
HTTP request initiated

  - Request interceptor 1
  - Request interceptor 2
  - Request interceptor ...n

HTTP Response received by the application

  - Response interceptor 1
  - Response interceptor 2
  - Response interceptor ...n
```

## Expected Behavior

I expect the interceptor functions to be executed in the same order, whichever HTTP method is used.

**Request** interceptor functions should be executed <ins>before the request reaches the API</ins>, and **Response** interceptor functions should be executed <ins>before the response reaches the application</ins>.

## Screenshots and Additional Information (if applicable)

image

## Steps to Reproduce (if applicable)

Please see the reproduction repo where the below behaviour can be seen:
Github Repo: https://github.com/JAspeling/mswjs-interceptors-angular/blob/main/src/main.ts
Stackblitz: https://stackblitz-starters-dgrmdd.stackblitz.io
Stackblitz editor: https://stackblitz.com/edit/stackblitz-starters-dgrmdd?file=src%2Fmain.ts

* Use the `XMLHttpRequestInterceptor` as a BatchInterceptor.
* Add request interceptors
* Add response interceptors
* Make an XMLHttpRequest call to an API endpoint:
```ts
xhrRequest(): void {
    console.log('============== XHR [APPLICATION REQUEST INITIATED] ==============');
    const xhr = new XMLHttpRequest();

    xhr.onreadystatechange = function () {
      if (xhr.readyState === 4 && xhr.status === 200) {
        var data = JSON.parse(xhr.responseText);
        console.log('============== XHR [APPLICATION RESPONSE RECIEVED] ==============');
      }
    };

    xhr.open('GET', this.api, true);
    xhr.send();
  }
```
* See the order of execution in the console logs.

## Environment

```
Angular CLI: 16.1.6
Node: 18.17.0
Package Manager: npm 9.6.7
OS: win32 x64

@angular-devkit/architect       0.1601.6
@angular-devkit/build-angular   16.1.6
@angular-devkit/core            16.1.6
@angular-devkit/schematics      16.1.6
@angular/cli                    16.1.6
@schematics/angular             16.1.6
rxjs                            7.8.1
typescript                      5.1.6
zone.js                         0.13.1
```

----

Please feel free to reach out if anything is unclear or if there is a misunderstanding around using the `@mswjs/interceptors` package.

### LINKED ISSUE #439: Support http.request flushHeaders

### LINKED ISSUE #443: response.complete is false in end listener
We set the `this.response.complete` after the `response.push(null)`, which emits the `end` event.
Therefore, I think the `response.complete` is always false in the `end` event.

IIUC the docs, it should be true:
https://nodejs.org/api/http.html#messagecomplete

I think we should switch these lines: https://github.com/mswjs/interceptors/blob/main/src/interceptors/ClientRequest/NodeClientRequest.ts#L489-L490

### LINKED ISSUE #457: Handle res.destroy in "response" event
```js 
const http = require('http');

  name: 'my-interceptor',
})
interceptor.apply();
  request.respondWith(new Response('hello'))
});

http.get('http://nowhere.com/', res => {
  res.destroy(new Error('hey'))
})
.on('error', e => {
  console.log(e) // never get here!!!
})
```

Instead of getting to the req's `error` listener, this throws an error.

Expected behavior: I think we need to listen to the `error` event in the request (from the socket) and pass it through.
[docs](https://nodejs.org/api/http.html#messagedestroyerror)

### LINKED ISSUE #458: req.write should throw error where chunk is undefined
https://github.com/mswjs/interceptors/blob/main/src/interceptors/ClientRequest/NodeClientRequest.ts#L101

`chunk` must not be undefined and throw:
```
TypeError [ERR_INVALID_ARG_TYPE]: The "chunk" argument must be of type string or an instance of Buffer or Uint8Array. Received undefined
```

 I can open a PR, but because you ignored this case positively, I want to ensure I don't miss something.

About the solution, we can either:

1. call to super.write() and let Node emit the error
2. Return a custom error

kettanaito:
Yet another attempt at the Socket-based interceptor. This time, we are tapping into the `HTTPParser` from Node.js to give us the HTTP message parsing. 

- Makes it possible to implement #374
- Closes #375 
- Closes #399
- Fixes #400
- Closes #439
- Closes #442 
- Fixes #443
- Closes #464
- Fixes #457
- Fixes #458
- Closes #495 

## Roadmap

- [x] Implement `SocketController.respondWith()`. The challenge is to convert the Fetch API `Response` to the HTTP message and push it to the socket instance. We can use https://github.com/har-tools/http-message if nothing else works. 
- [x] Implement error replays. Depends on how we handle the compatibility tests below. 
  - I think duplicate `connect` and `ready` events are OK since those are internal and no consumers really depend on them. The main worry is the duplicate `.write()` callbacks since the request body would be written _twice_ in the case of request listener (body read 1) + a bypassed request to an existing hostname (body read 2). We can always suppress the callbacks and then replay them. 
- [x] ~~Push the `mockConnect()` event until _after_ the "request" interceptor event has been awaited~~. 
  - Cannot do this. `ClientRequest` won't start writing the request body until it receives the `"connect"` and `"ready"` events from the socket. 
- [x] Add a new test suite for compatibility
  - 1: Bypassed request to an existing host (no request listeners).
  - 2: Bypassed request to a non-existing host (no request listeners).
  - 3: Mocked request to an existing host (no response returned).
  - 4: Mocked request to an existing host (mocked response returned).
  - 5: Mocked request to a non-existing host (no response returned).
  - 6: Mocked request to a non-existing host (mocked response returned).
- [x] Socket compliance:
  - [x] Respect the `signal` Socket constructor options (I presume for aborting the connection). Consider `AbortSignal.timeout(n)` as well. https://github.com/mswjs/interceptors/pull/559
  - [x] Respect the right order of events in all compatibility scenarios. 
  - [x] Support `Socket.setTimeout()` that should time out the connection if it hasn't been established within the given time window. **Support these for _actual_ socket connections**. https://github.com/mswjs/interceptors/pull/558
  - [x] #594 Add compliance tests for the case when a single Socket is reused for multiple requests. We need to collocate request/responses correctly and handle their streams independently (likely bound the request/response stream to particular request/response instances instead of setting it as a property on the `MockHttpSocket` class).
- [x] **Support TLS connections**. Those use `tls.connect()`, not `net.connect()`.
- [x] ~~Add compliance test for non-HTTP transfer over Socket (e.g. SMTP, see [this](https://github.com/mswjs/interceptors/pull/375/files#diff-47f950babb2214bc9edd1f6c9d923f52d9e289f43485c4a7d7b635fc6b1852fcR20-R26)). We should do nothing about these, it's out of scope for this feature.~~
  - Irrelevant since we aren't patching `net.connect()` anymore. We utilize `MockSocket` to create a custom `http.Agent` and it's only applied to the `http` module. 
- [x] Destroy the `this.requestStream` and `this.responseStream` at the right time. Consider that the consumer might've not consumed either of the streams by the time the `*End()` methods on `MockHttpSocket` are called (that will cause a premature close error).
- [x] Clean up `ClientRequest/utils` for it contains a lot of unused code now. https://github.com/mswjs/interceptors/pull/557
- [x] Inherit TLSSocket properties like `.authorized` and `.encrypyed` (see #455). This is related to inheriting other properties like timeout and signal. (https://github.com/mswjs/interceptors/pull/556)
- [x] Update `README.md` since the interception algorithm is fundamentally different now. 
- [x] **Fix `follow-redirect-http` test, it's pending forever** (skipped in 230c238ee86806688df57829cb4827a3b8b8f8c0). 
- [x] https://github.com/mswjs/interceptors/pull/548
- [x] Implement deferring of the response `end` event until all response event listeners are done (#591 backport)


mikicho:
> Yet another attempt at the Socket-based interceptor

The N time's a charm. :)

> Since the body stream is the last thing we need to call the request listener, we can call the listener even before deciding whether to emit the connect+ready on the socket instance.

I like the direction you are going!

Would it work in the following case?
```js
const req = http.request(...)
req.write('sync') // OK

await sleep(1000)
req.write('async') // I think it will work, but I'm not sure about it. my concern is that the client library would await for the `connect` event to continue
req.end()
```

mikicho:
I think this http_parser is about [to be deprecated](https://github.com/nodejs/llhttp?tab=readme-ov-file#why), as well as [process.binding](https://github.com/nodejs/node/issues/22064) itself. (although it seems like no one bothers to actually remove it).

src/interceptors/Socket/SocketInterceptor.ts
```diff
@@ -0,0 +1,450 @@
...
+
+  constructor() {
+    super(SocketInterceptor.symbol)
+  }
+
+  protected setup(): void {
+    const self = this
+    const originalConnect = net.Socket.prototype.connect
+
+    net.Socket.prototype.connect = function (normalizedOptions) {
```
gioragutt, line 43: Should this be `...normalizedOptions` here, if it's passed on as arguments to connect?

kettanaito:
## Request event duplication issue

With this approach, we have to mock the connection to the host every time in order for the `ClientRequest` (the higher-level consumer) to start streaming the buffered request body through the socket (it won't start that until it receives the `connect` event).

This means a few different sequences of events depending on the type of the request. 

### Request to an existing endpoint with a mocked response

1. lookup (mocked).
1. connect (mocked).
1. ready (mocked).
1. INTERNAL.
    1. Reading the request body.
    1. Firing "request" on the intercepted.
    1. Piping the mocked response to the socket.
1. data.
1. response.
1. end.

### Request to a non-existing endpoint with a mocked response

> The same thing as the one above.

### Request to a non-existing host without a mocked response

This is where things get a bit tricky. 

1. lookup (mocked).
1. connect (mocked).
1. ready (mocked).
1. INTERNAL (the same request streaming to the interceptor).
1. **error**. Emit the suppressed lookup (connection) error since there's no mock in the interceptor so the consumer must receive the error about requesting a non-existing host. 

This scenario is different from the event order in the unpatched scenario. The initial `lookup` event would include an error, and mirror itself in the `error` event on the socket also. 

In our case, the lookup/connect/ready _are successful_. This is problematic for a few reasons:

1. Diverges from the "normal" event flow.
1. The socket errored and when the client will try writing to it (request body), while we will be able to forward those chunks to the HTTPParser, the underlying socket will throw `Error: write EBADF` error, which I believe indicates an invalid write.

src/interceptors/Socket/SocketInterceptor.ts
kettanaito: @gioragutt, I think it should be but Node.js provides here a single normalized object of options. Let's dive into the source for `net.Socket.connect` to verify that's the case.

gioragutt: If that's the case, how is the spread in line 44 working? JS syntax wise it wouldn't make sense otherwise

kettanaito:
@mikicho, I believe we've achieved an MVP here with https://github.com/mswjs/interceptors/pull/515/commits/1c74f2399f9af87c8ef3800568e2757db07eaadd. I've got the HTTP and Socket compliance test suites passing, and we even support request-less `net.Socket().connect()` use cases, which is nice. 

> That actually makes me wonder if the Socket interceptor should have a `"connect"` event that allows one to decide how to _resolve_ (in the DNS sense) the outgoing connection. This is a food for thought for the future, we aren't going to address that now because it's a non-existing use case for us. 

Could you please give this a try in the Nock test suite? That'd be rad.

src/interceptors/Socket/SocketInterceptor.ts
kettanaito: I write about this here:

https://github.com/mswjs/interceptors/blob/a755028997ffd8f2a02e91b316e67ec9084bdb51/src/interceptors/Socket/SocketInterceptor.ts#L61-L73

TL;DR Node.js will normalize the args for `.connect()` _in some cases_ but not the others. That's why we get different type of args if the socket connection is called as a part of `http.request()` and raw like `new Socket().connect()`.

mikicho:
> this.mockConnect(this.shadow)

How is this propogate the `connect/ready/lookup` events to the ClientRequest so it will continue to write the data?

> Catch 22

>Problems with this solution:

Probably not a problem, but I want to make sure I understand, in this option we will emit `connect/ready/lookup` twice in case of non-mocked request, right?

> While no direct consumers care about the lookup/connect/ready events

It seems like `Axios` doesn't use this, but `got` [exposes this data](https://github.com/sindresorhus/got/blob/897f3857a8af1baa440d5fbf3918be7bac6938bb/source/core/response.ts#L76) to its users, I'm not sure for what tho...

mikicho:
I opened a feature request for official mocking API like undici has: https://github.com/nodejs/node/issues/51979

kettanaito:
> It seems like Axios doesn't use this, but got [exposes this data](https://github.com/sindresorhus/got/blob/897f3857a8af1baa440d5fbf3918be7bac6938bb/source/core/response.ts#L76) to its users, I'm not sure for what tho...

It doesn't expose that data but rather hooks into those socket events to provide timings measurements. That's okay. If you take a look at the current state of the branch, it emits the events correctly for all scenarios we've talked about. There's no duplication. 

> I opened a feature request for official mocking API like undici has: https://github.com/nodejs/node/issues/51979

This is great! I'd certainly love to hear Node's opinion on this. I would count on this landing any time soon so our work on the Socket interceptor is relevant for at least a few years. 

> Probably not a problem, but I want to make sure I understand, in this option we will emit connect/ready/lookup twice in case of non-mocked request, right?

Not anymore.

mikicho:
> Not anymore.

This is impressive! how so?

kettanaito:
## Need help

I've pushed the latest state of the changes that supports both HTTP and HTTPS requests. I've moved things around a bit, mainly made the `MockSocket` to be agnostic of the type of the messages sent (just giving us the observability), introduced a `MockHttpSocket` and utilizes the mock socket and pipes its streams through the HTTP parser, and then integrates with the new `_ClientRequestInterceptor` that's now `Agent`-based (we aren't mocking the `ClientRequest` at all now). 

I'm encountering errors when intercepting HTTP/HTTPS requests _with a body_. Error:

```
pnpm test:node test/modules/http/intercept/http.request.test.ts
```

```
 FAIL  modules/http/intercept/http.request.test.ts > intercepts a POST request
AbortError: The operation was aborted
 ❯ Readable.emit ../node:events:513:28

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: { code: 'ABORT_ERR' }
Caused by: Error: Premature close
 ❯ Readable.emit ../node:events:513:28

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: { code: 'ERR_STREAM_PREMATURE_CLOSE' }
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
```

This indicates either that the `MockSocket` connection has been closed (not the case, the connection is never established to begin with), or that the `MockSocket` is still in writable state when we try to `.emit('data')` when forwarding data events from the original socket (in the `passthrough()` method). The latter seems to be more likely since `this.writable` is still `true`. You can also observe the `finish`/`end` events on the mock socket to see that they are emitted _after_ the `passthrough()` start. 

This makes sense because right now we call `socket.passthrough()` in the interceptor on the `request` event, which fires _as soon as the request headers are sent_. There may be some time passing until the request body has been written completely. I think this indicates an issue that we mustn't start the passthrough until the socket is finished or is in the `readable` state. 

I've tried things like

```js
// the interceptor
socket.once('end', () => socket.passthrough())
```

But they haven't fixed the problem.

kettanaito:
## `ERR_INVALID_ARG_TYPE` on the `listener` in JSDOM

The `callback` argument to any `http*.*` methods throws the following error _only in JSDOM_:

```
npx vitest -c test/vitest.config.js test/third-party/axios.test.ts
```
```
Error: TypeError [ERR_INVALID_ARG_TYPE]: The "listener" argument must be of type function. Received an instance of Object
    at Object.dispatchError (mswjs/interceptors/node_modules/.pnpm/jsdom@16.7.0/node_modules/jsdom/lib/jsdom/living/xhr/xhr-utils.js:63:19)
    at EventEmitter.<anonymous> (mswjs/interceptors/node_modules/.pnpm/jsdom@16.7.0/node_modules/jsdom/lib/jsdom/living/xhr/XMLHttpRequest-impl.js:655:18)
    at EventEmitter.emit (node:events:526:35)
    at mswjs/interceptors/node_modules/.pnpm/jsdom@16.7.0/node_modules/jsdom/lib/jsdom/living/xhr/xhr-utils.js:328:49
    at processTicksAndRejections (node:internal/process/task_queues:77:11)
```

This passes in Node.js so I suspect JSDOM is meddling with the callback function somehow so it's no longer a function.

kettanaito:
Have to solve the interceptors event deduplication somehow without having to rely on `x-request-id` anymore. AsyncLocalStorage works great but it's Node-specific. Tried throwing promises to little success. I can almost feel we can do some `.emit` magic on the `BatchInterceptor` if each individual interceptor will describe how to identify same events.

mikicho:
@kettanaito isn't this Node only interceptor?

mikicho:
I think I found a small edge case on empty streams:
```js
const got = require('got')

interceptor.on('request', function rootListener({ request }) {
    request.respondWith(new Response(
      new ReadableStream({
        start(controller)  {controller.close()}
      })
    ))
})
interceptor.apply()

got('http://example.text').then(console.log)
```

I solved this by pushing an empty string before closing, but I'm not an expert in streams, so maybe there is a more "correct"/straightforward solution.

mikicho:
in `createResponse`, we should set the body as null on `HEAD` requests.

mikicho:
HEAD requests with response body:
```js
const got = require('got')

interceptor.on('request', function rootListener({ request }) {
    request.respondWith(new Response('hello'))
})
interceptor.apply()

got.head('http://example.text').then(console.log)
```

Throw error: `RequestError: Parse Error: Expected HTTP/`

Now, it may make sense because [HEAD requests should not have a body](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD), but nothing enforces this, so I'm wondering if we need to throw a more friendly error on `createResponse`

kettanaito:
@mikicho, not really. The issue is fundamental. In the `BatchInterceptor`, you can compose multiple interceptors into a single interface. An interceptor can run in any environment. It's up to the `checkEnvironment` method on the interceptor, really. But the issue of one interceptor "falling through" to the other doesn't end with Node.js.

kettanaito:
@mikicho, I've fixed the empty ReadableStream issue. I've added a test for the response body for a HEAD request but couldn't reproduce it. May be a `got` issue. Feel free to look at the test (`test/modules/http/compliance/http-head-response-body.test.ts`) and see if something is missing there.

mikicho:
It seems like there is no authorized property in the https socket.
image
image

mikicho:
```js
const http = require('http')
const { Stream } = require('stream')

const a = new Stream.PassThrough()
interceptor.on('request', function rootListener({ request }) {
  request.respondWith(new Response(a))
  setTimeout(() => a.push('test'),3000)
})
interceptor.apply()

http.get('http://example.com', res => {
  res.on('data', console.log) 
  res.on('end', () => console.log(1)) // Never get called with stream. it works for new Response('hello')
  console.log('hey');
})
```

Somehow, we skip the `end` event when the response is a stream. The `data` event emitted normally.
To summary:
1. For `request.respondWith(new Response('hello'))`
image

2. `request.respondWith(new Response(a))`
image

kettanaito:
@mikicho, are you ending the stream though? It looks like you write to it but never actually close it, thus the `end` is never emitted? 

```diff
const a = new Stream.PassThrough()
interceptor.on('request', function rootListener({ request }) {
  request.respondWith(new Response(a))
  setTimeout(() => {
    a.push('test')
+   a.push(null) // indicate that's the end of the stream
  }, 3000)
})
```

mikicho:
You are correct. I always forget this null thing!

kettanaito:
## Destroying HTTP parsers

The HTTP parsers are now destroyed like this:

- The request parser is destroyed on the `finish` event of the socket. That event indicates that the socket has been ended (`.end()` was called) and the socket has flushed any pending data (`finish` is emitted). Once these two happen, you can no longer write to the socket, and so the parser won't ever receive any new request HTTP messages.
- The response parser is destroyed on the `close` event of the socket. If the `close` is emitted, the socket has been destroyed, and nothing can push to it anymore. 

This should work fine memory-wise. 

We still need to design a better way of handling `this.requestStream` and `this.responseStream` since those are inherent to individual requests/responses, and not the entire Socket instance. Basically, need to add a test for when a single Socket is reused for multiple requests/responses.

kettanaito:
@mikicho, I'm going to take a break from this, potentially until the end of the month. Please, feel free to tackle any remaining tasks and report/fix any issues you discover from the Nock's test suite. I believe we are at an extremely solid foundation here and I cannot wait to see this merged. 

Full disclosure, I don't see Nock compatibility as the requirement to merge this. As soon as we land all the remaining todos in the pull request's description, I consider this task done. We should merge and release it and then tackle missing bits for Nock separately. 

Thank you once again for the help with this. Excited to circle back on this in April!

mikicho:
`write` callback never get called.
I guess we don't emit the required event in the socket.
```js
const got = require('got')
const http = require('http')

interceptor.on('request', async function rootListener({ request }) {
  request.respondWith(new Response('OK!'))
})
interceptor.apply()


const r= http.request('http://example.com', res => {
  console.log(1);
  res.on('data', console.log)
  res.on('end', () =>console.log('end'))
})

r.write('dsa', undefined, () => {
  console.log(2); // never get called. it does when I comment out interceptor.apply()
  r.end()
})

```

mikicho:
No `abort` event
```js
const http = require('http')
const sinon = require('sinon')

interceptor.on('request', async function rootListener({ request }) {
  request.respondWith(new Response('OK!'))
})
interceptor.apply()

const req = http.request('http://example.com')

req.on('finish', () => {
      req.abort()
    })
    req.on('error', err => {
      console.log(err);
    })
    req.end()


    setTimeout(() => {
      const events = emitSpy.args.map(i => i[0])
console.log(events)
    }, 1000)
```

With `apply()`
> [ 'socket', 'prefinish', 'response', 'close' ]
Without
> [ 'socket', 'prefinish', 'finish', 'abort', 'error', 'close' ]

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -0,0 +1,450 @@
...
+      url.username = ''
+      url.password = ''
+    }
+
+    // Create a new stream for each request.
+    // If this Socket is reused for multiple requests,
+    // this ensures that each request gets its own stream.
+    // One Socket instance can only handle one request at a time.
+    if (canHaveBody) {
+      this.requestStream = new Readable()
```
mikicho, line 338: Don't we need to implement the `read` function here?

I get:
> Error [ERR_METHOD_NOT_IMPLEMENTED]: The _read() method is not implemented
    at Readable._read (node:internal/streams/readable:904:9)
    at Readable.read (node:internal/streams/readable:737:12)
    at maybeReadMore_ (node:internal/streams/readable:891:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:

When trying to use stream request:
```js
const http = require('http')
const { Readable } = require('stream');

interceptor.on('request', async function rootListener({ request }) {
  console.log(await request.arrayBuffer());
  request.respondWith(new Response('OK!'))
});
interceptor.apply();

const req = http.request('http://webhook.site/b1c9419d-4706-4f1c-9039-6825cfa5e77f', { method: 'POST' }, res => {
  console.log(res.statusCode)
})
let counter = 0
const r = new Readable({
  read: () => {
    if (counter++ < 3) {
      r.push('dsa1')
    } else {
      r.push(null)
    }
  }
})

r.pipe(req)

setTimeout(req.end.bind(req), 300)
```

kettanaito:
@mikicho, `connect`, `secure`, `secureConnect`, and `session` all should be called in TLS sockets. I have a compliance test that confirms that. If they are not, you are likely threading in a use case we are not covering properly? Need to distill that use case to the minimum and turn it into an automated test 👀

mikicho:
I think we need to add the chunk size on write when transfer encoding is chunked. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding

I think can open a PR to fix this.

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: Added in #527

kettanaito:
@mikicho, let's address the `Text-Encoding` issue separately once this merges. 

Please, if you have a moment, could you look into why `test/third-party/follow-redirect-http.test.ts` is pending forever with the Socket interceptor?

```sh
pnpm test:node test/third-party/follow-redirect-http.test.ts
```

It'd be a great help! 🙏

kettanaito:
I suspect the following redirects test suffers from a recursive `net.Socket` construction, which leads to a pending test and the worker getting out of memory. There's no protection in place for scenarios when a passthrough request creates another Socket that matches that request, which I suspect happens in the `follow-redirect` scenario. I have not confirmed this, just a wild guess.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -0,0 +1,457 @@
...
+  private shouldKeepAlive?: boolean
+
+  private responseType: 'mock' | 'bypassed' = 'bypassed'
+  private responseParser: HTTPParser<1>
+  private responseStream?: Readable
+
+  constructor(options: MockHttpSocketOptions) {
+    super({
+      write: (chunk, encoding, callback) => {
+        this.writeBuffer.push([chunk, encoding, callback])
```
mikicho, line 65: I think we need to call the callback after we "flush" the chunk to the OS.
Related to this problem: https://github.com/mswjs/interceptors/pull/515?notification_referrer_id=NT_kwDOAK7cMLM5Njk2NjM3NjYxOjExNDU5NjMy#issuecomment-1995549971

kettanaito: We shouldn't be doing any of this since we are writing to the socket, unless I'm getting it wrong, of course. That's the point of relying on `net.Socket` under the hood 🤔 

@mikicho, can you please share a use case that illustrates how this manifests?

mikicho: @kettanaito The thing is that the native `write` function we override (or one of its derivatives) changes the internal `_writableState`, which is used later.

>  can you please share a use case that illustrates how this manifests?
I think this is the root cause of this: https://github.com/mswjs/interceptors/pull/515#issuecomment-1995549971

And here is a failing test for `@mswjs/interceptors`:

```js
it.only('should fire finish event', async () => {
  interceptor.on('request', async ({ request }) => {
    request.respondWith(new Response(null, { status: 200 }))
  })
  const finishSpy = vi.fn()
  const request = http.request(httpServer.http.url('/resource'), {
    method: 'POST',
  })

  request.on('finish', finishSpy)
  request.end()

  const { res } = await waitForClientRequest(request)
  expect(res.statusCode).toBe(200) 
  expect(finishSpy).toHaveBeenCalledTimes(1) // AssertionError: expected "spy" to be called 1 times, but got 0 times
})
```

kettanaito: @mikicho, opened a fix in https://github.com/mswjs/interceptors/pull/542. Your test is passing now. Also added a few missing tests. We weren't calling write callbacks in the mocked response scenarios at all, like you've pointed out. Also found a bug around `.push()` and the `end` event.

mikicho:
It seems like we don't propogate mocked response error to the client:

```js
it.only('error', async () => {
  interceptor.removeAllListeners()
    const replyBody = new PassThrough()
    request.respondWith(new Response(new ReadableStream({
      start(controller) {
        replyBody.on('data', (chunk) => controller.enqueue(chunk))
        replyBody.on('error', (error) => controller.error(error))
      },
    })))
    replyBody.push('start')
    setTimeout(() => { replyBody.emit('error', 'oh no!') }, 50)
  });
  const responseErrorPromise = new DeferredPromise<Error>()

  http.get('http://localhost:3001/resource', response => {
    response.on('error', responseErrorPromise.resolve)
  })

  const responseError = await responseErrorPromise

  expect(responseError.message).toBe('oh no!')
})
```

The error comes from [this line](https://github.com/mswjs/interceptors/blob/92343e64d6fcdf4737d93056ed8514f80ba7b172/src/interceptors/ClientRequest/MockHttpSocket.ts#L245), but I couldn't figure out how to propagate the error.
I tried `try {...} catch(e) this.emit('error', e)` but for some reason, it didn't work.

kettanaito:
@mikicho, that's a great catch! Do you know how that error is handled in an non-mocked scenario? Does it propagate as the `error` event on the `ClientRequest`? 

I think doing this should work:

```ts
await reader.read().catch((error) => /* handle */)
```

mikicho:
> Do you know how that error is handled in a non-mocked scenario?

Hm.. this is a good question, I couldn't reproduce it with a real request.
According to [the docs](https://nodejs.org/api/http.html#class-httpincomingmessage), `IncomingMessage` shouldn't throw and `error` event, so maybe we should cover this only from mocked request perspective?

mikicho:
The `socket.address()` returns an empty object in `connect` event.
I guess it's because we don't establish connection for mocked requests, I'm wondering if we should fake the object, just in case some library expect for these properties (e.g `socket.address().port.toString()`)

```js
const http = require('http')

  name: 'my-interceptor',
})
interceptor.apply();

const req = http.get('http://example.com')
req.once('socket', socket => {
  socket.once('connect', () => {
    console.log(socket.address()) // Mocked: {}. Non-mocked: { address: '172.27.36.212', family: 'IPv4', port: 42158 }
  })
})
```

kettanaito:
Opened a fix to forward TLS Socket properties (authorized, encrypted, etc.): https://github.com/mswjs/interceptors/pull/556

mikicho:
Maybe we want to merge [this](https://github.com/mswjs/interceptors/pull/489) to fix IPv6?

mikicho:
Nock (unintentionally?) covers an edge case of a request that ends in the `connect` event.
With current socket-based implementation the request never ends.

```js
const http = require('http');

  name: 'my-interceptor',
})
interceptor.apply();
  request.respondWith(new Response())
});


const req = http.request('http://example.com')

req.on('socket', socket => {
  socket.once('connect', () => {
    console.log(1);
    req.end()
  })
  socket.once('secureConnect', console.log)
})
```

WDYT?

mikicho:
How do we need to handle `Expect: 100-continue` request header?

What Node.js does:
1. flush headers ([docs](https://nodejs.org/api/http.html#event-checkcontinue:~:text=Sending%20an%20%27Expect%27%20header%20will%20immediately%20send%20the%20request%20headers.%20Usually%2C%20when%20sending%20%27Expect%3A%20100%2Dcontinue%27%2C%20both%20a%20timeout%20and%20a%20listener%20for%20the%20%27continue%27%20event%20should%20be%20set.%20See%20RFC%202616%20Section%208.2.3%20for%20more%20information.), [source](https://github.com/nodejs/node/blob/93da936f14321b681d6e6ea66a9060f057a44e4d/lib/_http_outgoing.js#L591))
2. Server side (should) returns 100 Continue response ([docs](https://nodejs.org/api/http.html#event-checkcontinue))
3. Node emits `continue` event. ([docs](https://nodejs.org/api/http.html#event-continue))

<details>
  <summary>Reproducible code</summary>

```js
const http = require('http');
  name: 'my-interceptor',
})
interceptor.apply();
  request.respondWith(new Response())
});
const req = http.request({
  host: 'example.com',
  path: '/',
  headers: { Expect: '100-continue' },
})
req.on('response', res => {
  res.on('data', console.log)
  res.on('end', () => { })
})
req.on('continue', () => {
  console.log(1); // never get called when interceptor applied.
  req.end()
})
```
</details>

I wonder if we need to emit a `continue` event before calling the interceptor.
I think I can open PR when we decide how to handle this.

src/interceptors/ClientRequest/index.ts
```diff
@@ -1,61 +1,218 @@
...
+            respondWith: socket.respondWith.bind(socket),
+            errorWith: socket.errorWith.bind(socket),
+          },
+        })
+
+        // After the listeners are done, if the socket is
+        // not connecting anymore, the response was mocked.
+        // If the socket has been destroyed, the error was mocked.
+        // Treat both as the result of the listener's call.
+        if (!socket.connecting || socket.destroyed) {
```
kettanaito, line 179: So in the Socket world, the socket won't be `writableEnded` until `.end()` is called, which can take time (e.g. request body is streaming). 

Instead, I rely on `.connecting`, which we immediately set to `false` now in `.mockConnect()`, or the `.destroyed`, which will also immediately be set to false it you mock an error.

kettanaito:
> Nock (unintentionally?) covers an edge case of a request that ends in the connect event.

@mikicho, we cannot cover this. `connect` depends on `req.end()` so it cannot be called in the connect event listener when the interceptor is enabled. Are there any other details around that test case in Nock (is it a regression)? 

> How do we need to handle Expect: 100-continue request header?

This reads to be an `http.ClientRequest` level of behavior. Is it not happening already? That's odd. 

We should support it but we should also be careful not to prematurely support it. As in, if it shouldn't belong to the Socket level and we ignore a potential bug that prevents Node.js from calling `checkContinue()` as it does in a non-interceptor scenario. 

Let's add an integration test for this first and then see! Would love to dive deeper to learn what exactly Node.js does step-by-step, and at which step it stops when using the interceptor.

kettanaito:
@mikicho, hi! You've mentioned there are also some remaining issues left here. Could you please post them in the comment to this pull request? I'd love to see what's left so I can tackle some of them whenever I have a moment. Thank you!

mikicho:
> This reads to be an http.ClientRequest level of behavior. Is it not happening already? That's odd.

I understand the issue.
Nock emits `continue` automatically when `Expect: 100-continue` is present ([source](https://github.com/nock/nock/blob/main/lib/intercepted_request_router.js#L104)).

---
As far as I understand, `mswjs/interceptors` should be a platform for building tools like Nock and MSW. So I tried to mimic the [`writeContinue`](https://nodejs.org/api/http.html#responsewritecontinue) behavior on Nock's side in the `request` event, but then I encountered a couple of problems:

1. `Response` does not support `100` HTTP code ([source](https://github.com/nodejs/node/blob/02bd8662985a2c4e11f3235f3fc5bb28b801c3c4/deps/undici/src/lib/web/fetch/response.js#L465))
> Uncaught RangeError: init["status"] must be in the range of 200 to 599, inclusive.
PS: I don't know why Undici has this limitation, and I couldn't find it in the spec.
2. Even if the point above is solved. With the current `respondWith` API, how can we return two responses (one with 100 and another when we get the request body)

Having said that, this looks like an esoteric edge case, so if it requires API changes, we might want to automatically emit the `continue` event inside `mswjs/interceptors` like Nock does.

kettanaito:
# Destroying request/response parsers

We are currently correctly destroying the parsers, specifically:

The request parser is destroyed once the request emits the `finish` event, indicating that the request can no longer be modified:

https://github.com/mswjs/interceptors/blob/18ce9e425bfc8efaa21893a1c247c28d67020a19/src/interceptors/ClientRequest/MockHttpSocket.ts#L113

The response parser is destroyed once the `.destroy()` method is called on the socket, indicating that the socket is no longer operable:

https://github.com/mswjs/interceptors/blob/18ce9e425bfc8efaa21893a1c247c28d67020a19/src/interceptors/ClientRequest/MockHttpSocket.ts#L130

> The `.destroy()` method is called for both normal request handling and in case of an exception. We are always freeing the response parser.

kettanaito:
I believe this feature is ready for merging! 🎉 The test suite passes, the compatibility looks solid. 

There are a few pending tasks for the Nock compatibility, which we've agreed with @mikicho to tackle after this is released. I don't wish to stall this update any longer. It's a huge change and it blocks any bugfixes with having to port them here as well. 

Looking forward to releasing the Socket-based interceptor sometime this week!

kettanaito:
## Released: v0.32.0 🎉

---

PR #518, opened 2024-03-12
fix(WebSocket): emit "connection" event on the next tick

kettanaito:
Right now, the `connection` event is emitted to the interceptors on the same tick as the WebSocket instance is created. It makes it impossible to observe the changes to that instance while connecting. 

For example, the client may set a different `.binaryType` on the socket _after_ it's constructed (i.e. starts connecting):

```js
const socket = new WebSocket(url)
socket.binaryType = 'arraybuffer'
```

For the interceptor to know about that change, we have to emit the connection event on the next tick, when any modifications to the socket are complete (including the event listeners). 

This will allow us to properly inherit `.binaryType` on the real socket when calling `server.connect()`. This also gives us a behavior similar to production code, where the `open` event will always be emitted on the next tick. The `connection` interceptor event must _not_ be emitted before the open event.

### DIFF
--- a/test/modules/WebSocket/compliance/websocket.connection.test.ts
+++ b/test/modules/WebSocket/compliance/websocket.connection.test.ts
@@ -4,6 +4,7 @@
 import { vi, it, expect, beforeAll, afterAll } from 'vitest'
 import { WebSocketInterceptor } from '../../../../src/interceptors/WebSocket/index'
 import { UUID_REGEXP } from '../../../helpers'
+import { waitForNextTick } from '../utils/waitForNextTick'
 
 const interceptor = new WebSocketInterceptor()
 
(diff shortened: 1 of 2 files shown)


test/modules/WebSocket/compliance/websocket.connection.test.ts
```diff
@@ -15,30 +16,35 @@ afterAll(() => {
...
-          addEventListener: expect.any(Function),
-          removeEventListener: expect.any(Function),
-        }),
-      })
-    )
-  })
+  // Must not emit the "connection" event on this tick.
+  expect(connectionListener).toHaveBeenCalledTimes(0)
+
+  await waitForNextTick()
```
kettanaito, line 28: Also modifying this compliance test to ensure the connection event is emitted on the next tick, not the current one.

kettanaito:
## Released: v0.26.8 🎉

This has been released in v0.26.8!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.26.8)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.26.8)

---

PR #522, opened 2024-03-18
fix: support Readable as the request body

mikicho:


src/interceptors/Socket/MockSocket.ts
```diff
@@ -34,7 +34,7 @@ export class MockSocket extends net.Socket {
   public write(...args: Array<unknown>): boolean {
     const [chunk, encoding, callback] = normalizeWriteArgs(args as WriteArgs)
     this.options.write(chunk, encoding, callback)
-    return false
+    return true
```
kettanaito, line 37: How does this affect the fix?

I'm hesitant to return `true` because it means that the chunk has been sent. However, this method will be called for request bodies that we forward to the interceptor first. Semantically, we are buffering the request chunks so returning `false` looks more appropriate.

test/modules/http/compliance/http-req-write.test.ts
```diff
@@ -93,6 +94,31 @@ it('writes Buffer request body', async () => {
   expect(await text()).toEqual(expectedBody)
 })
 
+it('writes Readable request body', async () => {
```
kettanaito, line 97:
```suggestion
it('supports Readable as the request body', async () => {
```

test/modules/http/compliance/http-req-write.test.ts
```diff
@@ -6,7 +6,8 @@ import http from 'node:http'
 import express from 'express'
 import { HttpServer } from '@open-draft/test-server/http'
 import { ClientRequestInterceptor } from '../../../../src/interceptors/ClientRequest'
-import { waitForClientRequest } from '../../../helpers'
+import { sleep, waitForClientRequest } from '../../../helpers'
+import { Readable } from 'node:stream'
```
kettanaito, line 10: Nit: could you please move this import to the top of the file? Thanks!

src/interceptors/Socket/MockSocket.ts
mikicho: We can emit drain instead in `onRequestBody/End` but this would be unnatural as well. 
The more "right" thing to do is to return true/false according to the `this.requestStream.push(chunk)` in the `onRequestBody`

kettanaito: My concern is that this is public facing `.write()` while `this.requestStream` is an internal stream we use to feed to the HTTP parser. Returning `true` will _notify the consumer_ (the user/test) that the chunk has been written while it hasn't been. It really gets buffered in the most direct sense of the word. I don't think returning `true` is correct here.

mikicho: We need either return `true` or emit `drain` event

kettanaito: Got it. Okay, so let's return `true` since it's the path of least resistance. If it ever causes issues, we iterate. Thanks for pointing this out.

kettanaito:
I've opened https://github.com/mswjs/interceptors/pull/527 as a feature branch from the socket interceptor branch (yours points to the same branch so it's a bit problematic to iterate on).

PR #527, opened 2024-03-21
fix: support Readable as request body

kettanaito:
- Closes #522


src/interceptors/Socket/MockSocket.ts
```diff
@@ -34,7 +34,7 @@ export class MockSocket extends net.Socket {
   public write(...args: Array<unknown>): boolean {
     const [chunk, encoding, callback] = normalizeWriteArgs(args as WriteArgs)
     this.options.write(chunk, encoding, callback)
-    return false
+    return true
```
kettanaito, line 37: @mikicho, let's return `true`.

kettanaito:
Resolved the tests, let's merge this!

PR #530, opened 2024-03-26
fix: use global "crypto.randomUUID()"

### LINKED ISSUE #470: Replace "uuid" with "crypto.randomUUID()"

kettanaito:
- Fixes #470

### DIFF
--- a/src/crypto-shim.ts
+++ b/src/crypto-shim.ts
@@ -1,0 +1,3 @@
+import crypto from 'node:crypto'
+
+export { crypto }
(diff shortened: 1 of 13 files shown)


kettanaito:
## `crypto` global

`crypto` has become global since Node v21. In v18, it's still behind the experimental flag (although REPL loads it as global, which is somewhat confusing). 

We will use a shim on the bundler level to inject `crypto` from `node:crypto` and to have the source code referencing the `crypto` global everywhere. Once we migrate to v20+ as the minimal supported, we can drop the shim without touching the source.

kettanaito:
## Released: v0.26.12 🎉

This has been released in v0.26.12!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.26.12)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.26.12)

---

PR #532, opened 2024-03-26
feat: treat unhandled interceptor exceptions as 500 responses

### LINKED ISSUE #516: Treat exceptions as 500 responses instead of a "Failed to fetch" error
Right now, any error that occurs inside of an interceptor is wrapped in a [network error](https://github.com/mswjs/interceptors/blob/ac4f22b1319401d946746d5ca34f6f98846465f1/src/interceptors/fetch/index.ts#L205). Depending on the system that is receiving the error, this can cause problems.

In my case, I'm running asserts inside of an interceptor, like this:

```js
server.use(
    http.get(createApiUrl(API_ENDPOINTS.folderItems), async ({ params, request }) => {
        assert.strictEqual(params.folder_id, folderId);

        const requestUrl = new URL(request.url);
        assert.strictEqual(requestUrl.searchParams.get("limit"), customOptions.limit);
        assert.strictEqual(requestUrl.searchParams.get("marker"), customOptions.marker);

        return HttpResponse.json(mockResponse);
    })
);
```

When any of the asserts fail, Mocha just displays "Failed to fetch", and that really destroys my productivity while debugging.

Not to mention, any custom properties on the thrown error aren't accessible on the wrapped error. That means the nice output that assertion errors typically create in the console are lost.

## Suggestions

1. Just throw the original error. I'm not sure what benefit wrapping the error in another error provides, but it seems like throwing the original error would be more useful.
2. If you must wrap the original error, at least append the original error message after "Failed to fetch", so we can see "Failed to fetch: My original error message".

kettanaito:
- Fixes #516 

Much like request handlers in MSW, everything you write in your interceptors is _from the server's perspective_. If an exception happens on the server, it's usually handled gracefully as the 500 error response to the client. This consistency is nice for predictability but also aligned behavior across the low-level Interceptors and high-level MSW. 

## Changes

- Any exceptions thrown in the `request` listeners are now coerced to `500 Unhandled Exception` mocked error response to emulate server suffering an exception.
- To mock request errors, do `request.respondWith(Response.error())`.


kettanaito:
## Released: v0.28.0 🎉

This has been released in v0.28.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.28.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.28.0)

---

PR #540, opened 2024-03-30
fix(ClientRequest): spread ClientRequest object argument to fix "follow-redirects"

mikicho:


src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.ts
```diff
@@ -210,7 +210,7 @@ export function normalizeClientRequestArgs(
   // Handle a given "RequestOptions" object as-is
   // and derive the URL instance from it.
   else if (isObject(args[0])) {
-    options = args[0] as any
+    options = { ... args[0] as any }
```
kettanaito, line 213: The problem with this, is that spreading an object will not keep its getters/setters/symbols/etc. I suspect this is the reason why the unit tests are failing now:

```diff
 FAIL  src/interceptors/ClientRequest/utils/normalizeClientRequestArgs.test.ts > handles [RequestOptions, callback] input
AssertionError: expected { method: 'POST', …(7) } to deeply equal { method: 'POST', …(5) }

- Expected
+ Received

  Object {
+   "_defaultAgent": Agent {
+     "_events": Object {
+       "free": [Function anonymous],
+       "newListener": [Function maybeEnableKeylog],
+     },
+     "_eventsCount": 2,
+     "_maxListeners": undefined,
+     "_sessionCache": Object {
+       "list": Array [],
+       "map": Object {},
+     },
+     "defaultPort": 443,
+     "freeSockets": Object {},
+     "keepAlive": false,
+     "keepAliveMsecs": 1000,
+     "maxCachedSessions": 100,
+     "maxFreeSockets": 256,
+     "maxSockets": Infinity,
+     "maxTotalSockets": Infinity,
+     "options": Object {
+       "noDelay": true,
+       "path": null,
+     },
+     "protocol": "https:",
+     "requests": Object {},
+     "scheduling": "lifo",
+     "sockets": Object {},
+     "totalSocketCount": 0,
+     Symbol(kCapture): false,
+   },
+   "agent": Agent {
+     "_events": Object {
+       "free": [Function anonymous],
+       "newListener": [Function maybeEnableKeylog],
+     },
+     "_eventsCount": 2,
+     "_maxListeners": undefined,
+     "_sessionCache": Object {
+       "list": Array [],
+       "map": Object {},
+     },
+     "defaultPort": 443,
+     "freeSockets": Object {},
+     "keepAlive": false,
+     "keepAliveMsecs": 1000,
+     "maxCachedSessions": 100,
+     "maxFreeSockets": 256,
+     "maxSockets": Infinity,
+     "maxTotalSockets": Infinity,
+     "options": Object {
+       "noDelay": true,
+       "path": null,
+       "rejectUnauthorized": undefined,
+     },
+     "protocol": "https:",
+     "requests": Object {},
+     "scheduling": "lifo",
+     "sockets": Object {},
+     "totalSocketCount": 0,
+     Symbol(kCapture): false,
+   },
    "headers": Object {
      "Content-Type": "text/plain",
    },
    "host": "mswjs.io",
    "method": "POST",
    "origin": "https://mswjs.io",
    "path": "/resource",
    "protocol": "https:",
  }
```

Looks like some properties from the options are not preserved, which would break compatibility. 

@mikicho, do you happen to have a link to the commit that made the change from `{ ...args[0] }` to `args[0]`? I wonder in which context that was made.

kettanaito:
Thanks, @mikicho! I've moved your changes to https://github.com/mswjs/interceptors/pull/541 and fixed the failing tests.

PR #542, opened 2024-03-31
fix(MockHttpSocket): exhaust .write() callbacks for mocked requests

kettanaito:
- The socket now correctly emits the `finish` event for mocked responses.
- The socket now calls all the `.write()` callbacks for mocked responses. 
- The socket now correctly emits the `end` event when the readable ends. 
- The socket now flushes all the write calls (callbacks) if someone starts reading `this.requestStream` ReadableStream. We still flush any present write calls in `.respondWith()` (nothing could've read the request stream up until that point). 

## Todos

- [x] Fix the event order: https://github.com/mswjs/interceptors/pull/542#discussion_r1545794943


src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -205,6 +207,16 @@ export class MockHttpSocket extends MockSocket {
...
     this.responseType = 'mock'
 
+    // Exhaust the "socket.write()" callbacks to
+    // transition the Writable into the right state
+    // (i.e. "request body written" state). There's nowhere
+    // to actually write the data so we disregard it.
+    // It has been used by the request parser to construct
+    // a Fetch API Request instance representing this request.
+    for (const [_, __, writeCallback] of this.writeBuffer) {
+      writeCallback?.()
```
kettanaito, line 217: The actual fix. 

When ClientRequest writes the data to the socket, the `.write()` method of the MockSocket just buffers the chunks in `this.writeBuffer`. It never actually does any writes (we cannot do that until we know if the connection is possible) and so it never called any write callbacks (such as the `onFinish` callback).

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -187,6 +188,7 @@ export class MockHttpSocket extends MockSocket {
       .on('prefinish', () => this.emit('prefinish'))
       .on('finish', () => this.emit('finish'))
       .on('close', (hadError) => this.emit('close', hadError))
+      .on('end', () => this.emit('end'))
```
kettanaito, line 191: We never forwarded the `end` event of the original socket. I think this should be added.

src/interceptors/Socket/MockSocket.ts
```diff
@@ -46,13 +50,6 @@ export class MockSocket extends net.Socket {
...
     this.options.read(chunk, encoding)
-
-    if (chunk !== null) {
-      this.emit('data', chunk)
-    } else {
-      this.emit('end')
-    }
-
-    return true
+    return super.push(chunk, encoding)
```
kettanaito, line 53: Call `super.push()` for the `.push(null)` to correctly transition the socket into the end state:

- emit `end` event
- set `readable` to `false`
- set `readableEnded` to `true`

See: https://github.com/nodejs/node/blob/3a456c6db802b5b25594d3a9d235d4989e9f7829/lib/internal/streams/readable.js#L1696

src/interceptors/Socket/MockSocket.ts
```diff
@@ -22,6 +22,10 @@ export class MockSocket extends net.Socket {
     super()
     this.connecting = false
     this.connect()
+
+    this._final = (callback) => {
+      callback(null)
```
kettanaito, line 27: Implement the `_final` method so socket writes transition the socket to the right "final" state:

- set `writableFinished` to `true`
- emit `finish` event

See: https://github.com/nodejs/node/blob/3a456c6db802b5b25594d3a9d235d4989e9f7829/lib/internal/streams/writable.js#L907

> `this._final()` is called on `prefinish`.

test/modules/http/compliance/http-req-write.test.ts
```diff
@@ -145,3 +145,35 @@ it('calls the callback when writing an empty Buffer', async () => {
...
+  request.on('finish', finishListener)
+  request.end()
+
+  await waitForClientRequest(request)
+
+  expect(prefinishListener).toHaveBeenCalledTimes(1)
+  expect(finishListener).toHaveBeenCalledTimes(1)
+})
+
+it('emits "finish" for a mocked request', async () => {
```
kettanaito, line 163: @mikicho, this is the failing write finish test you sent me. It's passing now.

src/interceptors/Socket/MockSocket.test.ts
```diff
@@ -150,43 +165,74 @@ it('calls the "read" on "socket.read(null)"', () => {
...
+
+  socket.push('hello')
+  socket.push(null)
+
+  expect(socket.readable).toBe(true)
+  expect(socket.readableEnded).toBe(false)
+
+  // Read the data to free the buffer and
+  // make Socket emit "end".
+  socket.read()
```
kettanaito, line 216: This was a broken test. We have to call `.read()` to exhaust the readable stream so it emits the `end` event.

kettanaito:
Hi, @mikicho 👋 This is the fix for the `finish` event not being emitted for mocked requests. Can you please give it a look and see if you spot anything off? Thanks.

kettanaito:
## Test run

For some reason, CI didn't trigger on this pull request. Sharing the test runs below.

### Unit tests

```sh
pnpm test:unit
```

```sh
 Test Files  24 passed (24)
      Tests  143 passed (143)
   Start at  12:51:24
   Duration  11.62s
```

### Integration tests (Node.js)

```sh
pnpm test:node
```

```sh
 Test Files  1 failed | 85 passed | 1 skipped (87)
      Tests  1 failed | 277 passed | 2 skipped (280)
   Start at  12:52:02
   Duration  37.81s
```

> The failing test can be ignored. It's `modules/http/compliance/http-errors.test.ts > forwards EHOSTUNREACH error for a bypassed request` and it fails only on my machine (it never emits that precise error for me).

src/interceptors/ClientRequest/MockHttpSocket.ts
mikicho: Maybe we now can/should return false in the `socker.write`?

mikicho:
I just tested it, and it fixed some tests, YaY.
However, I encountered an interesting time-based edge case that our tests caught.
I'm not sure if this is the expected behavior, but somehow, Nock preserves the same event order as Node.
Anyway, if you think about something quickly, it should be cool; if not, we may revisit it later.

Scenario A - quick abort:
```js
const http = require('http')
const sinon = require('sinon')

interceptor.on('request', async function rootListener({ request }) {
  request.respondWith(new Response('OK!'))
})
// interceptor.apply()

const req = http.request('http://example.com')

req.on('finish', () => {
  setTimeout(() => {
    req.abort()
  }, 10)
})
req.end()

setTimeout(() => {
  console.log('unmocked');
  console.log(emitSpy.args.map(i => i[0]));
}, 500)
```
> ➜  nock git:(Michael/fetch-support) ✗ node a.js
unmocked
[ 'socket', 'prefinish', 'finish', 'response', 'close', 'abort' ]
➜  nock git:(Michael/fetch-support) ✗ node a.js
mocked
[ 'socket', 'prefinish', 'finish', 'response', 'close', 'abort' ]

Scenario B - abort after long time:
```js
const http = require('http')
const sinon = require('sinon')

interceptor.on('request', async function rootListener({ request }) {
  request.respondWith(new Response('OK!'))
})
// interceptor.apply()

const req = http.request('http://example.com')

req.on('finish', () => {
  setTimeout(() => {
    req.abort()
  }, 1000)
})
req.on('error', err => {
  // console.log(err);
})
req.end()

setTimeout(() => {
  console.log('unmocked');
  console.log(emitSpy.args.map(i => i[0]));
}, 5000)
```
> ➜  nock git:(Michael/fetch-support) ✗ node a.js
mocked
[ 'socket', 'prefinish', 'finish', 'response', 'close', 'abort' ]
➜  nock git:(Michael/fetch-support) ✗ node a.js
unmocked
[ 'socket', 'prefinish', 'finish', 'abort', 'error', 'close' ]
➜  nock git:(Michael/fetch-support) ✗

For small abort timeout, we emit `response` event that Node.js doesn't.

src/interceptors/ClientRequest/MockHttpSocket.ts
mikicho: Now I see we are calling the write callback after the response, which causes a deadlock:

```js
const http = require('http')
const sinon = require('sinon')

interceptor.on('request', async function rootListener({ request }) {
  console.log('before');
  await request.arrayBuffer()
  console.log('never get here');
  request.respondWith(new Response('OK!'))
})
interceptor.apply()

const reqWriteCallback = sinon.spy()

const req = http.request(
  {
    host: 'example.com',
    method: 'POST',
    path: '/',
    port: 80,
  },
  res => {
    console.log(2)
    res.on('end', () => {
      console.log(3);
    })
    // Streams start in 'paused' mode and must be started.
    // See https://nodejs.org/api/stream.html#stream_class_stream_readable
    res.resume()
  },
)

req.write('mamma mia', null, () => {
  console.log(1);
  reqWriteCallback()
  req.end()
})
```

kettanaito:
@mikicho, we should improve event compatibility but it's out of scope of this change. The order of abort/error/close events is also off compared to Node.js. Let's tackle those separately. 

For now, some XHR tests are failing and I need to look at why. Fail in the base feature branch also, so unlikely to be relevant to these changes.

mikicho:
@kettanaito Agreed.

> For now, some XHR tests are failing and I need to look at why. Fail in the base feature branch also, so unlikely to be relevant to these changes.

I saw this failing test and expected them to reflect in one of Nock's, but it didn't (although I skipped on around ~40 for now). It surprised me a little.

Can you please take a look at this before merging this PR? https://github.com/mswjs/interceptors/pull/542#discussion_r1545794943

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: Are you talking about the response _event_ when you mention "response"?

mikicho: I mean we call the `write` callback inside the `responseWith` function. sorry for the confusion.

kettanaito: Got it! Thanks for clarifying. 

In your example, this is the order I'd expect:

```
1
before
never get here
2
3
```

Is that the expected order?

mikicho: > Is that the expected order?

TL;DR:  I think yes.

It's a design choice. If I understand you correctly, you would rather the whole request body to be ready for the user in the interceptor so that they can do `await request.arrayBuffer` without a fear of deadlock like this.
We can also say that the body isn't guaranteed to be ready, and they user needs to take care of this scenario by itself (I think they can with `request.arrayBuffer().then(..)`)

kettanaito: Actually, not quite. The request event is emitted as soon as the request's headers are sent. So the request body may still be streaming. If the user decides to read it, whichever body reading method will return a Promise, and the entire request listener would have to wait for that promise. 

I wonder if that conflicts with the write callbacks. As in, in order to read the request body, the socket has to call the callbacks of .write(), and we are calling those only as a part of `respondWith()`. 

That shouldn't be the case though. The "request body" the user is reading in the request listener is our internal request body buffer where we buffer the chunks pushed.

kettanaito:
The failing test has been resolved in the base feature branch. I forgot to migrate the new interceptor to use a fixed request ID header vs the hard-coded `x-request-id` string, which broke the XHR -> ClientRequest interceptors deduplication and resulted in extra request/response events on the interceptor.

src/interceptors/ClientRequest/MockHttpSocket.ts
mikicho: > That shouldn't be the case though. The "request body" the user is reading in the request listener is our internal request body buffer where we buffer the chunks pushed.

But the internal request body never ends because we never call the write callback, which ends the request body. (we never do: `this.requestStream.push(null)`)

kettanaito: Not sure we should be doing this. Need to look more into this behavior but it's the actual request that calls `.end()` on the writable stream. Perhaps the issue is that we are not translating that to `this.requestStream.push(null)`.

kettanaito: I've pushed a fix in 7be0242029db25f9d3638ffc4a4ea9f4777c73f9. 

Basically, if something attempts to read the request buffer (e.g. `await request.text()` in the interceptor), the `MockHttpSocket` will immediately start flushing all the write callbacks. 

If nothing reads the buffer, the write callbacks will be flushed before the mocked response starts streaming or before the original request is made.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -356,7 +357,13 @@ export class MockHttpSocket extends MockSocket {
          * used as the actual request body (the stream calls "read()").
          * We control the queue in the onRequestBody/End functions.
          */
-        read: () => {},
+        read: () => {
+          // If the user attempts to read the request body,
+          // flush the write buffer to trigger the callbacks.
+          // This way, if the request stream ends in the write callback,
+          // it will indeed end correctly.
+          this.flushWriteBuffer()
```
kettanaito, line 365: If someone starts reading the request stream, flush all the write calls immediately. This way, if the stream ends in the write callback, it will end correctly:

```js
interceptor.on('request', async ({ request }) => {
  // Start reading the request stream before
  // deciding whether to mock it or not.
  // This triggers "read()" on the request's ReadableStream. 
  await request.arrayBuffer()
})

req.write('foo', () => req.end()
```

mikicho:
It fixed more tests!
image

(P.S.: Some tests have revealed that Nock is not behaving properly, and the new interceptor has found the issue!)

kettanaito:
That is the best kind of discovery! Yohoo! Should we merge this?

mikicho:
image

PR #548, opened 2024-04-12
fix(MockHttpSocket): handle response stream errors

kettanaito:
When a mocked response stream receives an error (e.g. `controller.error(error)`), translate that to an error response.


kettanaito:
Alright, so treating those response stream errors as request errors isn't right. Even the [original comment](https://github.com/mswjs/interceptors/pull/515#issuecomment-2050547056) says so.

The problem is, I cannot reproduce the `error` event on response when communicating with a Node.js server. I'm doing this:

```ts
function createErrorStream() {
  return new ReadableStream({
    start(controller) {
      controller.enqueue(new TextEncoder().encode('hello'))
      controller.error(new Error('stream error'))
    },
  })
}

const httpServer = new HttpServer((app) => {
  app.get('/resource', (req, res) => {
    res.pipe(Readable.fromWeb(createErrorStream()))
  })
})
```

```ts
  const request = http.get(httpServer.http.url('/resource'))
  request.on('error', requestErrorListener)
  request.on('response', (response) => {
    console.log('RESPONSE!')

    response.on('error', (error) => {
      console.log('ERROR!', error)
    })
  })
```

The `response` event is emitted but not the `error` event on the response. 

@mikicho, is this roughly how you are reproducing this in a non-mocked scenario?

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -239,34 +239,47 @@ export class MockHttpSocket extends MockSocket {
...
-        if (httpHeaders.length > 0) {
-          flushHeaders(value)
-          continue
+      } catch (error) {
+        if (error instanceof Error) {
+          // Flush the headers on response stream errors.
+          // This way, the client still receives the "response" event,
+          // and the actual stream error is forwarded as the "error"
+          // event on the http.IncomingMessage instance.
+          flushHeaders()
```
kettanaito, line 269: This will emit the `response` event on the client because it will receive the response headers.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -239,34 +239,47 @@ export class MockHttpSocket extends MockSocket {
...
+
+          // Forward response streaming errors as response errors.
+          /** @todo This doesn't quite do it. */
+          // this.destroy(error)
```
kettanaito, line 273: I'm struggling to find a way to trigger this:

https://github.com/nodejs/node/blob/3a456c6db802b5b25594d3a9d235d4989e9f7829/lib/_http_incoming.js#L239

The `error` event on `http.IncomingMessage` is emitted when the socket is destroyed (implemented in the `._destroy()` method). For some reason, calling `this.destroy()` on the `HttpMockSocket` doesn't trigger that.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -239,34 +239,47 @@ export class MockHttpSocket extends MockSocket {
...
       }
     }
 
     // If the headers were not flushed up to this point,
     // this means the response either had no body or had
     // an empty body stream. Flush the headers.
-    if (httpHeaders.length > 0) {
-      flushHeaders()
-    }
+    flushHeaders()
```
kettanaito, line 282: It's safe to call `flushHeaders()` because it will do nothing if the headers have already been flushed.

kettanaito:
## How this works in non-mocked world

When there's a response stream error, the server throws an error and, usually, handles it as a 500 response:

```
response 500 Internal Server Error
RESPONSE DATA <!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Error: Cannot pipe, not readable<br>
```

This means that erroring the response stream just results in an error thrown. It doesn't get forwarded to ClientRequest or IncomingMessage in any special way, aside from a 500 response arriving to the client. 

@mikicho, is this the same expectation Nock has in this scenario?

kettanaito:
This should be fixed by #532 but this PR still contains viable tests and the fix (flushing response headers on unhandled response stream errors).

kettanaito:
@mikicho, I think this is done! Response stream errors are now handled as 500 error responses. I believe that's what's happening in the actual server as well.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -248,34 +248,53 @@ export class MockHttpSocket extends MockSocket {
...
+          if (httpHeaders.length > 0) {
+            flushHeaders(value)
+            continue
+          }
+
+          // Subsequent body chukns are push to the stream.
+          this.push(value)
         }
+      } catch (error) {
+        if (error instanceof Error) {
```
mikicho, line 273: This is not a blocker, but I think we need to return 500 for any error.
Current implementation hides the error from the user in case it is not extends `Error`

mikicho:
> Is this the same expectation Nock has in this scenario?

No, Nock propagates the error event to the response (`IncmingMessage`).
I believe the scenario you described in [here](https://github.com/mswjs/interceptors/pull/548#issuecomment-2059459338) may not cover the entire picture because Node itself could emit an error for the underlying readable.

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: Since anything can be thrown, how to do propose handling arbitrary data here? 

I'd say thrown errors likely indicate issues, and are handled as 500 responses as such. If a non-error is thrown, it's likely to be an explicit intention of the user. Perhaps actually throwing that would be the right thing to do?

kettanaito:
> I think Nock's behavior provides better DX and is more predictable from the user's POV.

While Nock compatibility is the goal, we should focus on how Node.js behaves here, not Nock. From everything I've tried so far, Node doesn't handle response stream errors in any special way. Errors via `controller.error(e)` simply throw [because by design](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultController/error), they meant to error _if you try to operate with the stream afterward_. 

When piping a regular `Readable`, the source can destroy the stream, and that will propagate as the `error` event of `IncomingMessage`. That looks like a different use case though, and we should handle it as such. Perhaps the confusion here is caused by the fact that `ReadableStream` is not entirely the same behavior-wise as `Readable`, although [from the source](https://github.com/nodejs/node/blob/3a456c6db802b5b25594d3a9d235d4989e9f7829/lib/internal/webstreams/adapters.js#L535), it seems `controller.error(e)` should destroy `Readable`:

kettanaito:
One more argument to preserving non-error exceptions is you can throw a `Response` instance and that will be used as the mocked response: https://github.com/mswjs/msw/blob/43a163b2aab87d124aecde05826729b896ad3484/test/node/rest-api/response/throw-response.node.test.ts#L26. 

This is an intentional design to support Remix-like response throwing.

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: I completely agree! I revisited by previous point and found it wrong. I'm alining the uncaught exceptions handling in #555, and then will update this branch to migrate the implementation from `NodeClientRequest.ts`.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -248,34 +251,41 @@ export class MockHttpSocket extends MockSocket {
...
+          }
+
+          // Subsequent body chukns are push to the stream.
+          this.push(value)
         }
+      } catch (error) {
+        // Coerce response stream errors to 500 responses.
+        // Don't flush the original response headers because
+        // unhandled errors translate to 500 error responses forcefully.
+        this.respondWith(createServerErrorResponse(error))
```
kettanaito, line 279: Treating all the response stream errors as 500 error responses! Consistency.

kettanaito:
I will merge this one. If I missed anything, please let me know, we can iterate on it in the follow-up pull request. Thanks!

kettanaito:
> We take a user-side error and convert it to 500 error although the "server" was ok.

But that's the point, it wasn't. When mocking, your request listener _is_ the server. If there's an exception there, it's equivalent to the actual server getting an exception on its runtime, which is often handled as a 500 error response.

Note that this has no effect on the bypassed requests. 

> In other words, one can ask when the user would expect us to emit an error event for the response object:

Technically, when the response stream is destroyed. That's the only occasion per spec when `IncomingMessage` emits the error event. Calling `controller.error(e)`, somehow, doesn't trigger the stream's destruction. 

Here's the only scenario where I think the error will be emitted correctly:

```js
http.get(url, (res) => {
  res.on('error', () => console.log('Error!'))
  res.destroy(new Error('Do not need this anymore')
})
```

Destroying the IncomingMessage stream directly will emit the error event. But since you don't have access to the IncomingMessage in the request listener, you can't destroy the response that way.

mikicho:
> When mocking, your request listener is the server

While this is true, in this case, I see the error as coming from Node.js itself, not from the server. I think it would provide better DX.
Anyway, I think this is not a blocker, and we should release it and see what the users think.

PR #549, opened 2024-04-13
fix: add "address()" on mock Socket

mikicho:


kettanaito:
Sharing our conversation from Discord:

> WDYT about something like realSocket property in the MockSocket class? so in the address function we can do something like: if (mocked) {..} else this.realSocket.address() 

That's tempting but I would recommend against it. `realSocket` introduces a _state_ and I'd rather not have any state (or as little as possible state). 

Instead, we know the mocked and passthrough contexts because we have methods `respondWith()` and `passthrough()`. Those methods can set the `address` function on the socket instance based on the context:

```js
passthrough() {
  const realSocket = createConnection()
  // Fro actual connections, infer the address function.
  this.address = realSocket.address
}

respondWith(response) {
  this.address = () => addressFromResponse(response)
}
```

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -374,6 +374,7 @@ export class MockHttpSocket extends MockSocket {
       credentials: 'same-origin',
       // @ts-expect-error Undocumented Fetch property.
       duplex: canHaveBody ? 'half' : undefined,
+      // @ts-ignore
```
kettanaito, line 377: This isn't something we should ignore. We likely have the wrong @types/node in the project so `Readable.toWeb()` is not present. The fix would be to find that type, not to add ts-ignore.

kettanaito: You can also ignore these errors for now since they aren't directly related to the bugfix!

src/interceptors/Socket/MockSocket.ts
```diff
@@ -35,6 +35,19 @@ export class MockSocket extends net.Socket {
...
+    // @ts-ignore
+    if (this.responseType === 'mock') {
+      return {
+        address: '::1', // should be this.connectionOptions.host?
+        family: 'IPv4',
+        port: 1,
+      }
+    }
+
+    return super.address()
```
kettanaito, line 48: This will have incorrect (empty?) value for bypass scenarios because `MockHttpSocket` has no connection information. 

I proposed to infer this from the original socket but now I wonder, `socket.address()` is generally available throughout the socket's life-cycle, yes? Like, you can grab the address at any point in time after constructing the socket. Hmm.

test/modules/http/compliance/http.test.ts
```diff
@@ -163,3 +163,40 @@ it('mocks response to a non-existing host', async () => {
...
+
+  await expect(connectPromise).resolves.toEqual({
+    address: '::1',
+    family: 'IPv4',
+    port: 1,
+  })
+});
+
+
+it.only('returns real socket address', async () => {
```
kettanaito, line 188: Oops, accidental `it.only` committed! Sorry, don't have an eslint for this.

src/interceptors/ClientRequest/MockHttpSocket.ts
mikicho: I just had the wrong version of node types. I reinstall deps.

mikicho:
@kettanaito I fixed this as suggested. I still think we need to separate the socket-related logic from the HTTP-specific one if we want to support more protocols, but that can wait for another PR.

I have a couple of questions before it's ready for merge from my side:

1. Where should the tests I added be?
2. What do you think about current fake values? Do you have a better idea?

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -124,6 +124,7 @@ export class MockHttpSocket extends MockSocket {
    */
   public passthrough(): void {
     const socket = this.createConnection()
+    this.address = () => socket.address()
```
kettanaito, line 127: Technically, this would be safer:

```suggestion
    this.address = socket.address.bind(socket)
```

In case `this` in the `address()` points to the socket instance, which I think it does. Let's preserve the context.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -205,6 +206,13 @@ export class MockHttpSocket extends MockSocket {
       return
     }
 
+    // Return fake address information for the socket.
+    this.address = () => ({ 
+      address: '0.0.0.0',
```
kettanaito, line 211: What do you think about setting `127.0.0.1` as the fake address? Seeing localhost can hint people that this address is, indeed, mocked.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -205,6 +206,13 @@ export class MockHttpSocket extends MockSocket {
...
+      family: 'IPv4',
```
kettanaito, line 212: I wonder if there's any info in `connectionOptions` that we can use to set the `family`? 🤔 

Yeah, there's `.family` and `.host`. Do you think we should respect those?

```ts
this.address = () => {
  return {
    address: this.connectionOptions?.host || '127.0.0.1',
    family: this.connectionOptions?.family || 'IPv6',
    port: this.connectionOptions?.port || 0
  }
}
```

We can also add automated tests for the cases when explicit connection options are set. Granted, this is likely better tested in the unit tests on `MockHttpSocket`. You can leave the existing `http.*` tests as-is, those are useful too!

test/modules/http/compliance/http.test.ts
```diff
@@ -163,3 +163,40 @@ it('mocks response to a non-existing host', async () => {
...
+it('returns mocked socket address', async () => {
+  interceptor.on('request', async ({ request }) => {
+    request.respondWith(new Response())
+  })
+
+  const connectPromise = new DeferredPromise<object>()
+  const request = http.get('http://example.com')
+  request.once('socket', socket => {
+    socket.once('connect', () => {
+      connectPromise.resolve(socket.address())
```
kettanaito, line 176: It would be great to also have a test for `socket.address()` _before_ the `connect` is emitted. I believe it should still be set?

src/interceptors/ClientRequest/MockHttpSocket.ts
mikicho: I think both ways preserve the context, but I agree that your solution is more explicit.

src/interceptors/ClientRequest/MockHttpSocket.ts
mikicho: I think the address info resolved in the socket. I didn't verfiy this, but the `connectionOptions` includes only the `port`.
Also, please note the `host` !== `address`.

test/modules/http/compliance/http.test.ts
mikicho: It seems like the lookup happens after the socket event is emitted.
```js
request.once('socket', socket => {
  console.log(socket.address()); // prints empty object {}
});
```

We probably need to cover the `lookup` event as well.

kettanaito: Yes, this is correct. ClientRequest emits `socket` as soon as the socket is assigned, then the socket goes through its life-cycle, emitting `lookup` and `connect`.

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: You're right. Let's infer the `port` only for now then?

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -298,7 +306,7 @@ export class MockHttpSocket extends MockSocket {
   }
 
   private mockConnect(): void {
-    this.emit('lookup', null, '::1', 6, this.connectionOptions.host)
+    this.emit('lookup', null, '127.0.0.1', 4, this.connectionOptions.host)
```
kettanaito, line 309: I think setting `this.address` should be done in `mockConnect()` for mocked responses. This will couple the `.address()` function withe actual address we emit in the lookup even so those two stay consistent (right now, they are not).

```
const address = {
  host: '...',
  family: '...',
  port: '...'
}
this.address = () => address

this.emit('lookup', null, address.host, address.family, this.connectionOptions.host)
```

mikicho: Wonderful! I'm ashamed that I didn't think about it myself.😅

kettanaito (approved):
This looks great! I've pushed a minor polishing to the tests, this should be ready for merge after CI passes.

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: You think of a ton of things that never crossed my mind 🙌 The power of collaboration.

PR #552, opened 2024-04-17
fix: give `node` export condition higher priority

phryneas:
in case of mixed environments that look for both
"node" and "browser" export conditions

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,8 +19,12 @@
       "default": "./lib/node/index.js"
     },
     "./ClientRequest": {
-      "browser": null,
       "types": "./lib/node/interceptors/ClientRequest/index.d.ts",
+      "node": {
+        "require": "./lib/node/interceptors/ClientRequest/index.js",
+        "import": "./lib/node/interceptors/ClientRequest/index.mjs"
+      },
+      "browser": null,
       "require": "./lib/node/interceptors/ClientRequest/index.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -56,15 +60,23 @@
       "default": "./lib/browser/interceptors/WebSocket/index.js"
     },
     "./RemoteHttpInterceptor": {
-      "browser": null,
       "types": "./lib/node/RemoteHttpInterceptor.d.ts",
+      "node": {
```
kettanaito, line 64: Should we have "default" here too? I wonder when does it fall through "default" 🤔 If it's nether `require` not `import`? So, in the current state, it will find no "node.default", and fallback to the "/node.default" on the export path level?

kettanaito (approved):
This is fantastic! 🚀

package.json
phryneas: I think any tool that will apply an export condition like "node" will also apply "import" or "require". I would probably not pollute too much now until someone comes around the corner with an example where it's missing.

kettanaito:
## Released: v0.28.1 🎉

This has been released in v0.28.1!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.28.1)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.28.1)

---

PR #555, opened 2024-04-17
fix: unify server error response across interceptors

kettanaito:
- If you throw a `Response` instance in a request listener, that instance will be used as a mocked response. 
- If you throw anything else, it will be coerced to a 500 error response, similar to how an actual server would handle an uncaught exception. 

> If you want to emulate a request error, do `request.respondWith(Response.error())`.

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestProxy.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestProxy.ts
@@ -4,6 +4,7 @@ import { XMLHttpRequestEmitter } from '.'
 import { toInteractiveRequest } from '../../utils/toInteractiveRequest'
 import { emitAsync } from '../../utils/emitAsync'
 import { XMLHttpRequestController } from './XMLHttpRequestController'
+import { createServerErrorResponse } from '../../utils/responseUtils'
 
 export interface XMLHttpRequestProxyOptions {
   emitter: XMLHttpRequestEmitter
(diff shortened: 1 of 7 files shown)


src/interceptors/XMLHttpRequest/XMLHttpRequestProxy.ts
```diff
@@ -99,31 +99,30 @@ export function createXMLHttpRequestProxy({
...
-                status: 500,
-                statusText: 'Unhandled Exception',
-                headers: {
-                  'Content-Type': 'application/json',
-                },
-              }
+          if (resolverResult.error instanceof Error) {
+            // Treat unhandled exceptions in the "request" listener
+            // as 500 server errors.
+            xhrRequestController.respondWith(
```
kettanaito, line 105: The argument of treating exceptions as 500 responses is good, see: #516. TL;DR, if exceptions are preserved, they will be handled by the higher request client. Not only different clients handle those exceptions differently, but that harms the observability of the thrown error. 

I also think that treating unhandled exceptions as 500 responses is more similar to the actual server, and you write these request listeners from the server's respective.

src/interceptors/XMLHttpRequest/XMLHttpRequestProxy.ts
```diff
@@ -99,31 +99,30 @@ export function createXMLHttpRequestProxy({
...
+          }
 
-          /**
-           * @todo Consider forwarding this error to the stderr as well
-           * since not all consumers are expecting to handle errors.
-           * If they don't, this error will be swallowed.
-           */
-          // xhrRequestController.errorWith(resolverResult.error)
+          // Otherwise, error the request with the thrown data.
+          xhrRequestController.errorWith(resolverResult.error)
```
kettanaito, line 125: The argument for translating any other (non-error) data to request errors is a bit weaker. There should be a way to trigger request errors.

Now that I think about it, we do have that way! It's this:

```js
request.respondWith(Response.error())
```

This translates to a request error across different clients. **I think we need to remove non-error handling changes and treat non-errors as 500 responses anyway.**

kettanaito:
[This](https://github.com/mswjs/interceptors/pull/555#discussion_r1568820802) is what I eventually ended up with. I think it makes the most sense and makes the exception handling predictable.

mikicho:
This is great! 
More context from discussion in Nock: https://github.com/nock/nock/issues/2211

kettanaito:
## Released: v0.28.3 🎉

This has been released in v0.28.3!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.28.3)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.28.3)

---

PR #556, opened 2024-04-17
fix(MockHttpSocket): forward tls socket properties

kettanaito:
Forwards the following TLS socket properties onto the `MockHttpSocket` (for both mocked and passthrough responses):

- `encrypted`
- `authorized`
- `getSession()`
- `getProtocol()`
- `isSessionReused()`

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -169,6 +169,28 @@ export class MockHttpSocket extends MockSocket {
       }
     }
 
+    // Forward TLS Socket properties onto this Socket instance
+    // in the case of a TLS/SSL connection.
+    if (Reflect.get(socket, 'encrypted')) {
+      const tlsProperties = [
+        'encrypted',
+        'authorized',
+        'getProtocol',
+        'getSession',
+        'isSessionReused',
+      ]
+
+      tlsProperties.forEach((propertyName) => {
+        Object.defineProperty(this, propertyName, {
+          enumerable: true,
+          get: () => {
+            const value = Reflect.get(socket, propertyName)
+            return typeof value === 'function' ? value.bind(socket) : value
+          },
+        })
+      })
+    }
+
     socket
       .on('lookup', (...args) => this.emit('lookup', ...args))
       .on('connect', () => {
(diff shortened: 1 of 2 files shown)


src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -331,9 +353,22 @@ export class MockHttpSocket extends MockSocket {
...
+        'session',
+        this.connectionOptions.session ||
+          Buffer.from('mock-session-renegotiate')
+      )
       this.emit('session', Buffer.from('mock-session-resume'))
+
+      Reflect.set(this, 'encrypted', true)
+      // The server certificate is not the same as a CA
+      // passed to the TLS socket connection options.
+      Reflect.set(this, 'authorized', false)
```
kettanaito, line 368: @mikicho, note that the `.authorized` must be set to `false`. It's false in a bypassed HTTPS communication as well. 

It will only be true if the certificate you pass to socket options is the same one that has signed the server you are requesting. That's false in most cases.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -169,6 +169,28 @@ export class MockHttpSocket extends MockSocket {
       }
     }
 
+    // Forward TLS Socket properties onto this Socket instance
+    // in the case of a TLS/SSL connection.
+    if (Reflect.get(socket, 'encrypted')) {
```
mikicho, line 174: Can we check the protocol instead?

kettanaito: We can but that should be the discerning property to tell apart Socket from TLSSocket, per Node.js docs:

> *This may be used to distinguish TLS sockets from regular `net.Socket` instances*.
> — [Node.js docs / TLS](https://nodejs.org/docs/latest/api/tls.html#tlssocketencrypted)

I don't mind checking the protocol but if the HTTPS request does have a TLSSocket, it's constructed incorrectly. 

Is there a particular test case that fails with this check?

mikicho: Oh, it's a weird choice by Node. But if it's official, I'm in.

mikicho (approved):

PR #558, opened 2024-04-18
test: add socket timeout tests

kettanaito:
Also adds a `this.destroyed` check in `passthrough()` and `respondWith()` so they'd have no effect if the socket has been destroyed (e.g. aborted or timed out).


kettanaito:
@mikicho, I think you should know about the difference I outlined above. See if Nock understands those two different timeouts and treats them the same. It's rather tricky and undocumented, and I wouldn't be surprised if it didn't.

kettanaito:
I also suspect that `AbortSignal.timeout(n)` will act similarly to setting a timeout. But I expect it to result in the "abort" event on the request.

mikicho:
> See if Nock understands those two different timeouts...

Nock [is aware of these two](https://github.com/nock/nock/?tab=readme-ov-file#delay-the-response), and it actually uses it for throwing a timeout error immediately if the timeout is bigger than what is defined in the interceptor. 

 This is a nifty feature that MSW (or mswjs/intrceptors) should consider.

P.S: when I [tried to implement](https://github.com/nock/nock/pull/2602#issuecomment-2043662191) this for the `fetch` API, it turned out that the `delay connection` feature doesn't make much sense because Fetch doesn't provide this level of granularity.

kettanaito:
> This is a nifty feature that MSW (or mswjs/intrceptors) should consider.

Interceptors do not decide that for you so they shouldn't do anything on top of what Node.js does. Node.js doesn't do anything to the request, not even destroys it. It only emits the `timeout` event for you do handle that timeout in whichever fashion you want. I think it should stay that way. 

> delay connection feature doesn't make much sense because Fetch doesn't provide this level of granularity.

Yep, I remember that!

mikicho:
> Interceptors do not decide that for you so they shouldn't do anything on top of what Node.js does

Nock does not change Node.js behavior here:
1. with "auto timeout,": timeout event emits immediately.
2. Without: Nock waits X seconds and emits a `timeout` event.

PR #561, opened 2024-04-20
fix(MockHttpSocket): set tls socket properties on init

mikicho:
It seems like Node.js set these properties on `TLSSocket` init, before the `connect` event ([source](https://github.com/nodejs/node/blob/main/lib/_tls_wrap.js#L574)).

With umocked `https`:
```js
https.get('https://example.com')
  .on('socket', socket => {
    console.log(socket.authorized); // false
    console.log(socket.encrypted);  // true
  }); 
```


kettanaito (approved):
Awesome! 🚀

Issue #431, opened 2023-09-18
Replace "request.respondWith()" with a Controller


kettanaito:
Instead of adding the `respondWith()` custom method on the Fetch API `Request` representation of the intercepted request, expose that method _alongside_ the request in the listener argument object.

```js
interceptor.on('request', (event) => {
  const { request } = event
  event.respondWith(new Response())
})
```

This also resembles the Service Worker's `fetch` event a bit more closely. Not that it has ever the intention but it's a nice little detail.

kettanaito:
## Request controller interface

```ts
interface RequestController {
  // Instruct the interceptor to perform the request as-is.
  passthrough(): void

  // Use the given response as the mock response for the request.
  respondWith(response: Response): void

  // Use the given error as the request error.
  errorWith(error?: Error): void

  // Abort the request.
  // In some clients (ClientRequest) it's synonymous to ".errorWith()",
  // in others (XMLHttpRequest) it's different. 
  abort(): void
}
```

The request controller is exposed in the `request` listener argument:

```js
interceptor.on('request', ({ request, controller }) => {
  controller.respondWith(new Response('hello world'))
})
```

The request controller can only handle the request _once_. Once any of its methods are called, calling methods on it again must throw an error.

kettanaito:
## Released: v0.33.0 🎉

This has been released in v0.33.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.33.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.33.0)

Make sure to always update to the latest version (`npm i @mswjs/interceptors@latest`) to get the newest features and bug fixes.

---

_Predictable release automation by [@ossjs/release](https://github.com/ossjs/release)_.

PR #562, opened 2024-04-21
fix: add RequestError class

kettanaito:
Right now, there's no way to prevent an error from being coerces as a 500 error response. In some cases, like [here](https://github.com/mswjs/msw/blob/43a163b2aab87d124aecde05826729b896ad3484/src/core/utils/request/onUnhandledRequest.ts#L36), we do want to throw an error for it to error the request instead of producing an error response. 

This adds a new `RequestError` class. If an instance of `RequestError` is thrown within a request listener, the respective request will error with the error thrown.


kettanaito:
I've implemented support for throwing `Response.error()` in https://github.com/mswjs/interceptors/commit/32fdba9b2178621e584de2b4300d8d4ea579a738.

kettanaito:
This is potentially solved by adding support for throwing `Response.error()` (https://github.com/mswjs/interceptors/pull/563).

kettanaito:
Not needed.

#566 allows you to opt-out from the default exception handling. In the future, `controller.errorWith()` will allow you to mock request errors explicitly (see #431).

PR #566, opened 2024-04-26
feat: add "unhandledException" event

kettanaito:
- Related to https://github.com/mswjs/msw/pull/2135

## Changes

Adds a new event to the HTTP events map called `unhandledException`. If the listener for that event is provided, whenever an unhandled exception occurs in the `request` listener, the `unhandledException` listener will be called (if present).

This allows the client to opt-out from the default handling of exceptions (producing 500 error responses), as well as do so conditionally. If the `unhandledException` listener does nothing, the default handling kicks in. 

**Note:** The `unhandledException` listener receives a _readonly_ `request` instance (non-interactive request). To affect the request, it receives an additional `controller` object with methods `.respondWith()` and `.errorWith()`. 

> `request.respondWith()` operates on a promise basis and by the time exception occurs, that promise has been awaited already. Besides, I very much wish to move away from patching the `Request` instance with `.respondWith()` to the designated `controller` argument for the `request` listener as well. I see this as an incremental change in that direction. 

```js
interceptor.on('unhandledException', ({ error, request, requestId, controller }) => {
  controller.respondWith(new Response('fallback response'))
  // controller.errorWith(new Error('fallback error'))
})
```

### DIFF
--- a/src/interceptors/fetch/index.ts
+++ b/src/interceptors/fetch/index.ts
@@ -88,18 +88,26 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
         )
       }
 
-      const respondWith = (response: Response): Response => {
-        // Clone the mocked response for the "response" event listener.
-        // This way, the listener can read the response and not lock its body
-        // for the actual fetch consumer.
-        const responseClone = response.clone()
+      const responsePromise = new DeferredPromise<Response>()
 
-        this.emitter.emit('response', {
-          response: responseClone,
-          isMockedResponse: true,
-          request: interactiveRequest,
-          requestId,
-        })
+      const respondWith = (response: Response): void => {
+        this.logger.info('responding with a mock response:', response)
+
+        if (this.emitter.listenerCount('response') > 0) {
+          this.logger.info('emitting the "response" event...')
+
+          // Clone the mocked response for the "response" event listener.
+          // This way, the listener can read the response and not lock its body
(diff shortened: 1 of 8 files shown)


src/interceptors/fetch/index.ts
```diff
@@ -88,18 +88,26 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
...
-        // for the actual fetch consumer.
-        const responseClone = response.clone()
-
-        this.emitter.emit('response', {
-          response: responseClone,
-          isMockedResponse: true,
-          request: interactiveRequest,
-          requestId,
-        })
+      const responsePromise = new DeferredPromise<Response>()
```
kettanaito, line 91: I had to rewrite the Fetch interceptor to return a deferred Promise so we can expose functions like `respondWith()` and `errorWith()` to asynchronous contexts (the `unhandledException` listener). The listener cannot "return" anything to the pending Fetch promise, so instead it controls the deferred promise, which is always returned.

kettanaito:
## Released: v0.29.0 🎉

This has been released in v0.29.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.29.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.29.0)

---

mikicho:
This is really nice! thanks!

PR #565, opened 2024-04-24
fix(ClientRequest): decode uri for username and password before setting basic authentication header

kettanaito:
- Fixes #564

PR #567, opened 2024-04-26
fix(ClientRequest): decode uri for username and password before setting basic authentication header

### LINKED ISSUE #564: ClientRequestInterceptor incorrectly encodes basic authentication header
Our application uses Axios to send basic authentication requests to a 3rd party application. This integration works ok until we add the ClientRequestInterceptor (for logging purposes). The interceptor seems to apply url encoding when it should not.

Authorization header without interceptor (after base64 decoding): `user@xyz.com:password`
Authorization header with interceptor (after base64 decoding): `user%40xyz.com:password`

mdesousa:
- Closes #565
- Fixes #564

### DIFF
--- a/test/third-party/axios.test.ts
+++ b/test/third-party/axios.test.ts
@@ -125,6 +125,7 @@ it('preserves "auth" (Authorization)', async () => {
 
   // Construct an Axios request with "auth".
   await axios.get(httpServer.http.url('/books'), {
+    adapter: 'http',
     auth: {
       // Use an email address as the username.
       // This must NOT be encoded.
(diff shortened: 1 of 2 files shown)


kettanaito (approved):
This is incredible! Thank you so much, @mdesousa 👏

test/third-party/axios.test.ts
```diff
@@ -125,6 +125,7 @@ it('preserves "auth" (Authorization)', async () => {
 
   // Construct an Axios request with "auth".
   await axios.get(httpServer.http.url('/books'), {
+    adapter: 'http',
```
kettanaito, line 128: Oh, so the trick was using the HTTP adapter. By default, Axios is using XHR, right? Good catch.

kettanaito:
Welcome to the contributors, @mdesousa 🎉

Issue #575, opened 2024-06-01
Nock compatibility


mikicho:
- [x] <s>should be safe to call in the middle of a request</s>
- [x] socket emits `connect` and `secureConnect` events @kettanaito (edge case - https://github.com/mswjs/interceptors/pull/515#issuecomment-2067702330)
- [ ] Request with `Expect: 100-continue` triggers continue event (https://github.com/mswjs/interceptors/pull/599) @mikicho 
- [x] Fetch: follow redirect responses #626 

## Transfer Encoding @mikicho 
support `transfer-encoding: chunked`

https://github.com/mswjs/interceptors/pull/596

- [x] should pass filteringPath options
- [x] should pass filteringRequestBody options
- [x] match body with form multipart
- [x] records and replays objects correctly
- [x] records and replays correctly with filteringRequestBody
- [x] records and replays gzipped nocks correctly
- [x] records and replays the response body

## preemtive timeout
We decided to remove this feature from Nock. It not relevant for `fetch`. 
We should consider if/how we want to support this for `httpClient`. this is a nice feature.
<s>
- [x] emits a timeout - with setTimeout
- [x] emits a timeout - with options.timeout
- [x] emits a timeout - with Agent.timeout
- [x] emits a timeout - options.timeout takes precedence over Agent.timeout
</s>

## Headers
- [x] folds duplicate headers the same as Node
- [x] when keys are duplicated, is evaluated once per input field name, in correct order

## Recorder
- [ ] does not record requests from previous sessions
- [x] logs recorded objects

kettanaito:
Thanks for putting this down, @mikicho! Let's split these between ourselves and finish the interceptor. Please, if you are working on a task, put your GitHub handle next to it so I'd know, and I will do the same.

As usual, we can start with adding a test for the thing, verifying if it's not passing already, and then opening pull requests to #515. Does that sound good to you?

kettanaito:
> socket emits connect and secureConnect

This should already be the case:

- Original requests: https://github.com/mswjs/interceptors/pull/515/files#diff-94fc27cca818a0da8d402e251763d5dbf450d62c3cde37ceef7ea8494a324ddeR210-R214
- Mocked requests: https://github.com/mswjs/interceptors/pull/515/files#diff-94fc27cca818a0da8d402e251763d5dbf450d62c3cde37ceef7ea8494a324ddeR375-R380

[...]

PR #576, opened 2024-06-01
test(http): add "connect" and "secureConnect" tests

kettanaito:
- Related to #575


kettanaito:
Hi, @mikicho. Could you please take a brief look at these tests? They confirm that the Socket interceptor correctly emits the `connect` and `secureConnect` events for both mocked and bypassed scenarios. Thanks!

mikicho:
Looks great!
The tests in Nock fail due to https://github.com/mswjs/interceptors/pull/515#issuecomment-2067702330, which we can tackle later.

PR #577, opened 2024-06-04
feat(WebSocket): add connection "info" to the "connection" event payload

DanielleHuisman:
Currently, there is no way to access the full list of protocols supported by the WebSocket client. `WebSocketOverride` only exposes the first value ([relevant line](https://github.com/mswjs/interceptors/blob/main/src/interceptors/WebSocket/WebSocketOverride.ts#L54)).

This PR exposes the client protocols in the `connection` event, so tests can verify the client provided the right protocols.

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,10 +322,11 @@ intereceptor.on('connection', ({ client }) => {
 
 The `connection` event exposes the following arguments:
 
-| Name     | Type                                                      | Description                                                      |
-| -------- | --------------------------------------------------------- | ---------------------------------------------------------------- |
-| `client` | [`WebSocketClientConnection`](#websocketclientconnection) | An object representing a connected WebSocket client instance.    |
-| `server` | [`WebSocketServerConnection`](#websocketserverconnection) | An object representing the original WebSocket server connection. |
+| Name        | Type                                                      | Description                                                      |
+| ----------- | --------------------------------------------------------- | ---------------------------------------------------------------- |
+| `client`    | [`WebSocketClientConnection`](#websocketclientconnection) | An object representing a connected WebSocket client instance.    |
+| `server`    | [`WebSocketServerConnection`](#websocketserverconnection) | An object representing the original WebSocket server connection. |
+| `protocols` | <code>string &#124; string[] &#124; undefined</code>      | The protocols supported by the WebSocket client.                 |
 
 ### `WebSocketClientConnection`
 
--- a/README.md
+++ b/README.md
@@ -322,11 +322,11 @@ intereceptor.on('connection', ({ client }) => {
 
 The `connection` event exposes the following arguments:
 
-| Name        | Type                                                      | Description                                                      |
-| ----------- | --------------------------------------------------------- | ---------------------------------------------------------------- |
-| `client`    | [`WebSocketClientConnection`](#websocketclientconnection) | An object representing a connected WebSocket client instance.    |
-| `server`    | [`WebSocketServerConnection`](#websocketserverconnection) | An object representing the original WebSocket server connection. |
-| `protocols` | <code>string &#124; string[] &#124; undefined</code>      | The protocols supported by the WebSocket client.                 |
+| Name     | Type                                                      | Description                                                                         |
+| -------- | --------------------------------------------------------- | ----------------------------------------------------------------------------------- |
+| `client` | [`WebSocketClientConnection`](#websocketclientconnection) | An object representing a connected WebSocket client instance.                       |
+| `server` | [`WebSocketServerConnection`](#websocketserverconnection) | An object representing the original WebSocket server connection.                    |
+| `info`   | `object`                                                  | Additional WebSocket connection information (like the original client `protocols`). |
 
 ### `WebSocketClientConnection`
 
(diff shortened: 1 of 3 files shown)


kettanaito:
Hi, @DanielleHuisman. Thanks for opening this. 

> WebSocketOverride only exposes the first value

That is correct. `WebSocketOverride` complies with `WebSocket`, which complies with WHATWG WebSocket, where the `protocol` property on the `WebSocket` interface is a single string representing the protocol that was eventually chosen by the instance ([source](https://websockets.spec.whatwg.org/#interface-definition)).

In other words, this is also true:

```js
let ws = new WebSocket(url, ['ws:', 'wss:'])
// You cannot access the full list of protocols anymore.
ws.protocol // "ws:"
```

Can you tell me more about your use case to access the list of protocols?

kettanaito:
Also a bit of information about how the `string[]` input of `protocols` is handled:

> Each string in the array is a subprotocol name. The connection will only be established if the server reports that it has selected one of these subprotocols.

As I understand it, the handshake request's `Sec-WebSocket-Protocol` header will combine all the protocols you've specified on the client so then the server can report which protocol it ended up using. 

Does this mean that you want to access the list of protocols to _choose_ between them within the event handler? I'm still a bit confused how that's beneficial. Would love to learn more.

DanielleHuisman:
I think you're mixing up URL protocols and WebSocket subprotocols. The URL protocol (`ws:` or `wss:`) is specified in the `url` parameter of the `WebSocket` constructor. The WebSocket subprotocols are specified in the `protocols` parameter of `WebSocket`. The naming of the second parameter is a bit unfortunate.

WebSocket subprotocols are user defined strings that could represent the capabilities of the client and server. As you mentioned, the client and server need to agree on the subprotocol they are using and this single value is stored in `WebSocket.protocol`. An example of these subprotocols can be found in [section 1.2 of the RFC](https://datatracker.ietf.org/doc/html/rfc6455#section-1.2). More information can be found in the RFC when searching for `Sec-WebSocket-Protocol` or `subprotocol`.

I have some WebSocket client code that conditionally chooses a subprotocol, so I would like to verify this behaviour in unit tests using MSW. Currently, I can only verify the first value sent by the client in `WebSocket.protocol`. My PR attempts to expose the full list of subprotocols provided by the client.

Now that I've looked at it in more detail, I think the default value (`ws`) in `WebSocketOverride` ([line](https://github.com/mswjs/interceptors/blob/main/src/interceptors/WebSocket/WebSocketOverride.ts#L54)) does not match the specification. I believe if no subprotocols are provided by the client, `WebSocket.protocol` is an empty string, because the client and server both agree not to use a subprotocol. Ofcourse assuming the server doesn't require the use of a subprotocol.

I hope this clarifies the intentions of my PR.

kettanaito:
> I think you're mixing up URL protocols and WebSocket subprotocols. 

🤦 I totally am mixing those up. Thanks for clarifying! 

> My PR attempts to expose the full list of subprotocols provided by the client.

I think this is reasonable then. Let me think of the exact way we can do that. 

> Now that I've looked at it in more detail, I think the default value (ws) in WebSocketOverride ([line](https://github.com/mswjs/interceptors/blob/main/src/interceptors/WebSocket/WebSocketOverride.ts#L54)) does not match the specification.

Yes, you are right. Please, could you open a new pull request to fix this? I would be grateful!

README.md
```diff
@@ -322,10 +322,11 @@ intereceptor.on('connection', ({ client }) => {
...
-| -------- | --------------------------------------------------------- | ---------------------------------------------------------------- |
-| `client` | [`WebSocketClientConnection`](#websocketclientconnection) | An object representing a connected WebSocket client instance.    |
-| `server` | [`WebSocketServerConnection`](#websocketserverconnection) | An object representing the original WebSocket server connection. |
+| Name        | Type                                                      | Description                                                      |
+| ----------- | --------------------------------------------------------- | ---------------------------------------------------------------- |
+| `client`    | [`WebSocketClientConnection`](#websocketclientconnection) | An object representing a connected WebSocket client instance.    |
+| `server`    | [`WebSocketServerConnection`](#websocketserverconnection) | An object representing the original WebSocket server connection. |
+| `protocols` | <code>string &#124; string[] &#124; undefined</code>      | The protocols supported by the WebSocket client.                 |
```
kettanaito, line 329: Maybe something like `info` would be more long-lasting? I don't want to prematurely optimize here but I'm thinking just in case we need to expose additional connection info in the future, this may be a safer API to extend. 

Wdyt?

kettanaito: `client.protocols`? 🤔

DanielleHuisman: Again, all good with me. I don't have strong opinions on how to expose those values 😉

kettanaito:
I've nested the protocols under `info` key of the event payload. This is how you can access it now:

```js
interceptor.on('connection', ({ info, client }) => {
  info.protocols // ['one', 'two']
  client.socket.protocol // 'one'
})
```

kettanaito (approved):
This is great! Thank you for improving the library, @DanielleHuisman! 👏

kettanaito:
## Released: v0.30.0 🎉

This has been released in v0.30.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.30.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.30.0)

---

PR #591, opened 2024-07-03
feat(ClientRequest): await response listeners before emitting the "end" response event

### LINKED ISSUE #569: Await all "response" listeners to finish
hi, we have code that looks like this:

```
const interceptor = new BatchInterceptor(...); // console.log to print requests and responses
const p1 = fetch('x.com');
const p2 = fetch('y.com');
await Promise.all([p1, p2]);
```

if there is an error with one of the requests, the process ends
in this scenario, we would like to make sure that responses are logged. however, there may be a small delay and the process terminates before the response is logged.

is there a way to wait until the interceptor has invoked the `response` hook for all pending requests?

kettanaito:
- Fixes #569 

## Changes

- ClientRequest: Awaits all the `response` event listeners before finishing the response (emitting its `end` event). This affects both mocked and original responses (the approach is different for each).

### DIFF
--- a/src/interceptors/ClientRequest/NodeClientRequest.ts
+++ b/src/interceptors/ClientRequest/NodeClientRequest.ts
@@ -368,20 +368,20 @@ export class NodeClientRequest extends ClientRequest {
 
         this.logger.info('emitting the custom "response" event...')
 
-        until(async () => {
-          await emitAsync(this.emitter, 'response', {
-            response: responseClone,
-            isMockedResponse: true,
-            request: capturedRequest,
-            requestId,
-          })
+        const responseListenersPromise = emitAsync(this.emitter, 'response', {
+          response: responseClone,
+          isMockedResponse: true,
+          request: capturedRequest,
+          requestId,
+        })
 
-          // Mark the response as complete once all the response
-          // listeners have finished.
-          this.response.emit('end')
-
+        responseListenersPromise.then(() => {
           this.logger.info('request (mock) is completed')
         })
+
+        // Defer the end of the response until all the response
+        // event listeners are done (those can be async).
+        this.deferResponseEndUntil(responseListenersPromise, this.response)
 
         return this
       }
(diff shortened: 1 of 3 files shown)


src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -419,7 +434,7 @@ export class NodeClientRequest extends ClientRequest {
         const firstClone = cloneIncomingMessage(response)
         const secondClone = cloneIncomingMessage(response)
 
-        this.emit('response-internal', secondClone)
+        this.emit('response-internal', secondClone, firstClone)
```
kettanaito, line 437: We need to expose the original `IncomingMessage` of the original response if we wish to delay its `end` event emission. This modifies the listener call signature of the `response-internal` event.

src/interceptors/ClientRequest/NodeClientRequest.ts
```diff
@@ -643,4 +658,23 @@ export class NodeClientRequest extends ClientRequest {
     // @ts-ignore "agent" is a private property.
     this.agent?.destroy?.()
   }
+
+  private deferResponseEndUntil(
```
kettanaito, line 662: Maybe premature but I'm creating a private method to help us reuse the same deferring logic for both mocked and original responses (the target `IncomingMessage` is different in case of both).

test/features/events/response.test.ts
```diff
@@ -69,6 +69,7 @@ beforeAll(async () => {
 })
 
 afterEach(() => {
+  interceptor.removeAllListeners('response')
```
kettanaito, line 72: Incredibly flaky test poorly written. Shame on me. 

I'm resetting the response listener because they apparently leak between the tests through the shared interceptor instance. 

I've been scoping request/response listeners within individual tests for some time to prevent this from happening but this is a really old test.

src/interceptors/ClientRequest/NodeClientRequest.ts
kettanaito: I will likely repurpose this for the Socket interceptor. It would have its own implementation of awaiting the response listeners.

kettanaito:
## Released: v0.31.0 🎉

This has been released in v0.31.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.31.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.31.0)

---

PR #596, opened 2024-07-06
fix(ClientRequest): use `ServerResponse` to build the HTTP response string

mikicho:
> Context: this allows us to support mocked responses with `Transfer-Encoding` and, potentially, other response transformations. 

I still need to give meaningful names to some variables, but it works as expected.

The idea here is that MSW acts as a Node.js server and uses the `ServerResponse` class to build the final chunks that we later push to our `socket`.

The only caveat I found is that we are maybe TOO close to the Node.js implementation 😅. For example, Nock's tests found that `ServerResponse` automatically returns `connection` and `date` headers, which we now add to the response. 
I think we can remove this default behavior, but I first want to know how you see it.
@kettanaito

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -5,7 +5,7 @@ import {
   type ResponseHeadersCompleteCallback,
 } from '_http_common'
 import { IncomingMessage, ServerResponse } from 'node:http'
-import { Readable, Writable } from 'node:stream'
+import { Readable } from 'node:stream'
 import { invariant } from 'outvariant'
 import { INTERNAL_REQUEST_ID_HEADER_NAME } from '../../Interceptor'
 import { MockSocket } from '../Socket/MockSocket'
@@ -272,22 +272,25 @@ export class MockHttpSocket extends MockSocket {
     // if it hasn't been flushed already (e.g. someone started reading request stream).
     this.flushWriteBuffer()
 
-    const w = new Writable({
-      write: (chunk, encoding, callback) => {
-        this.push(chunk, encoding);
-        callback()
-      },
-    })
-    const fakeResponse = new ServerResponse(new IncomingMessage(new net.Socket()))
-    // @ts-expect-error Node has test for this: https://github.com/nodejs/node/blob/10099bb3f7fd97bb9dd9667188426866b3098e07/test/parallel/test-http-server-response-standalone.js#L32
-    fakeResponse.assignSocket(w)
-    fakeResponse.statusCode = response.status;
-    fakeResponse.statusMessage = response.statusText;
+    const fakeResponse = new ServerResponse(new IncomingMessage(this))
+
+    // Node has test for this: https://github.com/nodejs/node/blob/10099bb3f7fd97bb9dd9667188426866b3098e07/test/parallel/test-http-server-response-standalone.js#L32
+    fakeResponse.assignSocket(
+      new MockSocket({
+        write: (chunk, encoding, callback) => {
+          this.push(chunk, encoding)
+          callback?.()
+        },
+        read() {},
+      })
+    )
+    fakeResponse.statusCode = response.status
+    fakeResponse.statusMessage = response.statusText
 
     // Get the raw headers stored behind the symbol to preserve name casing.
     const headers = getRawFetchHeaders(response.headers) || response.headers
     for (const [name, value] of headers) {
-      fakeResponse.setHeader(name, value);
+      fakeResponse.setHeader(name, value)
     }
 
     if (response.body) {
@@ -301,7 +304,7 @@ export class MockHttpSocket extends MockSocket {
             fakeResponse.end()
             break
           }
-          fakeResponse.write(value);
+          fakeResponse.write(value)
         }
       } catch (error) {
         // Coerce response stream errors to 500 responses.
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -4,8 +4,8 @@ import {
   type RequestHeadersCompleteCallback,
   type ResponseHeadersCompleteCallback,
 } from '_http_common'
-import { STATUS_CODES } from 'node:http'
-import { Readable } from 'node:stream'
+import { IncomingMessage, ServerResponse } from 'node:http'
+import { Readable, Writable } from 'node:stream'
 import { invariant } from 'outvariant'
 import { INTERNAL_REQUEST_ID_HEADER_NAME } from '../../Interceptor'
 import { MockSocket } from '../Socket/MockSocket'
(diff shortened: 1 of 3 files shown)


src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -272,36 +272,22 @@ export class MockHttpSocket extends MockSocket {
...
-        }\r\n`
-      )
-    )
+    const w = new Writable({
+      write: (chunk, encoding, callback) => {
+        this.push(chunk, encoding);
+        callback()
+      },
+    })
+    const fakeResponse = new ServerResponse(new IncomingMessage(new net.Socket()))
```
kettanaito, line 281: Any chance this can also handle compression for us?

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -272,36 +272,34 @@ export class MockHttpSocket extends MockSocket {
     // if it hasn't been flushed already (e.g. someone started reading request stream).
     this.flushWriteBuffer()
 
-    const httpHeaders: Array<Buffer> = []
+    // Create a `ServerResponse` instance to delegate HTTP message parsing,
+    // Transfer-Encoding, and other things to Node.js internals.
+    const serverResponse = new ServerResponse(new IncomingMessage(this))
```
kettanaito, line 277: I passed `this` socket instance as the socket of the `IncomingMessage` used here. I think it should be more error-proof in case this server response decides to terminate the socket (rightfully), for example.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -272,36 +272,34 @@ export class MockHttpSocket extends MockSocket {
...
+    /**
+     * Assign a mock socket instance to the server response to
+     * spy on the response chunk writes. Push the transformed response chunks
+     * to this `MockHttpSocket` instance to trigger the "data" event.
+     * @note Providing the same `MockSocket` instance when creating `ServerResponse`
+     * does not have the same effect.
+     * @see https://github.com/nodejs/node/blob/10099bb3f7fd97bb9dd9667188426866b3098e07/test/parallel/test-http-server-response-standalone.js#L32
+     */
+    serverResponse.assignSocket(
+      new MockSocket({
```
kettanaito, line 288: I used the `MockSocket` we already have instead of `Writable`. While sockets extend streams, they implement additional methods on top. I know Node.js uses a simple `Writable` in their test but we have to acknowledge that test case is controlled. Our use cases are going to be _everything_ out there. Using an actual socket-compatible instance is crucial here.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -272,36 +272,34 @@ export class MockHttpSocket extends MockSocket {
...
     // Get the raw headers stored behind the symbol to preserve name casing.
     const headers = getRawFetchHeaders(response.headers) || response.headers
     for (const [name, value] of headers) {
-      httpHeaders.push(Buffer.from(`${name}: ${value}\r\n`))
-    }
-
-    // An empty line separating headers from the body.
-    httpHeaders.push(Buffer.from('\r\n'))
-
-    const flushHeaders = (value?: Uint8Array) => {
```
kettanaito, line 294: I _really_ like that we can drop this now. Fantastic!

test/modules/http/response/http-response-transfer-encoding.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        headers: { 'Transfer-Encoding': 'chunked' },
+      })
+    )
+  })
+
+  const request = http.get('http://localhost')
+  const { res, text } = await waitForClientRequest(request)
+
+  expect(res.statusCode).toBe(200)
+  expect(res.headers).toHaveProperty('transfer-encoding', 'chunked')
```
kettanaito, line 36: Added a few assertions that the client also receives the right response headers.

kettanaito (approved):
This is nothing short of fantastic. Great job, @mikicho 👏

kettanaito:
Regarding the `Connection` and `Date` headers being added to the mocked response, I'm not sure. I'm leaning toward removing them. Only the headers the developer themselves provided in the mocked response instance should be set. 

@mikicho, what do you think about this?

src/interceptors/ClientRequest/MockHttpSocket.ts
mikicho: Ok. I used a new one because I was afraid of infinity loops, but if it makes sense to you I'm ok with this

kettanaito:
From the RFC9110, the [Date](https://www.rfc-editor.org/rfc/rfc9110.html#section-6.6.1) section:

> An origin server with a clock (as defined in [Section 5.6.7](https://www.rfc-editor.org/rfc/rfc9110.html#http.date)) MUST generate a Date header field ...
> An origin server without a clock MUST NOT generate a Date header field.

So the addition of the `Date` (and I assume by extension the `Connection` header) seems to be dictated by the specification. Node.js just follows it. 

Now, it's debatable who should be in charge of this. I think it's may be okay for us to leave those headers if they mean better spec compatibility out of the box. It'd even be good, in fact. As long as we are sure that the developer can override (or remove) those headers if they want to.

kettanaito:
My opinion on this jumps back and forth every time I think about it. 

The connection and date headers make sense from the _server's_ perspective. Request listeners are not technically servers. You can emulate one if you want, but you are not required to. 

Moreover, `new Response()` doesn't attach those headers by default. It may be confusing to see them added by the interceptor when the developer had no intention of having them added. 

**Let's remove those server headers**. Keep the existing behavior, make the mocked response a WYSIWYG.

mikicho:
> I'm leaning toward removing them.

Me, too. I think we are a mocking tool and the DX > compatibility. It may be confusing because, in our case, compatibility is usually better than DX (it prevents terrible production errors).
Anyway, in this specific case, I think we should disable this and pass the headers as-is.

kettanaito:
@mikicho, agreed. Pushed the fix and also adjusted the test for the empty mocked response to assert that the received `IncomingMessage` headers are empty.

src/interceptors/ClientRequest/MockHttpSocket.ts
mikicho: Hmm.. good point.
I thought we would get this OOTB from Node.js when the response goes back to the client.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -272,36 +272,46 @@ export class MockHttpSocket extends MockSocket {
...
+    serverResponse.assignSocket(
+      new MockSocket({
+        write: (chunk, encoding, callback) => {
+          this.push(chunk, encoding)
+          callback?.()
+        },
+        read() {},
+      })
     )
+    serverResponse.statusCode = response.status
```
kettanaito, line 296: Removing these server response headers before the developer-sent headers. They can always set these if they wish to.

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: I hope so! I think we have some tests for compression but I got a new issue about it in MSW recently, which made me think something was still missing. Anyways, we'll see.

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: Should arrive at the infinite loops. This logic is only triggered on `respondWith()`, and that's called by the developer. If I spot any issues with this, will revert back to use an empty socket.

kettanaito: Should arrive at the infinite loops. This logic is only triggered on `respondWith()`, and that's called by the developer. If I spot any issues with this, will revert back to use an empty socket.

kettanaito:
## Released: v0.32.1 🎉

This has been released in v0.32.1!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.32.1)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.32.1)

---

PR #597, opened 2024-07-06
fix(ClientRequest): destroy socket when destroying IncomingMessage

### LINKED ISSUE #449: Propagate socket error event on res.destroy
[According to docs](https://nodejs.org/api/http.html#messagedestroyerror), `response.destroy(err)` emits `error` event.:

> If error is provided, an 'error' event is emitted on the socket and error is passed as an argument to any listeners on the event.

https://github.com/nodejs/node/blob/main/lib/internal/streams/destroy.js#L136

--- 
Nock [handles ](https://github.com/nock/nock/blob/feaa66fa64d24f95937ef759cdd5a7ca07646f1c/lib/intercepted_request_router.js#L125) this by overriding the Socket class and propagate the event to the request.

@kettanaito 
I tried to open a PR for this but couldn't understand how to do this, but I believe I'll be able to with your direction.

kettanaito:
- Fixes #449

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -140,6 +140,11 @@ export class MockHttpSocket extends MockSocket {
     // Normally, we shoud listen to the "close" event but it
     // can be suppressed by using the "emitClose: false" option.
     this.responseParser.free()
+
+    if (error) {
+      this.emit('error', error)
+    }
+
     return super.destroy(error)
   }
 
@@ -153,6 +158,12 @@ export class MockHttpSocket extends MockSocket {
     }
 
     const socket = this.createConnection()
+
+    // If the developer destroys the socket, destroy the original connection.
+    this.once('error', (error) => {
+      socket.destroy(error)
+    })
+
     this.address = socket.address.bind(socket)
 
     // Flush the buffered "socket.write()" calls onto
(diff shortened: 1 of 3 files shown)


src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -153,6 +158,12 @@ export class MockHttpSocket extends MockSocket {
     }
 
     const socket = this.createConnection()
+
+    // If the developer destroys the socket, destroy the original connection.
+    this.once('error', (error) => {
```
kettanaito, line 163: Added this: if the socket gets destroyed by the developer, abort the pending actual socket connection.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -308,6 +319,11 @@ export class MockHttpSocket extends MockSocket {
     serverResponse.removeHeader('connection')
     serverResponse.removeHeader('date')
 
+    // If the developer destroy the socket, gracefully destroy the response.
+    this.once('error', () => {
```
kettanaito, line 323: The same for the mocked responses: if the socket gets destroyed, abort the pending server response.

mikicho (approved):
Great!

src/interceptors/ClientRequest/index.test.ts
```diff
@@ -60,25 +64,23 @@ it('forbids calling "respondWith" multiple times for the same request', async ()
...
-    requestEmitted.resolve()
-    await sleep(10_000)
+    await sleep(1_000)
     request.respondWith(new Response())
   })
 
   const abortController = new AbortController()
   const request = http.get(requestUrl, { signal: abortController.signal })
 
-  await requestEmitted
```
kettanaito, line 73: Awaiting the request listener, for some reason, breaks this test. It's not crucial for the thing we are trying to check so I'm omitting it.

kettanaito:
## Released: v0.32.1 🎉

This has been released in v0.32.1!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.32.1)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.32.1)

---

PR #599, opened 2024-07-06
fix(ClientRequest): support 100 continue flow

mikicho:
- Related to #575 

Relevant comments:
https://github.com/mswjs/interceptors/pull/515#issuecomment-2197566539
https://github.com/mswjs/interceptors/pull/515#issuecomment-2067760131


HonzaMac:
For mine use case, this error occur when using localstack and calling S3 upload. Uploading for first byte take so long (timeouts), that localstack sends back 100. 100 is not handled by undici/native fetch in v20 correctly and it throws error of not handle status code by undici.

test/modules/http/request/http-request-continue.test.ts
```diff
@@ -0,0 +1,40 @@
+import { it, expect, beforeAll, afterAll } from 'vitest'
```
kettanaito, line 1: I think this should be in the compliance tests.

test/modules/http/request/http-request-continue.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+afterAll(async () => {
+  interceptor.dispose()
+  await httpServer.close()
+})
+
+it('Unmocked request with `Expect: 100-continue` triggers continue event', async () => {
+  const body = 'this is the full request body'
+  const request = http.get(httpServer.http.url('/resource'), {
+    headers: { Expect: '100-continue' },
```
kettanaito, line 28: How does this header affect the response resolution?

kettanaito:
Updates the tests, no idea why they are failing only when (1) the interceptor is on; (2) the `request.end()` is nested in the `continue` event callback. Sharing the log outputs for two scenarios (bypass and mocked)

## Bypass

```
SOCKET WRITE POST /resource HTTP/1.1
expect: 100-continue
Host: 127.0.0.1:62344
Connection: close
Transfer-Encoding: chunked


SOCKET EMIT [ 'resume' ]

stdout | modules/http/compliance/http-request-continue.test.ts > emits "continue" event for a request with "100-continue" expect header
SOCKET EMIT [ 'connect' ]
SOCKET CONNECT
SOCKET EMIT [ 'ready' ]
!!![server] added req.on(data)
SOCKET PUSH HTTP/1.1 100 Continue


SOCKET EMIT [ 'data', 'HTTP/1.1 100 Continue\r\n\r\n' ]
REQ CONTINUE
REQ END
!!!! writing request...
SOCKET WRITE 5
SOCKET WRITE 

SOCKET WRITE hello
SOCKET WRITE 

SOCKET WRITE 0


REQ FINISH

[server] req data: hello

SOCKET PUSH HTTP/1.1 200 OK
X-Powered-By: Express
Date: Sun, 08 Sep 2024 14:59:48 GMT
Connection: close
Transfer-Encoding: chunked

5
hello
0


SOCKET EMIT [
  'data',
  'HTTP/1.1 200 OK\r\n' +
    'X-Powered-By: Express\r\n' +
    'Date: Sun, 08 Sep 2024 14:59:48 GMT\r\n' +
    'Connection: close\r\n' +
    'Transfer-Encoding: chunked\r\n' +
    '\r\n' +
    '5\r\n' +
    'hello\r\n' +
    '0\r\n' +
    '\r\n'
]
REQ RESPONSE

SOCKET FINISH

SOCKET EMIT [ 'prefinish' ]
SOCKET EMIT [ 'finish' ]
SOCKET EMIT [ 'close', false ]

SOCKET CLOSE
```

## Mocked

```
SOCKET WRITE POST /resource HTTP/1.1
expect: 100-continue
Host: 127.0.0.1:62777
Connection: close
Transfer-Encoding: chunked


[*] request POST http://127.0.0.1:62777/resource
SOCKET EMIT [ 'resume' ]
SOCKET EMIT [ 'resume' ]
SOCKET EMIT [ 'connect' ]
SOCKET CONNECT
SOCKET EMIT [ 'ready' ]
!!![server] added req.on(data)

SOCKET PUSH HTTP/1.1 100 Continue


SOCKET EMIT [ 'data', 'HTTP/1.1 100 Continue\r\n\r\n' ]
REQ CONTINUE
REQ END
!!!! writing request...
SOCKET WRITE 5
SOCKET WRITE 

SOCKET WRITE hello
REQ BODY! hello true
SOCKET WRITE 

SOCKET WRITE 0


REQ FINISH

// HANGS HERE FOR A WHILE UNTIL TEST TIMESOUT
// BECAUSE "req.on(data)" NEVER EMITS, AND THUS
// THE SERVER NEVER SENDS A RESPONSE.

SOCKET END

SOCKET FINISH

SOCKET EMIT [ 'end' ]
SOCKET EMIT [ 'close' ]
SOCKET EMIT [ 'prefinish' ]
SOCKET EMIT [ 'finish' ]
SOCKET EMIT [ 'close', false ]

SOCKET CLOSE
SOCKET CLOSE
```

JoaquinFernandez:
I'll post here, since I see it is already in conversation for NockE. This issue is also happening for me, when uploading a big file, the client first sends the `Expect: 100 Continue`, before sending the body. This is expected behavior from the http lib.

I'd be happy to help implement a solution, since I don't see how I could work around this elsewhere and it is necessary for me to upgrade to nock v14 (with fetch support, thanks to mswjs/interceptors).

JoaquinFernandez:
Bump

mikicho:
@kettanaito Maybe things changed since you worked on it, but this is working to me:
```js
const { FetchResponse } = require('./lib/node')
const { ClientRequestInterceptor } = require('./lib/node/interceptors/ClientRequest')
const http = require('http');

const interceptor = new ClientRequestInterceptor();
interceptor.apply()

// Simple HTTP server
  if (req.method === 'POST') {
    let body = '';
    req.on('data', chunk => (body += chunk));
    req.on('end', () => {
      res.writeHead(200, { 'Content-Type': 'text/plain' });
      res.end('Received: ' + body);
    });
  } else {
    res.writeHead(200, { 'Content-Type': 'text/plain' });
    res.end('Hello World');
  }
});

server.listen(3001, () => {
  console.log('Server listening on http://localhost:3001');
});

// Example client making a request with 'Expect: 100-continue'
const options = {
  port: 3001,
  method: 'POST',
  headers: {
    'Content-Length': Buffer.byteLength('test body'),
    'Expect': '100-continue'
  }
};

const req = http.request(options, res => {
  let data = '';
  res.on('data', chunk => (data += chunk));
  res.on('end', () => {
    console.log('Response:', data);
    server.close();
  });
});

req.on('continue', () => {
  req.write('test body');
  req.end();
});
```

The problem is if the user wants to return 100 response as mocked response:
```js
interceptor.on('request', ({controller}) => {
  controller.respondWith(new FetchResponse(null, { status: 100 }))
});
```

error:
> Server listening on http://localhost:3001
node:_http_client:542
    emitErrorEvent(req, new ConnResetException('socket hang up'));
                        ^

Error: socket hang up
    at MockHttpSocket.socketOnEnd (node:_http_client:542:25)
    at MockHttpSocket.emit (node:events:530:35)
    at <anonymous>
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
Emitted 'error' event on ClientRequest instance at:
    at emitErrorEvent (node:_http_client:104:11)
    at MockHttpSocket.socketOnEnd (node:_http_client:542:5)
    at MockHttpSocket.emit (node:events:530:35)
    at <anonymous>
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  code: 'ECONNRESET'
}

I continue (🥁) to investigate it.

JoaquinFernandez:
Hi! Any updates on this? Anyway I can help?

kettanaito:
Implemented in #770.

JoaquinFernandez:
First of all, thank you for #770. Moving interception down to `net.connect()`/`tls.connect()` and patching the TCP/TLS wraps is a much bigger undertaking than patching `ClientRequest`, and it shows: requests now go through nearly the whole Node networking stack before we touch anything. I know this thread has been open a while, and I appreciate that the fix came as a proper architectural change rather than a patch on top of the old model.

I tested `0.42.3` against the case this PR was opened for, and I think the 100-continue flow is only half covered.

**What works now:** the request *is* intercepted, which it wasn't before — that alone is a real step forward.

**What still deadlocks:** reading the request body inside the `request` listener. The client withholds the body until it receives `100 Continue`, but the interceptor never sends one — so a listener that awaits the body waits forever, and the client never gets to send it.

```js
import http from 'node:http'
import { HttpRequestInterceptor } from '@mswjs/interceptors/http'

const interceptor = new HttpRequestInterceptor()
interceptor.on('request', async ({ request, controller }) => {
  await request.clone().arrayBuffer() // <- never resolves
  controller.respondWith(new Response('mocked'))
})
interceptor.apply()

const body = 'hello'
const request = http.request({
  host: 'example.invalid', // RFC 2606, can never resolve
  port: 80,
  method: 'POST',
  path: '/upload',
  headers: { Expect: '100-continue', 'Content-Length': Buffer.byteLength(body) },
})

request.on('continue', () => request.end(body))
request.on('response', response => console.log('response:', response.statusCode))
request.on('error', error => console.log('error:', error.code))
setTimeout(() => console.log('still pending after 5s'), 5000)
```

Drop the `await request.clone().arrayBuffer()` line and the same request resolves with `200`. Keep it and nothing happens: the `continue` event never fires.

`@mswjs/interceptors@0.42.3`, Node `v22.22.1`.

**Why this still blocks the original use case:** nock has to read the body to decide which interceptor matches a request, so it hits exactly this deadlock. That's why it currently passes every `Expect: 100-continue` request straight through to the network instead (nock/nock#2877). AWS SDK v3 sets that header on its own for uploads, so in practice S3 tests either hang or silently escape the mock.

So while #770 fixed the architecture, consumers that need the request body still can't handle these requests.

Would you accept a change where the interceptor replies `100 Continue` automatically as soon as a listener accesses the request body, or would you rather expose it explicitly on the controller (e.g. `controller.continue()`) and leave the decision to the consumer? Happy to put up a PR for whichever direction you prefer

PR #602, opened 2024-07-11
fix(ClientRequest): improve rawHeaders recording

mikicho:
@kettanaito The [problem](https://github.com/mswjs/interceptors/pull/598#discussion_r1671313975) was that we didn't unset the `kRestorePatches` symbol on `dispose`

I think my solution may not be "private" enough because I set it to be `configurable`. Alternatively, we can conditionally return undefined by toggling a variable, but it's smelly and error-prone. 
What is the best solution here?

### DIFF
--- a/src/interceptors/ClientRequest/utils/recordRawHeaders.test.ts
+++ b/src/interceptors/ClientRequest/utils/recordRawHeaders.test.ts
@@ -57,6 +57,35 @@ it('records raw headers (Headers / Headers as init', () => {
   ])
 })
 
+it('records raw headers added via ".set()"', () => {
+  recordRawFetchHeaders()
+  const headers = new Headers([['X-My-Header', '1']])
+  headers.set('X-Another-Header', '2')
+
+  expect(getRawFetchHeaders(headers)).toEqual([
+    ['X-My-Header', '1'],
+    ['X-Another-Header', '2'],
+  ])
+})
+
+it('records raw headers added via ".append()"', () => {
+  recordRawFetchHeaders()
+  const headers = new Headers([['X-My-Header', '1']])
+  headers.append('X-My-Header', '2')
+
+  expect(getRawFetchHeaders(headers)).toEqual([
+    ['X-My-Header', '1'],
+    ['X-My-Header', '2'],
+  ])
+})
+
+it('deletes the header when called ".delete()"', () => {
+  const headers = new Headers([['X-My-Header', '1']])
+  headers.delete('X-My-Header')
+
+  expect(getRawFetchHeaders(headers)).toEqual([])
+})
+
 it('records raw headers (Request / object as init)', () => {
   recordRawFetchHeaders()
   const request = new Request(url, {
--- a/src/interceptors/ClientRequest/utils/recordRawHeaders.test.ts
+++ b/src/interceptors/ClientRequest/utils/recordRawHeaders.test.ts
@@ -1,0 +1,130 @@
+// @vitest-environment node
+import { it, expect, afterEach } from 'vitest'
+import {
+  recordRawFetchHeaders,
+  restoreHeadersPrototype,
+  getRawFetchHeaders,
+} from './recordRawHeaders'
+
+const url = 'http://localhost'
+
+afterEach(() => {
+  restoreHeadersPrototype()
+})
+
+it('returns an empty list if no headers were set', () => {
+  expect(getRawFetchHeaders(new Headers())).toEqual([])
+  expect(getRawFetchHeaders(new Headers(undefined))).toEqual([])
+  expect(getRawFetchHeaders(new Headers({}))).toEqual([])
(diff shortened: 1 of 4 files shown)


src/interceptors/ClientRequest/utils/recordRawHeaders.ts
```diff
@@ -46,8 +46,13 @@ export function recordRawFetchHeaders() {
 
       globalThis.Request = OriginalRequest
       globalThis.Response = OriginalResponse
+
+      Object.defineProperty(Headers, kRestorePatches, {
```
kettanaito, line 50: You can use `Reflect.delete()` and delete the symbol altogether. Deleting it doesn't violate the non-configurable property.

test/modules/http/compliance/http-res-raw-headers.test.ts
```diff
@@ -172,3 +172,24 @@ it('preserves raw response headers for unmocked request', async () => {
   )
   expect(res.headers['x-custom-header']).toEqual('Yes')
 })
+
+it('make sure we patch the Headers object correctly after dispose and reapply', async () => {
```
kettanaito, line 176: This likely belongs to the unit test for `ClientRequest/index.test.ts`.

src/interceptors/ClientRequest/utils/recordRawHeaders.ts
mikicho: @kettanaito Seems like the object must be configurable for `deleteProperity` to work
```js
      console.log(Reflect.deleteProperty(Headers, kRestorePatches)) // returns false
      console.log(Reflect.get(Headers, kRestorePatches)) // return some value (should return undefined)
```

kettanaito: Oh, good catch. It's okay to set it configurable from the start so we are able to delete it later.

mikicho: @kettanaito Set it to `configurable`

kettanaito:
Some tests are failing 🤦 Forgot to test on latest v18/v20.

kettanaito:
There's a breaking change in Headers behavior between the minor versions of v18 of Node. The latest v18 works as expected. I will suggest people to update to the latest if they encounter any issues.

src/interceptors/ClientRequest/utils/recordRawHeaders.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+it('records raw headers (Response / Headers as init)', () => {
+  recordRawFetchHeaders()
+  const headers = new Headers([['X-My-Header', '1']])
+  const response = new Response(null, { headers })
+
+  expect(getRawFetchHeaders(response.headers)).toEqual([['X-My-Header', '1']])
+})
+
+it('stops recording once the patches are restored', () => {
```
mikicho, line 124: I'm wondering if we need the test I wrote if we have this test

kettanaito: Unit vs integration. Let's keep yours. It tests that ClientRequest interceptors applies/restores the headers patching. The unit tests don't test that.

kettanaito:
Fixed the compatibility with the latest v18/v20. The tests are finally green again. 

@mikicho, could you please give these changes another look once you have a moment? Thank you.

src/interceptors/ClientRequest/utils/recordRawHeaders.ts
```diff
@@ -105,14 +136,27 @@ export function recordRawFetchHeaders() {
...
+        Reflect.has(args[1].headers, kRawHeaders)
       ) {
-        request.headers[kRawHeaders] = inferRawHeaders(args[1].headers)
+        args[1].headers = args[1].headers[kRawHeaders]
+      }
+
+      const request = Reflect.construct(target, args, newTarget)
+
+      if (typeof args[1] === 'object' && args[1].headers != null) {
+        defineRawHeaders(request.headers, inferRawHeaders(args[1].headers))
```
kettanaito, line 159: Latest Node.js v18 and v20 does _not_ copy Headers symbols (previous versions did!). We need to manually copy the `kRawHeaders` symbol (or set it by creating a new Headers instance) when constructing Request/Response instances.

mikicho:
@kettanaito LGTM

kettanaito:
Forgot to add unit tests for `.append()` and `.set()`. Will add and merge today/tomorrow /cc @mikicho

kettanaito (approved):

kettanaito:
## Released: v0.33.1 🎉

This has been released in v0.33.1!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.33.1)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.33.1)

---

PR #604, opened 2024-07-12
fix(fetch): support `Content-Encoding` response header

### LINKED ISSUE #446: GZIP doesn't work with got
```js
const got = require('got')
const zlib = require('zlib')

  name: 'my-interceptor',
})
interceptor.apply();
  const compressed = zlib.gzipSync('hello')
  request.respondWith(new Response(compressed, {
    status: 200, headers: {
      'X-Transfer-Length': String(compressed.length),
      'Content-Length': undefined,
      'Content-Encoding': 'gzip',
    }
  }))
});

(async function () {
  console.log(1);
  console.log(await got('http://example.test/foo'))
  console.log(2);
})()
```
Specifically, got exits [in this line](https://github.com/sindresorhus/got/blob/v11.8.6/source/core/utils/get-buffer.ts#L9) with error code 0 (ok).
I couldn't catch the error (even with a try-catch), so I'm unsure what's happening.

It does work with native `http`.

May be related to: #308

mikicho:
- Fixes #446
- Related to #575 

Added failing test.

@kettanaito 
1. At first, I thought to adopt [Nock's implementation (inspired by undici)](https://github.com/nock/nock/blob/beta/lib/create_response.js#L24), but IIRC, `interceptors` is cross-platform and AFAIK `zlib` does not exist in the browser. How does the browser decompress the response body? Am I missing something?
2. From what I see, `IncomingMessage` does not decompress the response body, and Node.js leaves this for the HTTP libraries (Axios, got, etc.). So, I think we should follow this behavior and leave the response body compressed.

---

## Summary

- We are not reusing Undici's decompression logic directly because that'd mean redistributing Undici alongside Interceptors. 
- We are using Compression Streams API to decompress GZIP and Deflate both in Node.js and the browser.
- We are using `zlib` and a custom transform stream to support Brotli decompression in Node.js.
- We are ignoring Brotli in the browser due to WASM issues.


test/modules/fetch/intercept/fetch.test.ts
```diff
@@ -329,3 +330,22 @@ it('intercepts an HTTPS PATCH request', async () => {
 
   expect(requestId).toMatch(REQUEST_ID_REGEXP)
 })
+
+it.only('support content-encoding: gzip, br', async () => {
```
kettanaito, line 334: Let's be careful not to let this slip through 👀

kettanaito:
Thanks, @mikicho! 

I believe the answers to both of your questions are in setting up an HTTP server and taking a look what happens to compressed bodies sent from the server. We can request the same endpoint in the browser and in Node.js, and see whether they handle that at all. 

But overall, yes, Interceptors ships browser interceptors, so we can't run things like `zlib` in `FetchInterceptor`. The mental model behind the request listeners (and MSW handlers by extension) is to act as a server route handler. If you need to pull an extra lib to decompress the request body in real life, you need to do the same with Interceptors. Since we are not a convenience library, we don't need to worry about anything outside of that.

test/modules/fetch/intercept/fetch.test.ts
```diff
@@ -329,3 +330,22 @@ it('intercepts an HTTPS PATCH request', async () => {
...
+  const message = 'Lorem ipsum dolor sit amet'
+  const compressed = zlib.brotliCompressSync(zlib.gzipSync(message))
+
+  interceptor.once('request', ({ controller }) => {
+    controller.respondWith(new Response(compressed, {
```
kettanaito, line 339: For mocked responses, you most definitely need a third-party. Interceptors mustn't meddle with your input based on any headers. 

I'm mostly curious about _receiving_ a compressed request in Node.js.

test/modules/fetch/intercept/fetch.test.ts
mikicho: I think there is a lint rule for this. Do you want me to check it out?

test/modules/fetch/intercept/fetch.test.ts
mikicho: > I'm mostly curious about receiving a compressed request in Node.js.

I hope I understand your question correctly but:
Fetch: decompress the response according to the content encoding header.
IncomingMessage: return the body as-is

kettanaito:
Some related stuff I found:

- https://github.com/nodejs/undici/issues/2260

test/modules/fetch/intercept/fetch.test.ts
kettanaito: If this project uses eslint, then let's do that!

kettanaito:
Gave this a try in Node.js:

```js
import fs from 'fs'
import http from 'http'
import zlib from 'zlib'

const stream = fs.createReadStream('./payload.txt')
  // Compress the response body.
  if (req.method === 'GET' && req.url === '/resource') {
    res.statusCode = 200
    res.writeHead(200, {
      'content-type': 'text/plain',
      'content-encoding': 'gzip',
    })
    stream.pipe(zlib.createGzip()).pipe(res)
    return
  }

  res.statusCode = 404
  res.end()
})

server.listen(3000, async () => {
  console.log('Server is running on http://localhost:3000')

  const response = await fetch('http://localhost:3000/resource', {
    headers: { 'Accept-Encoding': 'gzip' },
  })
  console.log(await response.text())
})
```

I receive the file's content decoded. I don't do the compression though.

mikicho:
@kettanaito 
> I receive the file's content decoded. I don't do the compression though.

What do you mean? You do compress the response in the server

test/modules/fetch/intercept/fetch.test.ts
mikicho: We don't use eslint in this repo? 
I mean to this rule: https://www.npmjs.com/package/eslint-plugin-no-only-tests

I also see that `vitest` support this OOTB and our test should fail: https://github.com/vitest-dev/vitest/pull/669

mikicho:
@kettanaito I added an unmocked test which passes.

kettanaito:
Thank you, @mikicho! Will look at this.

kettanaito:
We do need to add `Content-Encoding` handling to the mocked responses in the fetch interceptor. Before proceeding, I've opened an issue in Undici to see if they would be willing to abstract their handling into a utility function we can consume directly: https://github.com/nodejs/undici/issues/3412. That would be the best case scenario. 

If not, we can integrate their handling into Interceptors manually, although that'd be a repetition I'd rather not introduce.

test/modules/fetch/intercept/fetch.test.ts
kettanaito: Got it! That seems to be the case, although we may want to add integration tests for the `ClientRequest` interceptor to verify it. I do believe it's going to expose the response body as-is and it's up to the developer to decompress it. 

The fetch client has decompression built-in, I can confirm it by looking at the [source](https://github.com/nodejs/undici/blob/c1f7d2a9a6556e5a34fdf5ed38c77ce636f1ff02/lib/web/fetch/index.js#L2139-L2161) of Undici.

kettanaito:
## Update

Opened a pull request that utilizes the Compression Streams API to decompress fetch responses: https://github.com/mswjs/interceptors/pull/661. This works for GZIP and Deflate, but has no Brotli support. 

@mikicho, would be curious to hear your thoughts on this one.

kettanaito:
@mikicho, that may be problematic. The environment check is fine, the module difference is problematic. We either need to ship two different `FetchInterceptors` for Node.js and the browser, or somehow hack a way to import different implementation of the decompression. 

That's non-trivial, unfortunately, since we have to support both CJS and ESM at the moment. Dynamic requires will throw when compiling to ESM (imports must be static). We can give this a try. 

The problem remains that the browser-compatible, web-stream-friendly Brotli decompression is non-existent right now. `brotli-wasm` is our best bet, if they agree to fix the export conditions and also bundle for Node.js ESM.

mikicho:
@kettanaito I see.
I see it is a bit hacky, but what about a different `tsconfig` with [path alias](https://www.typescriptlang.org/tsconfig/#paths)?

I like this approach because it may solve other environment-specific code problems that may occur in the future in an elegant way.

kettanaito:
I've finished the decompression support, please see the conclusion here: https://github.com/mswjs/interceptors/pull/661#issuecomment-2429045746.

Will merge once the tests are passing.

tsup.config.ts
```diff
@@ -29,6 +35,12 @@ const browserConfig: Options = {
   format: ['cjs', 'esm'],
   sourcemap: true,
   dts: true,
+  esbuildOptions(options) {
+    options.alias = {
+      [`internal:brotli-decompress`]:
```
kettanaito, line 40: This is the environment-based decompression for the build.

test/vitest.config.js
```diff
@@ -10,4 +10,12 @@ export default defineConfig({
       'vitest-environment-react-native-like': './envs/react-native-like',
     },
   },
+  resolve: {
+    alias: {
+      // Create a manual alias for Vitest so it could resolve this
+      // internal environment-dependent module in tests.
+      'internal:brotli-decompress':
```
kettanaito, line 17: This is the environment-based decompression for the tests.

kettanaito (approved):
Looks good to me.

@mikicho, please, would you have a moment to browser through the changes? Would love to hear your thoughts too.

mikicho:
@kettanaito LGTM

Just curious about `PipelineStream`. I'm not sure why we need this.

kettanaito:
@mikicho, I added `PipelineStream` as an alternative for `pipeline` in Node.js but for web streams. We use it to compose a dynamic list of `TransformStream` (decompressors) into a chain of streams to apply to the response body. 

Do you see a better approach here?

mikicho:
Ohh Ok! I see..
Not sure if I'd stick with Web API here (DecompressionStream) because AFAIU this code is used only by Node.js, but the current implementation is great too.

kettanaito:
It's good to be consistent across the browser and Node.js, and the `DecompressionStream` is available in both of those environments. We are also working with `ReadableStream`, which is the response body. It's easier to convert `zlib` to a web stream than convert everything else to a Node.js stream.

kettanaito:
Right now, we're applying decompression left-to-right based on the `content-encoding` header:

https://github.com/mswjs/interceptors/pull/604/files#diff-8828d872f58df5de9fe0ddeacea2d0e5f24dc843916c68a2643769555d838694R47

Maybe that's the problem? 

Basically:

```
content-encoding: gzip, br
```

The decompression chain would be: input -> DecompressionStream('gzip') -> BrotliDecompression -> output.

mikicho:
This behavior looks correct; I'm not sure the code does it.

kettanaito:
Your example is incorrect. `compressResponse` behaves similarly to how a server handles the `Content-Encoding` header. It lists all the encodings in the order they were applied, left-to-right. So, for your example, the order would be gzip, then brotli.

Decompression must be reversed, and there's a bug currently that I will fix soon. 

## `DecompressionStream` in Node.js 18

Here's a fun find:

### Node.js v18

```sh
> let x = new DecompressionStream('gzip')
undefined
> x
DecompressionStream {
  readable: ReadableStream { locked: false, state: 'readable', supportsBYOB: false },
  writable: WritableStream { locked: false, state: 'writable' }
}
```

### Node.js v20

```sh
> let x = new DecompressionStream('gzip')
undefined
> x
undefined
```

### Node.js v22

```sh
> let x = new DecompressionStream('gzip')
undefined
> x
DecompressionStream {
  readable: ReadableStream { locked: false, state: 'readable', supportsBYOB: false },
  writable: WritableStream { locked: false, state: 'writable' }
}
```

In other words, Node.js v18-20 _expose_ the `DecompressionStream` class, but it returns nothing upon creation. The implementation itself seems to have been added in version 22 only. 

This is extremely odd. I will raise this with the folks in Node.js.

mikicho:
> Your example is incorrect. 

Why not? If I compressed by `gzip, br`, I expect to decompress by be -> gzip. Isn't it?

kettanaito:
I think we mean different things by `->`. I mean order, you seem to mean relation. Composition is easier:

```js
Content-Encoding: gzip, br

br(gzip(data))
```

@mikicho, I've pushed the fix that applies the decompression right-to-left to respect the right order of multiple compressions. Mocked tests pass, but I have two bypass tests failing:

- gzip, deflate
- gzip, br

The error seems to be coming from Undici unable to handle that compression:

```
ypeError: terminated
 ❯ Fetch.emit ../node:events:518:28

Caused by: Error: incorrect header check
 ❯ Zlib.zlibOnError [as onerror] ../node:zlib:189:17

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: { errno: -3, code: 'Z_DATA_ERROR' }
```

Our interceptor is not involved here at all. Am I constructing an invalid compressed response with `compressResponse()`? 🤔 I've fixed that function, too, to handle correct order of compression.

mikicho:
> I think we mean different things by ->. I mean order, you seem to mean relation.

Oh lol :D 

----
> I've pushed the fix that applies the decompression

I think that the `compose` function reverses the compression order:

```js
function compose(...fns: Array<Function>) {
  return fns.reduce((f, g) => {
    return (...args: Array<unknown>) => f(g(...args))
  })
}

function toUpperCase(s: string) {
  return s.toUpperCase()
}
function toLowerCase(s: string) {
  return s.toLowerCase()
}

compose(toLowerCase, toUpperCase)('hello') // prints hello, instead oh HELLO
```
I know the `compose` usually does that, but in our case, I think we want to preserve the order: first compress by `gzip` and only then compress by `br`.
So I think we want to inverse the invocation to `g(f(...args))`

kettanaito:
I fixed the `compose` function by removing it. The compose order is correct. 

The issue is in Undici, I can reproduce it standalone as well. Will report it to their repo.

mikicho:
@kettanaito Maybe it has something to do with: https://github.com/nodejs/undici/blob/30305063937a87e70561d2aeb5795788ed628ca2/lib/web/fetch/index.js#L2138-L2143

kettanaito:
Since the behavior is reproducible in isolation, our side of the implementation seems to be done. The intention of bypass tests is to ensure the interceptor doesn't interfere with the regular behavior of the request client, but in this case the request client has a bug so we cannot allow that to influence our tests.

I will skip those two scenarios and link the issue above them. Will do that after we discuss the issue with the Undici folks.

kettanaito (approved):

kettanaito:
This looks good to me. @mikicho, should we merge this?

kettanaito:
🚀

kettanaito:
## Released: v0.36.6 🎉

This has been released in v0.36.6!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.36.6)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.36.6)

---

PR #608, opened 2024-07-31
feat(WebSocket): support passthrough, correctly emit the `error` event

kettanaito:
> [!WARNING]
> Since WebSocket connections will now be established as-is if you haven't provided any interceptors, this is technically a breaking change despite fixing a bug. 

Interceptors has a bug that it will always emit the `open` even on the WebSocket client, even when there are no interceptor set up for any connections (so the connection cannot physically be intercepted). This results in such connections pending forever. 

> This may be a big against previously established expectations of WebSocket interception always being mock-first. But this is still a bug for bypassed connections, and it has to be fixed. 

I'm fixing this here without breaking any previously established promises. 

- WebSocket connection will be mock-first if _at least one `connection` listener on the interceptor is present_;
- WebSocket connection will be established as-is if no `connection` listeners on the interceptors are present (previously, the connection would be OPEN and pend forever). 

## Discoveries

- The `error` event is NOT dispatched on non-configurable closure codes, even if those codes represent errors (e.g. data cannot be accepted). 
- The `error` event is dispatched in cases like connection errors or _abrupt_ connection closures. 
- Also, Undici v5 has a bug validating WebSocket protocols: https://github.com/nodejs/undici/issues/2844. Fixed in Node.js v20 but not in v18.

## Roadmap

- [x] Add a failing test.
- [x] Establish the connection as-is if there were no interceptors set (`this.emit()` returned `false`). 
- [x] Fix the issue where `.close()` would dispatch an `error` event on non-configurable closure codes (that's not how WebSocket behave).

### DIFF
--- a/src/interceptors/WebSocket/WebSocketOverride.ts
+++ b/src/interceptors/WebSocket/WebSocketOverride.ts
@@ -2,6 +2,7 @@ import { invariant } from 'outvariant'
 import type { WebSocketData } from './WebSocketTransport'
 import { bindEvent } from './utils/bindEvent'
 import { CloseEvent } from './utils/events'
+import { DeferredPromise } from '@open-draft/deferred-promise'
 
 export type WebSocketEventListener<
   EventType extends WebSocketEventMap[keyof WebSocketEventMap] = Event
@@ -10,6 +11,7 @@ export type WebSocketEventListener<
 const WEBSOCKET_CLOSE_CODE_RANGE_ERROR =
   'InvalidAccessError: close code out of user configurable range'
 
+export const kPassthroughPromise = Symbol('kPassthroughPromise')
 export const kOnSend = Symbol('kOnSend')
 export const kClose = Symbol('kClose')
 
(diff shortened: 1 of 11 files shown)


kettanaito:
## Alternative solution

Push the WebSocket interception a layer higher, letting the interceptor decide on the connection, whether to mock it, error it, or establish as-is.

```js
interceptor.on('connection', ({ handshake }) => {
  handshake.accept()
  // handleshake.errorWith()
  // handshake.passthrough()
})
```

This may also be an answer to handling the handshake (protocol upgrade) request. It's been reported that we don't do that at the moment, which may cause issues (mostly in the browser).

kettanaito:
## Released: v0.34.0 🎉

This has been released in v0.34.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.34.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.34.0)

---

PR #611, opened 2024-08-06
fix(WebSocket): set `wasClean` to `true` on normal connection closures

### LINKED ISSUE #610: WebSocket: `wasClean` is incorrectly set to `true` on clean closures
Per WHATWG WebSocket spec:

> [7.1.4](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.4).  The WebSocket Connection is Closed
>
> When the underlying TCP connection is closed, it is said that _The
> WebSocket Connection is Closed_ and that the WebSocket connection is
> in the CLOSED state.  If the TCP connection was closed after the
> WebSocket closing handshake was completed, the WebSocket connection
> is said to have been closed _cleanly_.
>
> If the WebSocket connection could not be established, it is also said
> that _The WebSocket Connection is Closed_, but not _cleanly_.
> [Source](https://datatracker.ietf.org/doc/html/rfc6455#page-41:~:text=closed-,_cleanly_.)

This means that regardless of the closure `code`, `wasClean` will be true as long as the closing handshake has been performed before that. 

I translate this as: only closure due to _errors_ are not considered clean. 

We should support a `wasClean?: boolean` parameter in the `[kClose]` method on `WebSocketOverride` so that we can set it to `false` only in the case when we close the connection with error (e.g. when an unhandled exception occurred in the interceptor).

kettanaito:
- Fixes #610 

This change makes all client and interceptor-caused WebSocket connection closures as _clean_ (`wasClean: true`). This is correct according to the spec. 

The only exception is when an unhandled exception occurs in the interceptor and the WebSocket connection hasn't been already in a closing/closed state, then the interceptor will (a) emit `error`; (b) emit `close`, where `wasClean` will be `false`. This emulates a situation when an exception occurred on the server, thus making it incapable of performing the closing handshake with the client (and resulting in an error per spec).


src/interceptors/WebSocket/index.ts
```diff
@@ -136,7 +140,7 @@ export class WebSocketInterceptor extends Interceptor<WebSocketEventMap> {
                 socket.readyState !== WebSocket.CLOSING &&
                 socket.readyState !== WebSocket.CLOSED
               ) {
-                socket.close(1000, error.message)
+                socket[kClose](1000, error.message, false)
```
kettanaito, line 143: This is the only use case when the interceptor will produce a non-clean connection closure.

kettanaito:
## Released: v0.34.1 🎉

This has been released in v0.34.1!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.34.1)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.34.1)

---

PR #613, opened 2024-08-09
fix(XMLHttpRequest): support request `upload` events

kettanaito:
- Fixes #573 

> [!WARNING]
> Upload events are not implemented in JSDOM (`total` is always `0`). This, as all the other XHR functionality, must be tested in the actual browser. 

## Roadmap 

- [x] Add browser test for `upload` callbacks (`onprogress`, etc).
- [x] Add browser tests for different request bodies (plain string, `URLSearchParams`, etc). Ideally, cover every possible request body type (mostly to assert the right body length in `total`).

PR #615, opened 2024-08-09
fix(XMLHttpRequest): infer response body length if `Content-Length` header is missing

### LINKED ISSUE #614: XHR: Calculate `total` of the response body if no `Content-Length` header was provided
Currently, the `total` of response events will be `0` if no explicit `Content-Length` response header was provided. We can improve that by inferring the total response body length from the actual response body. 

https://github.com/mswjs/interceptors/blob/e937d4fe1b8af54a8c87ffa508136cc126a0d832/src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts#L306-L311

Something like this should suffice:

```js
const buffer = await response.clone().arrayBuffer()
const totalLength = buffer.byteLength
```

> Careful on cloning the response at the right time! 

This also needs in-browser tests.

abhishekpaturkar:
- Fixes #614 

This PR enhances the handling of response body length by inferring it when the Content-Length header is missing. If the header is present, its value is used; otherwise, the response body is cloned and its byte length is calculated.

```ts
const totalResponseBodyLength = response.headers.has('Content-Length')
      ? Number(response.headers.get('Content-Length'))
      : (async () => {
          const buffer = await response.clone().arrayBuffer();
          return buffer.byteLength;
        })()
```


src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts
```diff
@@ -305,10 +305,11 @@ export class XMLHttpRequestController {
 
     const totalResponseBodyLength = response.headers.has('Content-Length')
       ? Number(response.headers.get('Content-Length'))
-      : /**
-         * @todo Infer the response body length from the response body.
-         */
-        undefined
+      : (async () => {
```
kettanaito, line 308: We should move this to a standalone function and await it here. I've just merged #613 that makes `respondWith` asynchronous, which will allow us this. 

The reasoning is that reading the response body is async. So this method will continue dispatching `loadstart` and other events while the body is being read, having `total` as `0` still.

kettanaito: This is also why we need tests for this.

abhishekpaturkar: Can you help me with tests

abhishekpaturkar:
I just moved it into a standalone function `inferResponseBodyLength`

src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts
kettanaito: Sure thing. I will add some next week.

src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts
```diff
@@ -769,3 +767,8 @@ async function getXMLHttpRequestBodyInitLength(
 
   return body.toString().length
 }
+
+async function inferResponseBodyLength(response: Response): Promise<number> {
+  const buffer = await response.clone().arrayBuffer();
```
kettanaito, line 772: This is so nice I'm considering using something like this for calculating request body length as well. As long as the environment will construct the correct `Request`, it will take account things like FormData boundaries, which we are calculating manually right now.

abhishekpaturkar: Thanks! I'm really glad you liked the approach. I'd be happy to help if you want to apply this method elsewhere. Let me know!

kettanaito:
Closed in favor of #616. @abhishekpaturkar, please don't open pull requests from `main`, always create a feature branch. PRs from `main` make it impossible for maintainers to push changes to your upstream. Thanks.

PR #616, opened 2024-08-13
fix(XMLHttpRequest): calculate `total` length of response body

### LINKED ISSUE #614: XHR: Calculate `total` of the response body if no `Content-Length` header was provided

kettanaito:
- Fixes #614 
- Marks an XHR as handled earlier since calculating request/response total body length is asychronous, and the `onceRequestSettled` promise can resolve before the calculation is complete.
  - I think this is still a race condition issue but we lose nothing by being explicit and marking the request as handled early during the `.respondWith()` and `.errorWith()` calls.
- Fixes an issue where XHR wasn't marked as handled in the `.errorWith()` scenarios. 
- Fixes an issue where using `Document` as XHR body would result in incorrect Fetch API `Request` body being set. 

## Todos

- [x] Add in-browser test for the `total` in XHR events when dealing with a response without an explicit `Content-Type` header (one scenario for a mocked response, one for bypassed).


src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts
```diff
@@ -722,50 +723,3 @@ function define(
     value,
   })
 }
-
-async function getXMLHttpRequestBodyInitLength(
```
kettanaito, line 726: Happy to get rid of this and rely on Undici for the rest 🎉

src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts
```diff
@@ -647,9 +641,16 @@ export class XMLHttpRequestController {
...
    */
-  private toFetchApiRequest(): Request {
+  private toFetchApiRequest(
+    body: XMLHttpRequestBodyInit | Document | null | undefined
+  ): Request {
     this.logger.info('converting request to a Fetch API Request...')
 
+    // If the `Document` is used as the body of this XMLHttpRequest,
+    // set its inner text as the Fetch API Request body.
+    const resolvedBody =
```
kettanaito, line 651: ## Bugfix

Sneaking a bugfix here: if the XHR body was a `Document`, its content will now correctly be set on the created request instance.

kettanaito:
Hi, @abhishekpaturkar 👋 Would you like to give these changes a review?

src/interceptors/XMLHttpRequest/utils/getBodyByteLength.test.ts
```diff
@@ -0,0 +1,164 @@
+// @vitest-environment node
+import { it, expect } from 'vitest'
+import { getBodyByteLength } from './getBodyByteLength'
+
+const url = 'http://localhost'
+
+it('returns explicit body length set in the "Content-Length" header', async () => {
```
kettanaito, line 7: Added unit tests for `getBodyByteLength`.

test/modules/XMLHttpRequest/compliance/xhr-response-progress.browser.test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+test.beforeAll(async () => {
+  await httpServer.listen()
+})
+
+test.afterAll(async () => {
+  await httpServer.close()
+})
+
+test('supports response progress for a mocked response', async ({
```
kettanaito, line 41: Added browser tests.

kettanaito:
## Released: v0.34.3 🎉

This has been released in v0.34.3!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.34.3)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.34.3)

---

PR #632, opened 2024-09-11
fix(xhr): clone the request before calculating its body size

### LINKED ISSUE #630: XHR is broken in 0.34.3
It seems that a regression was introduced into the XMLHttpRequestController in the 0.34.3 release.

I have not yet been able to identify the root cause, but here is a [codesandbox](https://codesandbox.io/p/devbox/5xl3y2?file=/src/index.spec.ts:48,1-56,1) showing a test passing with `window.fetch()` and failing with `axios.put()`. The XHR request simply never returns and results in a timeout.

robbtraister:
- Fixes #630

The root cause of the issue is that cloning the Request to calculate the `Content-Length` header occurs too late. At the point in which clone() is called, the body buffer has already been consumed, so cloning fails.

Since this `kFetchRequest` value is only used for this specific purpose of calculating content-length, there is no harm in creating and storing a clone immediately upon creation of the Request object.

### DIFF
--- a/src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts
+++ b/src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts
@@ -163,7 +163,7 @@ export class XMLHttpRequestController {
 
             // Delegate request handling to the consumer.
             const fetchRequest = this.toFetchApiRequest(requestBody)
-            this[kFetchRequest] = fetchRequest
+            this[kFetchRequest] = fetchRequest.clone()
 
             const onceRequestSettled =
               this.onRequest?.call(this, {
@@ -296,7 +296,7 @@ export class XMLHttpRequestController {
      */
     if (this[kFetchRequest]) {
       const totalRequestBodyLength = await getBodyByteLength(
-        this[kFetchRequest].clone()
+        this[kFetchRequest]
       )
 
       this.trigger('loadstart', this.request.upload, {
--- a/test/modules/XMLHttpRequest/regressions/xhr-request-body-length.test.ts
+++ b/test/modules/XMLHttpRequest/regressions/xhr-request-body-length.test.ts
@@ -1,0 +1,41 @@
+// @vitest-environment jsdom
+import { vi, beforeAll, afterEach, afterAll, it, expect } from 'vitest'
+import { XMLHttpRequestInterceptor } from '../../../../src/interceptors/XMLHttpRequest'
+import { createXMLHttpRequest } from '../../../helpers'
+
+const interceptor = new XMLHttpRequestInterceptor()
+
+beforeAll(() => {
+  interceptor.apply()
+})
+
+afterEach(() => {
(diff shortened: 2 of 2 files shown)


ddolcimascolo:
Hey all, can we get this merged and msw released?

Thx,
David

robbtraister:
@kettanaito I tried adding a test, but I can't seem to duplicate the behavior. In the test environment, reading the request buffer does not seem to lock it from future attempts to read/clone.

This is a simplification of what I tried:
```js
  interceptor.once('request', async ({ controller, request }) => {
    await request.arrayBuffer() // <-- this is the addition to the existing test, along with changing to POST
    controller.respondWith(
      new UndiciResponse('Hello world') as unknown as Response
    )
  })
```

ddolcimascolo:
@robbtraister I'm impacted by this one and I'm using Vite + Vitest. Please see https://github.com/mswjs/msw/issues/2273#issuecomment-2340935025

kettanaito:
> I was able to figure out a hacky workaround by importing Request from "undici" and setting it to global.Request in the beforeAll hook, then reverting in the afterAll hook. Does that seem reasonable?

Hmm, this seems a bit too much. We _are_ using the Undici's `Request`. Let me take a look at what we can do to improve these tests, and how to drive out this problematic scenario.

kettanaito:
I've added an isolated test and confirm it failing with the following error without the proposed fix:

```
TypeError: unusable
 ❯ Request.clone ../node:internal/deps/undici/undici:7289:17
 ❯ XMLHttpRequestController.respondWith ../src/interceptors/XMLHttpRequest/XMLHttpRequestController.ts:299:29
    297|     if (this[kFetchRequest]) {
    298|       const totalRequestBodyLength = await getBodyByteLength(
    299|         this[kFetchRequest].clone()
       |                             ^
    300|       )
    301| 
 ❯ Object.onResponse ../src/interceptors/XMLHttpRequest/XMLHttpRequestProxy.ts:70:24
 ❯ handleResponse ../src/utils/handleRequest.ts:55:21
 ❯ Module.handleRequest ../src/utils/handleRequest.ts:207:12
 ❯ processTicksAndRejections ../node:internal/process/task_queues:95:5
 ❯ XMLHttpRequestController.xhrRequestController.onRequest ../src/interceptors/XMLHttpRequest/XMLHttpRequestProxy.ts:64:34
```

With the fix, the test is passing.

kettanaito (approved):
This looks great. Waiting for the tests before merging.

robbtraister:
It looks like the key difference is using the `jsdom` environment in lieu of `happy-dom`. Either way, thanks for getting it working.

kettanaito:
## Released: v0.35.3 🎉

This has been released in v0.35.3!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.35.3)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.35.3)

---

PR #637, opened 2024-09-14
Fix Request Proxy

mikicho:
https://github.com/nock/nock/issues/2780

Added failing test.

@kettanaito the root cause is that we use the [`kRawHeaders`](https://github.com/mswjs/interceptors/blob/main/src/interceptors/ClientRequest/utils/recordRawHeaders.ts#L153) to populate the `headers`.

Do you have any thoughts on how to solve this conflict? I might be able to sort this out later this week, but right now I'm not sure about the best way to proceed.


kanadgupta:
This may have been solved by https://github.com/mswjs/interceptors/pull/639

kettanaito:
@mikicho, thanks for raising this. I believe this might be fixed by #639, just like @kanadgupta suggests. Could you please bump it in Nock and let me know if the issue still persists? Thanks.

kettanaito:
🎉

PR #643, opened 2024-09-27
fix(ClientRequest): support intercepting buffered requests

kettanaito:
- Related https://github.com/mswjs/msw/issues/2259

## Changes

- The `MockHttpSocket` class now has a special scenario for _buffered_ requests. A buffered request is one whose stream is pending while the connection is being established. Example:

```js
req.on('socket', (socket) => {
  socket.on('connect', () => req.end())
})
```

> Here, the request header won't be sent until the socket connection is established. Previously, the interceptor would hang on this forever since writing the request header was mandatory as it triggers the request parser). 

- Refactors `MockHttpSocket` class a bit so both conventional write+end requests as well as write + end (deferred) requests trigger the same `_onRequestStart` processing pipeline.

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -59,6 +59,7 @@ export class MockHttpSocket extends MockSocket {
   private requestParser: HTTPParser<0>
   private requestStream?: Readable
   private shouldKeepAlive?: boolean
+  private requestHeaderSent: boolean
 
   private responseType: 'mock' | 'bypassed' = 'bypassed'
   private responseParser: HTTPParser<1>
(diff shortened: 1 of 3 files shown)


kettanaito:
## Issue

`_read` is non-deterministic and is called multiple times during the socket life cycle. We cannot rely on it to check if someone has written to the socket yet. Also, writes can happen after `_read`, which defeats its entire purpose. 

I'm trying to see what we can do to make this process more streamlined... Ideally, we can trigger `onRequest` on `MockHttpSocket.connect`. If the request listener consumes the request body, we are flushing the request header buffer anyway right now to make that read possible.

kettanaito:
## Released: v0.42.0 🎉

```
```

---

PR #645, opened 2024-09-27
fix(WebSocket): make server "close" event cancelable

### LINKED ISSUE #644: WebSocket: Close event is not cancelable, so its default cannot be prevented
This may call back to the Node's issue where `MessageEvent` is not cancelable. I can see the same being true for the close event:

https://github.com/mswjs/interceptors/blob/2a06bce76dddaeeeb38fc37f2d3e78fbe458ce23/src/interceptors/WebSocket/WebSocketServerConnection.ts#L349

Checking this always yields `false` even if `event.preventDefault()` has been called on the `close` event on the server object. 

We currently do a custom [`CancelableMessageEvent`](https://github.com/mswjs/interceptors/blob/2a06bce76dddaeeeb38fc37f2d3e78fbe458ce23/src/interceptors/WebSocket/utils/events.ts#L11) for message events. We may want to do the same for the `close` event.

kettanaito:
- Fixes #644

### DIFF
--- a/src/interceptors/WebSocket/WebSocketServerConnection.ts
+++ b/src/interceptors/WebSocket/WebSocketServerConnection.ts
@@ -345,7 +345,9 @@ export class WebSocketServerConnection {
     const closeEvent = bindEvent(
       this.realWebSocket,
       new CancelableCloseEvent('close', {
-        ...event,
+        code: event.code,
+        reason: event.reason,
+        wasClean: event.wasClean,
         cancelable: true,
       })
     )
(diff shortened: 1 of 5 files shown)


kettanaito:
## Released: v0.35.9 🎉

This has been released in v0.35.9!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.35.9)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.35.9)

---

PR #648, opened 2024-09-30
feat: update the node.js version range

mikicho:
Following this: https://github.com/mswjs/interceptors/pull/598#discussion_r1672224165

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
   "author": "Artem Zakharchenko",
   "license": "MIT",
   "engines": {
-    "node": ">=18.20.0 || >=20.12.1 || >=22"
+    "node": ">=18.20.0 <20 || >=20.12.1"
   },
   "scripts": {
     "start": "tsc --build -w",
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
   "author": "Artem Zakharchenko",
   "license": "MIT",
   "engines": {
-    "node": ">=18"
+    "node": ">=18.20.0 || >=20.12.1 || >=22"
   },
   "scripts": {
     "start": "tsc --build -w",


package.json
```diff
@@ -93,7 +93,7 @@
   "author": "Artem Zakharchenko",
   "license": "MIT",
   "engines": {
-    "node": ">=18"
+    "node": ">=18.20.0 <20 || >=20.12.1"
```
kettanaito, line 96: Why skipping v21? 

Also, could you give more context why we need to update the version?

mikicho: Because of the `rawHeaders` symbol change in Node.js.

> Why skipping v21?

What do you mean?
This range translates to: "any version between 18.20.0 and 20, OR anything above 20.12.1"
You can play this it [here](https://jubianchi.github.io/semver-check/#/constraint/%3E%3D18.20.0%20%3C20%20||%20%3E%3D20.12.1)

kettanaito: Still a bit confused. We currently don't depend on Node.js for raw headers and use our own proxy to gather them. Does the version change still apply in that case?

mikicho: Interesting..
```js
const http = require('http')

  name: 'my-interceptor',
})
interceptor.apply();
interceptor.on('request', async ({ controller }) => {
  controller.respondWith(new Response(null, { headers: { 'set-cookie':  ['mocked-cookie-0;mocked-cookie-1', 'mocked-cookie-2'] }}))
});

(async function () {
   http.get('http://example.com', (res) => {
    console.log(res.headers);
   });
})()
```

With version 20.11.1 the output is:
> { 'set-cookie': [ 'mocked-cookie-0;mocked-cookie-1,mocked-cookie-2' ] }

With version 20.12.2 the output is:
> {  'set-cookie': [ 'mocked-cookie-0;mocked-cookie-1', 'mocked-cookie-2' ] }

mikicho:
@kettanaito I'm closing this. I updated the version in `nock` instead. It may be related to the rawHeaders feature, which we will probably remove soon anyway 🎉

PR #650, opened 2024-10-01
fix(WebSocket): do not call public "server.send()" in message forwarding

kettanaito:
MSW proxies the `server.send()` method to log a message whenever _the user_ is sending something from the mock to the server. But since Interceptors also call `server.send()` during the message forwarding, the logs display an entry as if the user has called `server.send()`.

Fix this by wrapping the send mechanism in a private method and calling _that_ private method during the forwarding.


kettanaito:
## Released: v0.36.1 🎉

This has been released in v0.36.1!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.36.1)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.36.1)

---

PR #653, opened 2024-10-03
fix(WebSocket): allow reusing the same listener

### LINKED ISSUE #652: TypeError: Cannot redefine property: Symbol(kBoundListener)
The following error is thrown when reusing the same function as an event listener for WebSocket:

```
TypeError: Cannot redefine property: Symbol(kBoundListener)
```

```js
client.addEventListener('message', clientMessageListener)
client.addEventListener('message', clientMessageListener)
```

It is thrown because the `kBoundListener` symbol is trying to be redefined for another listener being attached:

https://github.com/mswjs/interceptors/blob/4fab617266b94e1c84235e355bad1e86687c1850/src/interceptors/WebSocket/WebSocketClientConnection.ts#L91-L96

Instead, we should check if `kBoundListener` is defined, and if it is, ignore that definition. That means you are reusing the same listener.

kettanaito:
- Fixes #652

### DIFF
--- a/src/interceptors/WebSocket/WebSocketClientConnection.ts
+++ b/src/interceptors/WebSocket/WebSocketClientConnection.ts
@@ -86,18 +86,21 @@ export class WebSocketClientConnection
     listener: WebSocketEventListener<WebSocketClientEventMap[EventType]>,
     options?: AddEventListenerOptions | boolean
   ): void {
-    const boundListener = listener.bind(this.socket)
+    if (!Reflect.has(listener, kBoundListener)) {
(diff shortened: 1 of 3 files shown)


kettanaito:
## Released: v0.36.2 🎉

This has been released in v0.36.2!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.36.2)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.36.2)

---

PR #659, opened 2024-10-10
fix: MockHttpSocket: if passthrough: dont flush write buffer on stream read

thecopy:
resolves https://github.com/mswjs/msw/issues/2309

**Underlying issue**
As i understand the code: if consumer attempts to start reading stream before buffer is sent, the items will be removed from the writeBuffer. If `passthrough()` is called after, it will miss those chunks.

This change disables calling `flushWriteBuffer` on `stream.read` if current request is a passthrough request.

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -63,6 +63,7 @@ export class MockHttpSocket extends MockSocket {
   private responseType: 'mock' | 'bypassed' = 'bypassed'
   private responseParser: HTTPParser<1>
   private responseStream?: Readable
+  private isPassthrough: boolean
 
   constructor(options: MockHttpSocketOptions) {
     super({
@@ -83,6 +84,8 @@ export class MockHttpSocket extends MockSocket {
         }
       },
     })
+
+    this.isPassthrough = false
 
     this.connectionOptions = options.connectionOptions
     this.createConnection = options.createConnection
@@ -156,6 +159,7 @@ export class MockHttpSocket extends MockSocket {
     if (this.destroyed) {
       return
     }
+    this.isPassthrough = true
 
     const socket = this.createConnection()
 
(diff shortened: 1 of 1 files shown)


kettanaito:
Hi, @thecopy. Thanks for opening this. 

If we don't flush request chunks on `read()`, then if you try to read the request's body in your handler, it will pend forever. 

Note that it's the handler that decides whether the request is mocked or passthrough, not the other way around. In other words, you cannot know if the request is passthrough until you flushed the request buffer on read so the handler can read the request body and decide on how to handle the request. 

What you are describing sounds like a bug in how we are handling the write buffer. We should still flush, but if the passthrough is established, we should send those chunks via the _original_ socket instance as well to make a proper request. Does that sound right to you? Wdyt?

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -63,6 +63,7 @@ export class MockHttpSocket extends MockSocket {
   private responseType: 'mock' | 'bypassed' = 'bypassed'
   private responseParser: HTTPParser<1>
   private responseStream?: Readable
+  private isPassthrough: boolean
```
kettanaito, line 66: Cannot this be represented via `responseType === 'bypassed`? We already have a flag to track this.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -472,7 +477,9 @@ export class MockHttpSocket extends MockSocket {
           // flush the write buffer to trigger the callbacks.
           // This way, if the request stream ends in the write callback,
           // it will indeed end correctly.
-          this.flushWriteBuffer()
+          if (!this.isPassthrough) {
+            this.flushWriteBuffer()
```
kettanaito, line 481: As I've mentioned, this introduces an impossible state. You won't know if the request is passthrough until the handler is finished running. For the handler to finish running, it must be able to read the request's body, for which we must flush these first chunks.

kettanaito:
## Released: v0.36.10 🎉

This has been released in v0.36.10!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.36.10)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.36.10)

---

PR #660, opened 2024-10-15
fix: define property in original init `Headers` as `Headers` instead of `string[]`

### LINKED ISSUE #651: Intercepting POST-requests handled by ky http client throws an TypeError "this._options.headers.get is not a function"
When setting up intercepting POST-requests with MSW which are being handled by ky ( https://github.com/sindresorhus/ky  )  I ran into an TypeError "this._options.headers.get is not a function" being raised in this line: https://github.com/sindresorhus/ky/blob/3ba40cc6333cf1847c02c51744e22ab7c04407f5/source/core/Ky.ts#L181

Since this behaviour occurs in MSW versions later than 2.4.3  I guess the issue is being right filed here.

paoloricciuti:
Closes #651 

Basically when defining the header property on the original `init` request or the original init headers we were assigning the value of `kRawHeaders` but that's just a string array. So if that `Request` is then used again trying to read the `headers` with `req.headers.get` it will fail since there's no `get` method in the array of strings.

Also i've noticed 
```ts
if (inferRawHeaders.length > 0) {
    defineRawHeadersSymbol(request.headers, inferredRawHeaders)
}
```
which is a classic TS mystake: you were checking if that function had more than one argument but that's always true so i switched to what i think you actually wanted to check.

I've added a couple of test to check this that fails before this PR.

### DIFF
--- a/src/interceptors/ClientRequest/utils/recordRawHeaders.test.ts
+++ b/src/interceptors/ClientRequest/utils/recordRawHeaders.test.ts
@@ -126,6 +126,21 @@ it('records raw headers (Reqest / Request as init)', () => {
   expect(getRawFetchHeaders(request.headers)).toEqual([['X-My-Header', '1']])
 })
 
+it("doesn't change the init headers instanceof (Request / Request as init)", () => {
+  recordRawFetchHeaders()
+  const init = new Request(url, { headers: [['X-My-Header', '1']] })
+  new Request(init)
+  expect(init.headers).toBeInstanceOf(Headers)
+})
+
+it("doesn't change the init headers instanceof (Request / Request with Headers as init)", () => {
+  recordRawFetchHeaders()
+  const headers = new Headers([['X-My-Header', '1']])
+  const init = new Request(url, { headers })
+  new Request(init)
+  expect(init.headers).toBeInstanceOf(Headers)
+})
+
 it('records raw headers (Request / Request+Headers as init)', () => {
   recordRawFetchHeaders()
   const init = new Request(url, { headers: [['X-My-Header', '1']] })
@@ -141,6 +156,7 @@ it('records raw headers (Request / Request+Headers as init)', () => {
     ['X-My-Header', '1'],
     ['X-Another-Header', '2'],
   ])
+  expect(request.headers).toBeInstanceOf(Headers)
 })
 
 it('records raw headers (Response / object as init)', () => {
@@ -156,6 +172,7 @@ it('records raw headers (Response / object as init)', () => {
     ['Content-Type', 'application/json'],
     ['X-My-Header', '1'],
   ])
+  expect(response.headers).toBeInstanceOf(Headers)
 })
 
 it('records raw headers (Response / array as init)', () => {
@@ -165,6 +182,7 @@ it('records raw headers (Response / array as init)', () => {
   })
 
   expect(getRawFetchHeaders(response.headers)).toEqual([['X-My-Header', '1']])
+  expect(response.headers).toBeInstanceOf(Headers)
 })
 
 it('records raw headers (Response / Headers as init)', () => {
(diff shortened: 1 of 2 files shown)


carpie:
I think the `Response` proxy needs this treatment as well.  Creating `HttpResponse` objects after these proxies are installed will throw `TypeError: init.headers.get is not a function`.  The following small script can demonstrate:

```
const { setupServer } = require('msw/node');
let s = setupServer();
s.listen();

const { HttpResponse } = require('msw');
HttpResponse.json('{}', { status: 200, statusText: 'OK' });
```

I was able to "fix" it locally by wrapping the right hand side of the assignment in https://github.com/mswjs/interceptors/blob/8fad9cfd3e51d4123afc1a43017687ccea63b70f/src/interceptors/ClientRequest/utils/recordRawHeaders.ts#L248 with a `new Headers()` after finding this PR.

paoloricciuti:
> I think the `Response` proxy needs this treatment as well. Creating `HttpResponse` objects after these proxies are installed will throw `TypeError: init.headers.get is not a function`. The following small script can demonstrate:
> 
> ```
> const { setupServer } = require('msw/node');
> let s = setupServer();
> s.listen();
> 
> const { HttpResponse } = require('msw');
> HttpResponse.json('{}', { status: 200, statusText: 'OK' });
> ```
> 
> I was able to "fix" it locally by wrapping the right hand side of the assignment in
> 
> https://github.com/mswjs/interceptors/blob/8fad9cfd3e51d4123afc1a43017687ccea63b70f/src/interceptors/ClientRequest/utils/recordRawHeaders.ts#L248
> 
> with a `new Headers()` after finding this PR.

You are absolutely right...i missed this because the way to trigger it is a bit weird. Pushed the fix now.

carpie:
> You are absolutely right...i missed this because the way to trigger it is a bit weird. Pushed the fix now.

Awesome! Thank you!

src/interceptors/ClientRequest/utils/recordRawHeaders.test.ts
```diff
@@ -126,6 +126,28 @@ it('records raw headers (Reqest / Request as init)', () => {
   expect(getRawFetchHeaders(request.headers)).toEqual([['X-My-Header', '1']])
 })
 
+it("doesn't change the init headers instanceof (Request / Request as init)", () => {
```
kettanaito, line 129: Maybe let's rephrase it to "preserves the original headers instance"?

paoloricciuti: Well it's not technically the same header instance...however if you think it's important we could rework this whole thing by pushing looping over the raw headers and set the  on the original instance

src/interceptors/ClientRequest/utils/recordRawHeaders.ts
```diff
@@ -225,8 +188,8 @@ export function recordRawFetchHeaders() {
           inferredRawHeaders.push(...inferRawHeaders(args[1].headers))
         }
 
-        if (inferRawHeaders.length > 0) {
-          defineRawHeadersSymbol(request.headers, inferredRawHeaders)
+        if (inferredRawHeaders.length > 0) {
```
kettanaito, line 191: This was a typo ✅

src/interceptors/ClientRequest/utils/recordRawHeaders.ts
```diff
@@ -225,8 +188,8 @@ export function recordRawFetchHeaders() {
...
+          ensureRawHeadersSymbol(request.headers, inferredRawHeaders)
```
kettanaito, line 192: Here I'm using `ensure` to prevent the raw headers symbol from being re-defined.

kettanaito:
Both interceptors and MSW test suites have passed with these changes. Let's get this out.

kettanaito (approved):

kettanaito:
@paoloricciuti, welcome to contributors! 🎉

kettanaito:
## Released: v0.36.5 🎉

This has been released in v0.36.5!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.36.5)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.36.5)

---

Issue #668, opened 2024-10-31
`globalThis.WebSocket` is undefined and it can't be proxied.


Afsoon:
Hello

Our current dependanbot PR is failing with the latest MSW version, after debugging and tracking changes in the repository I have conclude the following:

- Prior to this PR, https://github.com/mswjs/interceptors/pull/665, the check of environment was based if `globalThis.WebSocket !== undefined`, but the new guard function doesn't check it anymore https://github.com/mswjs/interceptors/blob/7aff641f8eddbbe53d5d23ddc4b8295e708dcdc0/src/utils/hasConfigurableGlobal.ts#L5-19
- I have tested this changes on Node 20 LTS and Node 22.11, new LTS version, with the same error `TypeError: Cannot create proxy with a non-object as target or handler`

Our setup is Vitest + Happy-DOM + Remix, I have tested prior version of MSW, and it didn't have this problem.

I'm not 100% sure of the fix, but I think it's a missing a check that is:
```
  if (typeof descriptor.get() === "undefined") {
    return false;
  }
```

If it's the fix, I can create a PR with the test.

kettanaito:
Hi, @Afsoon. Thanks for reporting this. 

Let's discuss it first. Just checking the getter not being there isn't enough—the property descriptor can have a `value` and no `get`, and that's a valid property that must return `true`. Will look at your pull request in more detail once I have a moment (thanks for opening that btw!).

kettanaito:
## Released: v0.36.8 🎉

This has been released in v0.36.8!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.36.8)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.36.8)

Make sure to always update to the latest version (`npm i @mswjs/interceptors@latest`) to get the newest features and bug fixes.

---

_Predictable release automation by [@ossjs/release](https://github.com/ossjs/release)_.

PR #669, opened 2024-10-31
fix(hasConfigurableGlobal): handle existing globals with undefined value

Afsoon:
fix: https://github.com/mswjs/interceptors/issues/668

### DIFF
--- a/src/utils/hasConfigurableGlobal.test.ts
+++ b/src/utils/hasConfigurableGlobal.test.ts
@@ -26,13 +26,13 @@ it('returns false if the global property does not exist', () => {
   expect(hasConfigurableGlobal('_non-existing')).toBe(false)
 })
 
-it('returns false if the global property exists and it\'s value is undefined', () => {
+it('returns false for existing global with undefined as a value', () => {
   Object.defineProperty(global, '_existsAndUndefined', {
     value: undefined,
     configurable: true,
   })
 
-  expect(hasConfigurableGlobal('_existsAndUndefined')).toBe(true)
+  expect(hasConfigurableGlobal('_existsAndUndefined')).toBe(false)
 })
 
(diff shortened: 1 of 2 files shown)


src/utils/hasConfigurableGlobal.test.ts
```diff
@@ -26,6 +26,15 @@ it('returns false if the global property does not exist', () => {
   expect(hasConfigurableGlobal('_non-existing')).toBe(false)
 })
 
+it('returns false if the global property exists and it\'s value is undefined', () => {
+  Object.defineProperty(global, '_existsAndUndefined', {
+    value: undefined,
+    configurable: true,
+  })
+
+  expect(hasConfigurableGlobal('_existsAndUndefined')).toBe(true)
```
kettanaito, line 35: This has to be `false` though. Proxying `undefined` will throw an error as it's not an object. 

You are correctly returning `false` in the implementation, but the test still expects `true`. Can you adjust the assertion please?

kettanaito (approved):
I've fixed the tests, added the check for the root cause at hand (getter that returns undefined). Looks good now!

src/utils/hasConfigurableGlobal.test.ts
Afsoon: You are right!

kettanaito:
@Afsoon, thanks for addressing this! Welcome to the contributors 👋

kettanaito:
## Released: v0.36.8 🎉

This has been released in v0.36.8!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.36.8)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.36.8)

---

PR #671, opened 2024-11-03
fix: export `RequestController` type

### LINKED ISSUE #670: [Potential bug] RequestController type definition not exported?
Hi all

we're on latest (as of today) version, 0.36.7. We are porting a piece of test code (for browser code, running under vitest with happy-dom). That test code used an older version of interceptors (0.27.2) where `Request` object was augmented with `respondWith()` method. In that scenario we manually "hacked" a type alias that extended `globalThis.Request` with such method, so that tsc was happy. So we had some private method at some point like

```ts
private async onRequestMade(request: OurAugmentedRequest, _requestId: string) { }
```

which was invoked as a handler for `this.interceptor.on('request', ...)`

Now in current version we're upgrading to use the `RequestController` for that (which BTW seems better from design POV). The handler method signature has become:

```ts
private async onRequestMade(request: Request, _requestId: string, controller: RequestController) { }
```

but it just happens that we cannot find type definition for that controller class:

immagine

We've checked into root [index.ts](https://github.com/mswjs/interceptors/blob/7aff641f8eddbbe53d5d23ddc4b8295e708dcdc0/src/index.ts), which in turn imports [glossary.ts](https://github.com/mswjs/interceptors/blob/7aff641f8eddbbe53d5d23ddc4b8295e708dcdc0/src/glossary.ts), but then RequestController is only imported there, not re-exported.

Maybe that's the root cause? Or maybe we're only doing something wrong.

HTH, and thanks for this very smart library!

kettanaito:
- Fixes #670

### DIFF
--- a/src/glossary.ts
+++ b/src/glossary.ts
@@ -2,6 +2,12 @@ import type { RequestController } from './RequestController'
 
 export const IS_PATCHED_MODULE: unique symbol = Symbol('isPatchedModule')
 
+/**
+ * @note Export `RequestController` as a type only.
+ * It's never meant to be created in the userland.
+ */
+export type { RequestController }
+
 export type RequestCredentials = 'omit' | 'include' | 'same-origin'
 
 export type HttpRequestEventMap = {


kettanaito:
## Released: v0.36.9 🎉

This has been released in v0.36.9!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.36.9)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.36.9)

---

PR #674, opened 2024-11-06
fix(ClientRequest): do not unshift write buffer when flushing request headers

kettanaito:
- Closes #659
- Fixes https://github.com/mswjs/msw/issues/2309

## Changes

Removes `this.writeBuffer.unshift()` when flushing request write buffer on the first `read()` of the request body stream. This will preserve the first chunk to be sent to the original server in case of passthrough. 

The fix itself is to call all the write callbacks in order but preserve `this.writeBuffer` so it's forwarded in its entirety over the original socket.


kettanaito:
## Released: v0.36.10 🎉

This has been released in v0.36.10!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.36.10)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.36.10)

---

PR #677, opened 2024-11-15
feat: support non-configurable responses

kettanaito:
> [!WARNING]
> A breaking change since `isResponseWithoutBody` utility is no longer exported from the package's root. Use `FetchResponse.isResponseWithBody` (notice inversion) instead. 

- Fixes https://github.com/mswjs/msw/issues/2307

## Motivation

Mostly to support passthrough responses like `101 Switching Protocols` that happen in the apps that use WebSockets. Technically, also unblocks mocking non-configurable responses, but you have to use custom logic to make that happen (since `new Response()` will still throw per spec). 

- [x] ClientRequest
- [x] XMLHttpRequest
- [x] fetch
- [x] RemoteHttpInterceptor

> Note: In tests, I'm often constructing a 101 response without `upgrade`, which is technically invalid. The issue is, I don't want to bring the way request clients handle upgrades themselves to tests because that's not the point. Interceptors just has to ensure the non-configurable response status codes are supported from the original server and from the mocks.


kettanaito:
## Released: v0.37.0 🎉

This has been released in v0.37.0!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.37.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.37.0)

---

PR #680, opened 2024-11-29
fix: Fix issue with reused send buffer

freehuntx:
Sometimes games or websites reuse the buffer they send.  
Especially game engines use this for performance reasons. This causes our mock to get already overwritten
bytes which were not sent like this.  

To fix this we should make a copy of the buffer as soon as possible.


freehuntx:
I currently have no access to a pc. But later i can share an example.  
In general a buffer can have a underlaying ArrayBuffer with a specific offset inside it.
E.g. 5mb arraybuffer where each send reuses the same space in that array (and just overwrites).

Our interceptor gets handed the same reference. And if our send gets delayed (e.g. by async/await operations) the underlaying arraybuffer already changed because it got reused.

So either the user of the library has to know this fact and do the copy himself or we make sure it cant change by copying it.

freehuntx:
Im a fan of preventing confusion and saving other peoples time. I know how long i debugged this shenanigans.

But its not my library and i respect this decision.
Ill delete my PR and sorry for wasting your time.

kettanaito:
No time wasted! I appreciate your effort on this a ton. Just not sure the library is the correct place to handle this particular use case. Let me know if you have any other feedback on Interceptors.

PR #685, opened 2024-12-01
fix(ClientRequest): prevent `req.write()` callbacks from being called twice on passthrough

### LINKED ISSUE #684: ClientRequest: write callback get called twice (`FormData` are broken)
The regression is introduced in `0.36.10`. When the interceptor is applied, we call the write callback twice.
```js
const https = require('https');

const int = new ClientRequestInterceptor()
int.apply()

const req = https.request('https://webhook.site/ae950c30-b7bf-46e7-bb40-2c32ca132689', {
  method: 'POST',
},
  res => {
    console.log(res.statusCode);
    res.resume();
  })


req.write('mamma mia', null, () => {
  console.log(1); // prints 1 twice
  req.end()
})
```

----

From what I see, for some reason, Node calls to the callback as well in https://github.com/nodejs/node/blob/3fb2ea8689250a3d39d9c66afd7301c7caa644c2/lib/internal/streams/writable.js#L764

I'm not sure why tho.

kettanaito:
- Fixes #684 
- Related to #674 
- Related to https://github.com/mswjs/msw/issues/2326

## Changes

During the `flushWriteBuffer()`, set the invoked callback on the write args to `undefined` so it doesn't get called again (twice) during passthrough of requests that call `request.end()` within the `request.write()` callback. The issue is only reproducible in that scenario and doesn't happen on sequential write/end calls.

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -420,13 +420,17 @@ export class MockHttpSocket extends MockSocket {
   }
 
   private flushWriteBuffer(): void {
-    for (const [, , callback] of this.writeBuffer) {
-      /**
-       * @note If the write callbacks are ever called twice,
-       * we need to mark them with a symbol so they aren't called
-       * again in the `passthrough` method.
-       */
-      callback?.()
+    for (const writeCall of this.writeBuffer) {
+      if (typeof writeCall[2] === 'function') {
+        writeCall[2]()
+        /**
+         * @note Remove the callback from the write call
+         * so it doesn't get called twice on passthrough
+         * if `request.end()` was called within `request.write()`.
+         * @see https://github.com/mswjs/interceptors/issues/684
+         */
+        writeCall[2] = undefined
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -420,13 +420,17 @@ export class MockHttpSocket extends MockSocket {
...
+    for (const writeCall of this.writeBuffer) {
+      if (typeof writeCall[2] === 'function') {
+        writeCall[2]()
+        /**
+         * @note Remove the callback from the write call
+         * so it doesn't get called twice on passthrough
+         * if `request.end()` was called within `request.write()`.
+         * @see https://github.com/mswjs/interceptors/issues/684
+         */
+        writeCall[2] = undefined
```
kettanaito, line 432: I decided to go with the replacement approach because it's simpler than introducing an internal symbol.

mikicho:
Thanks! `nock` tests pass. 🎉 

> so it doesn't get called again (twice) during passthrough

I forgot that `interceptors` passthrough the request by default 🤦

src/interceptors/ClientRequest/MockHttpSocket.ts
mikicho: I agree.

mikicho (approved):

kettanaito:
Thanks for looking at this one! 🎉

kettanaito:
## Released: v0.37.2 🎉

This has been released in v0.37.2!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.37.2)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.37.2)

---

PR #692, opened 2024-12-14
fix(ClientRequest): support requests made via `ClientRequest`

mikicho:


test/modules/http/intercept/client-request.test.ts
```diff
@@ -0,0 +1,67 @@
...
+const httpServer = new HttpServer((app) => {
+  app.get('/user', (req, res) => {
+    res.status(200).send('user-body')
+  })
+})
+
+const resolver = vi.fn<HttpRequestEventMap['request']>()
+
+const interceptor = new ClientRequestInterceptor()
+interceptor.on('request', resolver)
```
kettanaito, line 18: I actually grew to dislike this pattern. Sorry it's still present in the test suite. It leads to a shared state in a shape of the `resolver` function, and that's really bad. 

May I please ask you to add the request listener within the test instead? It will yield a much simpler setup. Thanks!

kettanaito (changes requested):
This looks great. I cannot believe we ignored the constructor itself. Thanks for fixing this, @mikicho!

I left one comment about the test, otherwise good.

kettanaito:
I'm thinking, will patching `ClientRequest` lead to double patching when using functions like `http.request()`? We don't replace their implementation anymore, only inject the `MockHttpAgent`, which means those functions will construct `new http.ClientRequest()` when called. Now that we are adding a patch over that class, does that lead to both of them being patched? 

I may be wrong about this. Would love to hear your thoughts on this. Also, I think we need tests for this anyway to have clear expectations toward how this should behave.

mikicho:
> I'm thinking, will patching ClientRequest lead to double patching when using functions like http.request()?

It also came to mind, but when I removed the `get/request` proxies, Node kept using the original `ClientRequest`. I think it happens because they destruct the `ClientRequest` in the `http` module: https://github.com/nodejs/node/blob/main/lib/http.js#L30

test/modules/http/intercept/client-request.test.ts
mikicho: Sure! like this?
image

kettanaito: I think so! 

And please add `interceptor.removeAllListeners()` to the `afterEach()` hook.

kettanaito:
_That_ is a great find. I also thought of some cases when we should keep the `.get`/`.request` proxies to be safe. I like this approach. I would just

- Check if the mock agent is already set on the `ClientRequest` level and then ignore everything else.
- Add at least one test that ensures http.request + ClientRequest combo. We may have to assert on implementation details here because proxies are not publicly observable. Something like: if I make `http.request` request, grab the `ClientRequest` instance it constructs and make sure that it uses a _single_ and same `MockHttpAgent` instance.

mikicho:
> Check if the mock agent is already set on the ClientRequest level and then ignore everything else.

In runtime or in a test?

kettanaito:
I've pushed some changes that construct a protocol-dependent mock agent (prevents "expected http: but got https: protocol" errors) as well as some HTTP/HTTPS tests for different constructor options to `ClientRequest`. Looks good now. 

Still need to add some tests to make sure there isn't double patching going on.

kettanaito (approved):
Looks good now! Thank you for working on this, @mikicho. I think we are ready to ship.

PR #695, opened 2024-12-27
fix(FetchResponse): set urlList #690

### LINKED ISSUE #690: FetchResponse does not set the urlList
Found in https://github.com/nock/nock/issues/2821

```js
const { FetchInterceptor } = require('@mswjs/interceptors/fetch');

(async () => {
  interceptor.on('request', ({ controller}) => {
    controller.responseWith(new Response('Hello, world!'))
  })
  // interceptor.apply()

  const response = await fetch('http://example.com')

  // When interceptor disabled
  console.log(response.url) // http://example.com
  console.log(response.clone().url) // http://example.com
  
  // When interceptor applied
  console.log(response.url) // http://example.com
  console.log(response.clone().url) // print empty string
})()
```

I think it happens because [`setUrl`](https://github.com/mswjs/interceptors/blob/main/src/utils/fetchUtils.ts#L30) does not push the URL into the `urlList`, and therefore does not get cloned.

dianliyang:
Fixes #690


mikicho:
@dianliyang why did you close this?

dianliyang:
> @dianliyang why did you close this?

I'm sorry about that, it might have been a mistake on my part. I'm still trying to understand how open source works.

dianliyang:
> > @dianliyang why did you close this?
> 
> I'm sorry about that, it might have been a mistake on my part. I'm still trying to understand how open source works.

Right, I get it now. I deleted the branch by accident.

PR #701, opened 2025-01-27
fix(ClientRequest): pass custom agent options in passthrough

mikicho:
resolves: https://github.com/nock/nock/issues/2828

### DIFF
--- a/src/interceptors/ClientRequest/agents.ts
+++ b/src/interceptors/ClientRequest/agents.ts
@@ -65,20 +65,24 @@ export class MockHttpsAgent extends https.Agent {
   }
 
   public createConnection(options: any, callback: any): net.Socket {
-    let _options = options
-    let createConnection: typeof net.createConnection
-    if (this.customAgent instanceof https.Agent) {
-      createConnection = this.customAgent.createConnection
-      _options = { ...options, ...this.customAgent.options }
-    } else {
-      createConnection = super.createConnection
-    }
+    const createConnection =
+      this.customAgent instanceof https.Agent
+        ? this.customAgent.createConnection
+        : super.createConnection
+
+    const createConnectionOptions =
+      this.customAgent instanceof https.Agent
+        ? {
+            ...options,
+            ...this.customAgent.options,
+          }
+        : options
 
     const socket = new MockHttpSocket({
       connectionOptions: options,
       createConnection: createConnection.bind(
         this.customAgent || this,
-        _options,
+        createConnectionOptions,
         callback
       ),
       onRequest: this.onRequest.bind(this),
--- a/src/interceptors/ClientRequest/agents.ts
+++ b/src/interceptors/ClientRequest/agents.ts
@@ -9,6 +9,7 @@ import {
 
 declare module 'node:http' {
   interface Agent {
+    options?: http.AgentOptions
     createConnection(options: any, callback: any): net.Socket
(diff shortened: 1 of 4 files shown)


test/modules/http/compliance/http-custom-agent.test.ts
```diff
@@ -62,3 +62,12 @@ it('preserves the context of the "createConnection" function in a custom https a
...
   expect(context.constructor.name).toBe('CustomHttpsAgent')
 })
+
+it('forward the custom agent\'s options', async () => {
+  const request = https.request(httpServer.https.url('/resource'), {
+    agent: new https.Agent({ rejectUnauthorized: false }),
+  })
+  request.end()
+
+  await expect(waitForClientRequest(request)).resolves.not.toThrow()
```
kettanaito, line 73: My only concern is that this test doesn't make it clear what's being asserted. Is the expectation for the request to complete without errors? Did it not do that before? We need to define the expectations here clearer, I think.

kettanaito:
This looks good! I've pushed little changes and also added this forwarding for HTTP agent as well. I think tests could do a bit better, let me know what you think.

src/interceptors/ClientRequest/agents.ts
```diff
@@ -66,15 +75,23 @@ export class MockHttpsAgent extends https.Agent {
...
+            ...this.customAgent.options,
+          }
+        : options
 
     const socket = new MockHttpSocket({
       connectionOptions: options,
       createConnection: createConnection.bind(
         this.customAgent || this,
-        options,
+        createConnectionOptions,
```
kettanaito, line 94: We don't really have any direct tests for this. Is this a reason for us to add `agents.test.ts` and test these things on the unit level?

test/modules/http/compliance/http-custom-agent.test.ts
mikicho: I see your point. 
If the options are not propagated correctly, the request will fail due to its HTTPS nature, as we did not provide a certificate.

kettanaito: I think I will remove this test. The option forwarding is implies in the `normalize*` function tests, which you've added.

kettanaito (approved):
This looks great 👍 Thanks for fixing it.

src/interceptors/ClientRequest/agents.ts
mikicho: I saw your merged it, but I'm wondering if we need here a unit test to make sure we merge the options correctly. 
I tested the normalize function, but I'm not sure if it's enough

kettanaito: Ideally, we can have a follow up PR that unit-tests the agents. I think that's the right surface to do it. Does this makes sense to you?

mikicho: Yes. I opened an issue for this, so we won't forget about this:
https://github.com/mswjs/interceptors/issues/703

kettanaito:
## Released: v0.37.6 🎉

This has been released in v0.37.6!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.37.6)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.37.6)

---

PR #704, opened 2025-02-04
fix(ClientRequest): add `rawHeaders` symbol to response listener instance

mikicho:
@kettanaito I hope I didn't create an abomination 😅 let me know if this is the way and if there is a more elegant way to achieve this.
The problem is that `FetchResponse` did not use our `Response` proxy, resulting in no `rawHeaders` symbol being created in the instance. By utilizing `Reflect.construct`, I ensured it went through the proxy as intended.

TODO:
1. There is a type error in [the test](https://github.com/mswjs/interceptors/blob/ac93a20e9d644c90cc1d9b3880d354aa1445dee4/test/features/events/response.test.ts#L148)
2. should/can we export the kRawHeaders symbol for the test or use `getOwnPropertySymbols`?
3. Add proper comments on the `extends` & `Reflect.construct` trick if it's a valid option.


kettanaito:
Hi, @mikicho. Thank you so much for opening this. 

> The problem is that FetchResponse did not use our Response proxy, resulting in no rawHeaders symbol being created in the instance. 

So, is it a timing issue? When `FetchResponse` is constructed, the proxy for recording headers isn't applied yet? Or something else?

src/utils/fetchUtils.ts
```diff
@@ -55,31 +55,32 @@ export class FetchResponse extends Response {
...
+    super()
     const status = init.status ?? 200
     const safeStatus = FetchResponse.isConfigurableStatusCode(status)
       ? status
       : 200
     const finalBody = FetchResponse.isResponseWithBody(status) ? body : null
-
-    super(finalBody, {
+    // construct the response manually so it will use the Response proxy for raw headers
+    const response = Reflect.construct(Response, [finalBody, {
```
kettanaito, line 66: My only concern with this is that it messes up the prototype chain. We can no longer do things like `x instanceof FetchResponse` because the class constructor returns an arbitrary object (a plain Fetch API Response instance).

src/utils/fetchUtils.ts
```diff
@@ -88,6 +89,7 @@ export class FetchResponse extends Response {
       }
     }
 
-    FetchResponse.setUrl(init.url, this)
+    FetchResponse.setUrl(init.url, response)
+    return response
```
kettanaito, line 93: I wonder if it wouldn't be better for us to flip the solution around:

- Export the header recording function to be able to apply it to any instance of `Request`/`Response`. 

So all we have to do in this class is 

```ts
return recordRawHeaders(this)
```

@mikicho, what do you think about this approach?

kettanaito (changes requested):

src/utils/fetchUtils.ts
mikicho: I like this approach more.. it's straightforward (kiss <3).

src/utils/fetchUtils.ts
mikicho: agree

mikicho:
@kettanaito WDYT?

test/features/events/response.test.ts
```diff
@@ -148,6 +149,7 @@ it('ClientRequest: emits the "response" event upon the original response', async
   expect(response.status).toBe(200)
   expect(response.statusText).toBe('OK')
   expect(response.headers.get('x-response-type')).toBe('original')
+  expect(Reflect.get(response.headers, kRawHeaders)).toContainEqual(['x-response-type', 'original'])
```
kettanaito, line 152: The only thing I don't quite like about this change is this assertion. We shouldn't tap into internal symbols anywhere. 

Is there any other way to check that the fix is working?

kettanaito:
@mikicho, this looks _good_! I have only one comment to share: https://github.com/mswjs/interceptors/pull/704#discussion_r2028491425. Would love for us to update the tests not to reference internal symbols.

test/features/events/response.test.ts
mikicho: I don't think so. This is what I'm planning to do in `nock` as well to extract the `rawHeaders` from the `response` instance in the `response` listener

mikicho:
@kettanaito, this is about the *response* raw headers, not the request's.

mikicho:
Closing this one favor of the `rawResponse` approach

PR #705, opened 2025-02-08
fix(fetch): clone passthrough request

mikicho:
Related to https://github.com/nock/nock/issues/2841

Clone the request so that it will be usable in the `response` listener


kettanaito:
Hi, @mikicho. Sorry, I didn't see this one. Already fixed in https://github.com/mswjs/interceptors/pull/710/commits/a2dff5277c8eff52edbf03a59566fa81eadb1850. I introduced a request clone for the `response` event instead of cloning the request provided to fetch. A minor difference, the behavior should be the same. 

That commit (and the PR) also improve on the fetch api tests, which now error in case you consume the request body in the request/response listeners and it hasn't been cloned correctly.

PR #706, opened 2025-02-08
fix(ClientRequest): inherit the connection `_handle` for passthrough requests

mikicho:
- Related to https://github.com/mswjs/msw/issues/2405

Since the `MockSocket` never actually connects to a server, it does not initialize its `_handle` property. As a result, [this push](https://github.com/mswjs/interceptors/blob/main/src/interceptors/Socket/MockSocket.ts#L56) creates a [connect listener](https://github.com/nodejs/node/blob/b18153598b25485ce4f54d0c5cb830a9457691ee/lib/net.js#L734) with every push.

@kettanaito Please guide me in the leftovers:
1. Fix the test name.
2. Should the test be in this file?
3. `this._handle` is private; there is an undocumented `handle` option in the Socket constructor. How do you think we should handle this?

Fixes: https://github.com/nock/nock/issues/2830


rossipedia:
I wonder if this could be the underlying cause of https://github.com/mswjs/msw/issues/2405?

mikicho:
Yes.. it might fix this as well.

test/modules/http/regressions/http-MaxListenersExceededWarning.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+afterAll(async () => {
+  interceptor.dispose()
+  await httpServer.close()
+})
+
+it('does ', async () => {
+  const url = httpServer.http.url('/')
+  const warningListener = vi.fn()
+  process.on('warning', warningListener)
```
kettanaito, line 40: This is an indirect assertion. I think you are onto something with that `connect` event. Can we somehow add a listener to that instead and assert that it's being called once? That would make a nice compliance test.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -178,6 +178,7 @@ export class MockHttpSocket extends MockSocket {
 
     const socket = this.createConnection()
     this.originalSocket = socket
+    this._handle = { reading: true, close: () => {} }
```
kettanaito, line 181: Why not infer the `_handle` of the original `socket` since we are in a passthrough context anyway? What do we gain by keeping a dummy socket, save that it passes the `this._handle` internal Node.js checks?

I would image we should do 

```ts
this._handle = socket._handle
```

One caution here is taking a look at _when_ the handle is assigned. Afaik, that happens in the Socket constructor, but in case it happens after, we should listen to that after and assign the `_handle` value then. 

What do you think?

kettanaito (changes requested):
This is a great discovery! It needs more discussion before we proceed with the fix. Would be grateful for your opinion on the comments I left.

src/interceptors/ClientRequest/MockHttpSocket.ts
mikicho: It's working, but I can't remember why I didn't take the `socket._handle`. I think I was afraid that the mocked socket and the real socket would somehow interfere with each other.
Anyway, it's working, and all the tests pass, so I'm okay to do this and fix any errors that occur.

test/modules/http/regressions/http-MaxListenersExceededWarning.test.ts
mikicho: Delightful comment! Fixed.

mikicho:
@kettanaito WDYT?

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: In this context, we are only handling passthrough requests so mocked functionality mustn't interfere with them. It should be safe to get the `_handle` create by Node.js.

mikicho:
Thank you for reporting this! Could you please open an issue with a minimal reproduction?

PR #714, opened 2025-04-04
fix(fetch): update `response.url` correctly

### LINKED ISSUE #690: FetchResponse does not set the urlList

kettanaito:
- Follows #695 
- Fixes #690
- Related to https://github.com/nock/nock/issues/2821

## Changes

Our Fetch API representation of the response now sets `response.url` value correctly. 

If working with Undici, instead of redefining the `url` property, it grabs the response internal state and pushes a new entry to its `urlList`. This way, Interceptors doesn't interfere with Undici's `url` getter on the response, which is implemented per spec.

> Also per spec, `response.url` will be `''` if the `Response` instance was constructed manually, which is the case for Interceptors. During regular requests, the response `urlList` becomes a clone of the request's `urlList` (see [this](https://github.com/nodejs/undici/blob/981636f5fb882ff684c11177a3bfc8b1cf061f3a/lib/web/fetch/index.js#L1598)). We cannot do that, so we push a new entry to the list instead.

If working with other Fetch implementations, Interceptors will keep redefining the `url` property because that is the lowest predictable surface for us (no guarantee those third-party implementations implement `urlList` at all).

This pull request also modifies the existing tests and adds new `response.url` compliance tests. 

> [!IMPORTANT]
> Interceptors will still check if the provided `response.url` is a valid `URL` instance. **Relative URLs will not work** as they never have**. Setting such invalid URLs will be skipped, and `response.url` will remain an empty string. This is done intentionally to provide a better, spec-aligned experience vs having the user deal with `Invalid URL` cryptic errors.

### DIFF
--- a/src/utils/fetchUtils.ts
+++ b/src/utils/fetchUtils.ts
@@ -1,3 +1,4 @@
+import { canParseUrl } from './canParseUrl'
 import { getValueBySymbol } from './getValueBySymbol'
 
 export interface FetchResponseInit extends ResponseInit {
@@ -49,7 +50,7 @@ export class FetchResponse extends Response {
   }
 
   static setUrl(url: string | undefined, response: Response): void {
-    if (url == null || url === 'about:') {
+    if (!url || url === 'about:' || !canParseUrl(url)) {
       return
     }
 
@@ -98,12 +99,10 @@ export class FetchResponse extends Response {
        * @note Undici keeps an internal "Symbol(state)" that holds
        * the actual value of response status. Update that in Node.js.
        */
-      const stateSymbol = Object.getOwnPropertySymbols(this).find(
-        (symbol) => symbol.description === 'state'
-      )
-      if (stateSymbol) {
-        const state = Reflect.get(this, stateSymbol) as object
-        Reflect.set(state, 'status', status)
+      const state = getValueBySymbol<UndiciFetchInternalState>('state', this)
+
+      if (state) {
+        state.status = status
       } else {
         Object.defineProperty(this, 'status', {
           value: status,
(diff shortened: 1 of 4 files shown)


src/utils/fetchUtils.ts
```diff
@@ -28,20 +49,25 @@ export class FetchResponse extends Response {
   }
 
   static setUrl(url: string | undefined, response: Response): void {
-    if (!url) {
+    if (url == null || url === 'about:') {
```
kettanaito, line 52: `about:` requests must keep `response.url` value equal to `''`:

Screenshot 2025-04-04 at 11 42 57

kettanaito:
Hi, @mikicho 👋 Solving the `response.url` issue in this one. Could you please take a look at this once you have a minute? Thank you.

kettanaito:
## Released: v0.38.1 🎉

This has been released in v0.38.1!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.38.1)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.38.1)

---

PR #715, opened 2025-04-04
fix(ClientRequest): allow GET requests with body

kettanaito:
- Fixes https://github.com/nock/nock/issues/2826

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -488,22 +488,20 @@ export class MockHttpSocket extends MockSocket {
     // If this Socket is reused for multiple requests,
     // this ensures that each request gets its own stream.
     // One Socket instance can only handle one request at a time.
-    if (canHaveBody) {
-      this.requestStream = new Readable({
-        /**
-         * @note Provide the `read()` method so a `Readable` could be
-         * used as the actual request body (the stream calls "read()").
-         * We control the queue in the onRequestBody/End functions.
-         */
-        read: () => {
-          // If the user attempts to read the request body,
-          // flush the write buffer to trigger the callbacks.
-          // This way, if the request stream ends in the write callback,
-          // it will indeed end correctly.
-          this.flushWriteBuffer()
-        },
-      })
-    }
+    this.requestStream = new Readable({
+      /**
+       * @note Provide the `read()` method so a `Readable` could be
+       * used as the actual request body (the stream calls "read()").
+       * We control the queue in the onRequestBody/End functions.
+       */
+      read: () => {
+        // If the user attempts to read the request body,
+        // flush the write buffer to trigger the callbacks.
+        // This way, if the request stream ends in the write callback,
+        // it will indeed end correctly.
+        this.flushWriteBuffer()
+      },
+    })
 
     const requestId = createRequestId()
     this.request = new Request(url, {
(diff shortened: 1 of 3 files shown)


mikicho:
@kettanaito I added `content-length` to the request, and now it throws the error.

Regarding the solution, we need the request body in the interceptor for interception and recording.
While I really admire your efforts to keep the API clear, there are edge cases (like this and/or the `rawHeaders`) that I'm not sure are worth going the extra mile for, as they're edge cases and, IMO, should be handled with a more straightforward solution.

kettanaito:
@mikicho, thanks for looking into this one.

There's a lot of difference between the raw request/response instances and the Fetch API representation that we are constructing. I understand how this can bite us back sometimes. But this alignment is still highly worth it! 

I am not a big fan of `rawHeaders` myself, but I don't see this stream change as a related issue. In fact, it's just a bug that we should fix (you can write to HTTP GET request streams, in the end). 

In the very early version of Interceptors, we also exposed the raw client request instance on the interceptor. I don't know if I like it, but we technically can do this:

```js
interceptor.on('request', ({ request, rawRequest }) => {
  request // Request
  rawRequest // ClientRequest OR XMLHttpRequest OR Request
})
```

This kind of API was removed because it introduces significant overhead at the usage layer. You become in charge to differentiate different request types, which kills the purpose of a single interception library. 

I also agree that all these use cases so far were edge cases. You normally don't need raw headers, for example, and not all request clients even support thats as a concept. I can also argue that HTTP headers are case-insensitive by design so the concept of recording and exposing raw headers is also flawed by design. I may be wrong on this so feel free to point that out.

mikicho:
@kettanaito We see eye to eye here. 
The thing I want to point out is that we may want to enable an "escape hatch" for legacy or non-standard behavior. New edge cases may arise on any platform, tool, library, etc. So, instead of trying to fit them into our main solution, we might develop a robust approach that allows us to expose these quirks in each client in a maintainable and convenient way. For example (not necessarily a good one that I recommend), instead of exposing the entire `rawRequest, ' which may be confusing and cumbersome, as you mentioned, we can expose the `extras` property, which includes specific properties like `rawHeaders` or `body`.
I totally agree that we need to be cautious with these escape hatches and introduce them only when there is no other option or when the cost of embedding a solution into the main flow is too high.

kettanaito:
> instead of exposing the entire rawRequest, ' which may be confusing and cumbersome, as you mentioned, we can expose the extras property, which includes specific properties like rawHeaders or body.

There's quite an important distinction between these two approaches, though. Exposing `kRawRequest` means "here's the keys, do your custom logic as you see fit". Exposing `kRawHeaders` or `kBody` means "here's an API we sort of maintain but it's not public.

Given how we are largely discussing edge cases, I think we should not build more logic around them than necessary (which is exactly the point you bring up and I agree). 

I wonder if we could remove the raw headers recording entirely and replace it with `kRawRequest` + custom logic on Nock's side (a matter of `request[kRawRequest].rawHeaders` or something)? From the back of your head, would that be possible? I don't think we rely on raw headers anywhere else but I may be wrong. 

> The symbol also isn't the only way to expose that API. In fact, perhaps we need something like:

```ts
import { getRawRequest } from '@mswjs/interceptors'

  const rawRequest = getRawRequest(request)
})
```

kettanaito:
## Released: v0.38.2 🎉

This has been released in v0.38.2!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.38.2)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.38.2)

---

mikicho:
> I wonder if we could remove the raw headers recording entirely and replace it with kRawRequest + custom logic on Nock's side 

Yes. Definitely into this kind of thing.

Issue #716, opened 2025-04-07
Fetch interceptor with query parameters as fetch input won't respect url


paulSerre:
Hi everyone,

```
      /**
       * @note Resolve potentially relative request URL
       * against the present `location`. This is mainly
       * for native `fetch` in JSDOM.
       * @see https://github.com/mswjs/msw/issues/1625
       */
      const resolvedInput =
        typeof input === 'string' &&
        typeof location !== 'undefined' &&
        !canParseUrl(input)
          ? new URL(input, location.origin)
          : input
```

This line made crash an application I have that use interceptors, because if someone run the following request:
`fetch('?cart=true')` being on `http://store-url.com/es-mx` the resolved url will be `http://store-url.com/?cart=true` instead of `http://store-url.com/es-mx?cart=true`

paulSerre:
> Hi, [@paulSerre](https://github.com/paulSerre). Thanks for reporting this.
> 
> It's a bit tricky. I agree that the URL resolution was incorrect in your case. But can also argue that a `?cart=true` request shouldn't happen in Node.js, to begin with. Our fetch interceptor is primarily designed for Node.js. You can, technically, use it in the browser but I highly discourage you from doing so (use MSW instead).
> 
> We do still have a bug though. Let's discuss the solution to it in the PR (thanks for opening it!).

Why did you write an XMLHttpInterceptor if it's only for Node.Js?

Unfortunately I'm forced to use this package instead of msw because I can"t rely on service worker.

kettanaito:
> Why did you write an XMLHttpInterceptor if it's only for Node.Js?

Because XMLHttpRequest has a significant usage in Node.js through JSDOM, which does run in a Node.js process. 

> Unfortunately I'm forced to use this package instead of msw because I can"t rely on service worker.

Got it! Then it's okay. 

Your fix is legitimate so we are getting it merged. Thank you for the help with it!

kettanaito:
Fixed by #717.

PR #717, opened 2025-04-07
fix(fetch): resolve relative urls against `location.href`, not `location.origin`

paulSerre:
See #716

### DIFF
--- a/src/interceptors/fetch/index.ts
+++ b/src/interceptors/fetch/index.ts
@@ -46,10 +46,8 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
         typeof location !== 'undefined' &&
         !canParseUrl(input)
           ? new URL(
-              input.startsWith('?') 
-                ? `${location.pathname}${input}`
-                : input,
-              location.origin
+              input,
+              location.toString()
             )
           : input
 
--- a/src/interceptors/fetch/index.ts
+++ b/src/interceptors/fetch/index.ts
@@ -45,7 +45,12 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
         typeof input === 'string' &&
         typeof location !== 'undefined' &&
         !canParseUrl(input)
-          ? new URL(input, location.origin)
+          ? new URL(
+              input.startsWith('?') 
+                ? `${location.pathname}${input}`
+                : input,
+              location.origin
+            )
           : input
 
       const request = new Request(resolvedInput, init)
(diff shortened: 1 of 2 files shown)


src/interceptors/fetch/index.ts
```diff
@@ -45,7 +45,12 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
         typeof input === 'string' &&
         typeof location !== 'undefined' &&
         !canParseUrl(input)
-          ? new URL(input, location.origin)
+          ? new URL(
+              input.startsWith('?')
```
kettanaito, line 49: Let's instead rely on the built-in resolution from the `URL` constructor.

```js
new URL(input, location)
```

The fix will be to swap `location.origin` with `location`. TIL that `URL` understands `location` as the base as well since the object is URL-like. So it will produce correct URLs, including your use case:

```js
new URL('/foo', 'https://a.b/c') // "https://a.b/foo"
new URL('?x=y', 'https://a.b/c') // "https://a.b/c?x=y"
```

It looks like we need a test for this, too!

paulSerre: Done

kettanaito (approved):
I pushed a few small adjustments. This looks great! Thank you so much for fixing this, @paulSerre 👏

test/modules/fetch/compliance/fetch-response-url.test.ts
```diff
@@ -87,3 +87,18 @@ it('returns the last response url in case of redirects', async () => {
   expect(response.url).toBe('http://localhost/destination')
   await expect(response.text()).resolves.toBe('hello world')
 })
+
+it('resolves relative URLs against location', async () => {
```
kettanaito, line 91: @paulSerre, looks like this test is missing the interceptor for the outgoing request. Could you please add one? See the examples in other tests in this file.

src/interceptors/fetch/index.ts
kettanaito: Thank you!

test/modules/fetch/compliance/fetch-response-url.test.ts
paulSerre: Yes sorry about this. I just add it

kettanaito (approved):

kettanaito:
Thank you so much for fixing this, @paulSerre. Welcome to contributors 🎉

kettanaito:
I am adding this fix to the base MSW as well in https://github.com/mswjs/msw/pull/2471.

PR #719, opened 2025-04-14
fix: add `getRawRequest` utility

kettanaito:
- Related to #704 

## Motivation

Some consumers need to access the raw request reference for their use cases (such as reading raw headers of `http.ClientRequest` in Nock). Instead of tailoring use case-driven APIs and escape hatches, like recording raw headers, I propose we (1) save the reference to the original request on the Fetch API `request` representation and (2) expose a new utility to get the value of that internal reference.

## Roadmap

- [x] Update the documentation.
- [x] Assess if this is enough to eventually drop raw headers recording.


src/interceptors/fetch/index.ts
```diff
@@ -45,13 +46,18 @@ export class FetchInterceptor extends Interceptor<HttpRequestEventMap> {
...
-            )
+          ? new URL(input, location.href)
           : input
 
       const request = new Request(resolvedInput, init)
+
+      /**
+       * @note Set the raw request only if a Request instance was provided to fetch.
+       */
+      if (input instanceof Request) {
```
kettanaito, line 57: There is no point of exposing references to strings/URLs as you can grab them at `request.url` already.

kettanaito:
Both the docs and raw headers replacement assessment will be done once this is released and tested. Nothing will be deprecated before that.

kettanaito:
## Released: v0.38.3 🎉

This has been released in v0.38.3!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.38.3)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.38.3)

---

PR #710, opened 2025-03-07
feat(RemoteHttpInterceptor): support fetch requests by default

kettanaito:
- Updates a bunch of dependencies.
- Removes a bunch of unused dependencies. 
- **Adds fetch interceptor to `RemoteHttpInterceptor`.**
- **Fetch: Clones `request` for the response event listener so its body can be read there.**
- Adjusts existing tests to the changes in dependencies. 
- Clones the fetch request to always have a clean instance to provide to `fetch()` on passthrough (closes https://github.com/mswjs/msw/issues/2408).

PR #720, opened 2025-04-17
fix: remove `jsdom` from `dependencies`

julienw:
See https://github.com/mswjs/interceptors/pull/710#issuecomment-2805495148 for context.

From my local testing, the JSDOM dependency wasn't used at all. I ran the tests with both node v18 and node v22.

Note that my `pnpm-lock.yaml` has some more changes, probably coming from my version of pnpm (I just installed it). Tell me if you'd like me to use a specific version of pnpm. I also added `pnpm-workspace.yaml` to the `.gitignore` file because it appeared after installing packages. Happy to remove this change if you think it shouldn't be there.

Tell me what you think!

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,6 +150,7 @@
     "follow-redirects": "^1.15.1",
     "got": "^14.4.6",
     "happy-dom": "^17.3.0",
+    "jsdom": "^26.1.0",
     "node-fetch": "3.3.2",
     "rimraf": "^6.0.1",
     "simple-git-hooks": "^2.7.0",
@@ -173,7 +174,6 @@
     "@open-draft/logger": "^0.3.0",
     "@open-draft/until": "^2.0.0",
     "is-node-process": "^1.2.0",
-    "jsdom": "^26.0.0",
     "outvariant": "^1.4.3",
     "strict-event-emitter": "^0.5.1"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -23,9 +23,6 @@ importers:
       is-node-process:
         specifier: ^1.2.0
         version: 1.2.0
-      jsdom:
-        specifier: ^26.0.0
-        version: 26.0.0
       outvariant:
         specifier: ^1.4.3
         version: 1.4.3
@@ -114,6 +111,9 @@ importers:
       happy-dom:
         specifier: ^17.3.0
         version: 17.3.0
+      jsdom:
+        specifier: ^26.1.0
+        version: 26.1.0
       node-fetch:
         specifier: 3.3.2
         version: 3.3.2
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -173,7 +173,6 @@
     "@open-draft/logger": "^0.3.0",
     "@open-draft/until": "^2.0.0",
     "is-node-process": "^1.2.0",
-    "jsdom": "^26.0.0",
```
kettanaito, line 176: Well, we still need `jsdom` for integration tests! Could you please move it to `dependencies `instead? Thanks.

.gitignore
```diff
@@ -5,4 +5,5 @@ node_modules
 lib
 package-lock.json
 .idea
-/test-results
\ No newline at end of file
+/test-results
+pnpm-workspace.yaml
```
kettanaito, line 9: I recommend you just checkout any unexpected files like this one. Don't add them to `.gitignore` (nothing wrong with committing a workspace config one day).

kettanaito (changes requested):
Thanks for following on this one, @julienw! I left a few comments. Please let me know your thoughts once you get to them.

package.json
julienw: are you sure? they seem to mostly work without jsdom. At least I get the same errors with and without jsdom... That's why I removed the dependency completely instead of moving to devDependency. But happy to add it back if you think about it strongly :-)

.gitignore
julienw: sounds good

package.json
kettanaito: We use jsdom for testing of the library. So yes, we depend on it while you don't.

julienw: Yes I understood this. But when after removing it, strangely I could still run all tests properly. I'm curious to know which tests need jsdom.

Anyway I pushed the commit adding jsdom to devDependencies. This also updates it to 26.1.0.

kettanaito: All XHR tests use JSDOM. If they ran, that must've been some issue or caching.

kettanaito (approved):

kettanaito:
Thanks for your work on this, @julienw! Welcome to contributors 🎉

kettanaito:
## Released: v0.38.6 🎉

This has been released in v0.38.6!

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.38.6)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/interceptors/v/0.38.6)

---

package.json
julienw: After removing it fully, when running `pnpm why jsdom` I still see it.

So I wondered if the peer dependencies were somehow installed automatically by pnpm.
And boom I see `autoInstallPeers: true ` is in `pnpm-lock.yaml`, so that's the reason: jsdom was present thanks to being a peer dependency of `vitest`.

However having it in `package.json` is useful because you're in control of the exact version, while `vitest`'s peer dependency is `*`.

I'm glad I got to the bottom of it :-)

Issue #729, opened 2025-06-10
If there are more than 30 HTTP headers, interceptor strips out the request headers and server does not receive them.


YangJonghun:
If there are more than 30 HTTP headers, interceptor strips out the request headers and server does not receive them

I make a PR #730, but it is temporarily bypass above request issue. and doesn't solve the same problem that occurs with `onResponseStart`.

`MockHttpSocket`'s `onResponseStart` and `onResponseStart` methods are doesn't provide correct headers(within rawHeaders) when it exceeds 30.

YangJonghun:
I'm attaching the response test code in case it helps someone who can fix it.

```typescript
const httpServer = new HttpServer((app) => {
  // ...
  app.get('/multiple-headers', (_req, res) => {
    res.setHeaders(new Headers(Object.fromEntries(Array.from({ length: 30 }, (_, i) => [`${i}`, `${i}`]))))
    res.status(200).send('/')
  })
})

// ...

it('HTTP headers are preserved even when their count exceeds 30', async () => {
  const responseListener =
    vi.fn<(...args: HttpRequestEventMap['response']) => void>()
  interceptor.on('response', responseListener)

  const req = http.get(httpServer.http.url('/multiple-headers'))

  const { res } = await waitForClientRequest(req)

  expect(responseListener).toHaveBeenCalledTimes(1)

  const { response } = responseListener.mock.calls[0][0]
  
  expect(res.headers).toMatchObject({
    "0": "0",
    "29": "29"
  })
  expect(Object.fromEntries(response.headers.entries())).toMatchObject({
    "0": "0",
    "29": "29"
  })
  expect(Object.fromEntries(response.headers.entries())).toMatchObject(res.headers)
})
```

YangJonghun:
@kettanaito 
In mswjs code, headers are lost within `this.requestParser.execute` inside `MockHttpSocket`.
In other words, headers are being dropped by the internal `http_parser` in Node.js.

Although I'm not very familiar with C++, it seems that [this code](https://github.com/nodejs/node/blob/main/src/node_http_parser.cc#L83-L84) inside Node.js is the cause.

(There are 32 in maximum, but since "connection" and "host" are assigned internally, custom headers are effectively limited to 30.)

YangJonghun:
@kettanaito 
I fixed response header issue also via PR #730

kettanaito:
Thanks for looking into this, @YangJonghun! Much appreciated.

[...]

PR #730, opened 2025-06-10
fix(http): preserve request http headers when their count exceeds 30

YangJonghun:
resolve issue #729
This PR is code to temporarily bypass above issue
(request is critical than response)

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -490,6 +490,11 @@ export class MockHttpSocket extends MockSocket {
     const headers = FetchResponse.parseRawHeaders(rawHeaders)
     const canHaveBody = method !== 'GET' && method !== 'HEAD'
 
+    const additionalHeaders: Record<string, string> | undefined = this.connectionOptions.headers
+    if (typeof additionalHeaders === 'object') { 
+      Object.entries(additionalHeaders).forEach(([k, v]) => void headers.set(k, v))
+    }
+
     // Translate the basic authorization in the URL to the request header.
     // Constructing a Request instance with a URL containing auth is no-op.
     if (url.username || url.password) {
--- a/test/modules/http/intercept/http-client-request.test.ts
+++ b/test/modules/http/intercept/http-client-request.test.ts
@@ -226,6 +226,32 @@ it('intercepts an HTTPS ClientRequest request with request options', async () =>
   expect(await text()).toBe('user-body')
 })
 
+it('HTTP headers are preserved even when their count exceeds 30', async () => {
+  const url = new URL(httpServer.https.url('/user?id=123'))
+
+  const requestListener =
+    vi.fn<(...args: HttpRequestEventMap['request']) => void>()
+  
+  interceptor.on("request", requestListener)
+
+  const req = new http.ClientRequest({
+    protocol: "https:",
+    hostname: url.hostname,
+    port: url.port,
+    path: url.pathname + url.search,
+    headers: Object.fromEntries(Array.from({ length: 30 }, (_, i) => [`${i}`, `${i}`]))
+  })
+  req.end()
+
+  await waitForClientRequest(req)
+  const { request } = requestListener.mock.calls[0][0]
+  expect(request).toBeInstanceOf(Request)
+  expect(Object.fromEntries(request.headers.entries())).toMatchObject({
+    "0": "0",
+    "29": "29"
+  })
+})
+
 it('restores the original ClientRequest class after disposal', async () => {
   interceptor.dispose()
 
--- a/test/modules/http/intercept/http.get.test.ts
+++ b/test/modules/http/intercept/http.get.test.ts
@@ -87,3 +87,24 @@ it('intercepts an http.get request given RequestOptions without a protocol', asy
   // Must receive the original response.
   expect(await text()).toBe('user-body')
 })
+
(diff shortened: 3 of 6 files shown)


test/features/events/response.test.ts
```diff
@@ -330,3 +335,27 @@ it('supports reading the request and response bodies in the "response" listener'
   expect(requestCallback).toHaveBeenCalledWith('request-body')
   expect(responseCallback).toHaveBeenCalledWith('mocked-response-text')
 })
+
+it('HTTP headers are preserved even when their count exceeds 30', async () => {
```
mikicho, line 339: I'm not sure if we need all of these tests.
Maybe what I wrote in the other PR (:face_palm:) is enough. https://github.com/mswjs/interceptors/pull/732/files#diff-02eaad8ff8105b4a1d9822c3e91031fc55fa93f6c29833634cda5602278c3538

YangJonghun: Yes, you're right — since request and response share most of the logic, having only response tests might be sufficient.
That said, I thought having a few more tests for request handling wouldn’t hurt.
There are already several request tests, but the existing ones are often written multiple times depending on whether it's `request` or `get`, or whether it's the `http` or `https` module — so I followed that pattern.

mikicho: We need to test both the response and the request (in my PR, I only implemented the response). Regarding our testing approach, I believe one test for each should suffice, and there's no need to test both HTTP and HTTPS or different methods.

YangJonghun: I agree! I'll simplify it to just one test for each :)

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -112,7 +117,9 @@ export class MockHttpSocket extends MockSocket {
 
     // Request parser.
     this.requestParser = new HTTPParser()
-    this.requestParser.initialize(HTTPParser.REQUEST, {})
+    this.requestParser.initialize(HTTPParser.REQUEST, {}, 0)
```
mikicho, line 120: @YangJonghun  What does the `0` mean?

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -485,6 +502,16 @@ export class MockHttpSocket extends MockSocket {
...
 
+    if (rawHeaders === undefined) { 
+      rawHeaders = this.headers
+      this.headers = []
+    }
+
+    if (path === undefined) {
+      path = this.url
+      this.url = ''
+    }
```
mikicho, line 513: Do we need this?

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -485,6 +502,16 @@ export class MockHttpSocket extends MockSocket {
   ) => {
     this.shouldKeepAlive = shouldKeepAlive
 
+    if (rawHeaders === undefined) { 
+      rawHeaders = this.headers
+      this.headers = []
```
mikicho, line 507: `this` here is the socket, not the `parser`

src/interceptors/ClientRequest/MockHttpSocket.ts
YangJonghun: Yes, I believe it's necessary.
I haven't identified the exact root cause yet, but from what I observed, when there were too many headers and they weren't properly delivered, `path` value was also missing.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -576,6 +603,15 @@ export class MockHttpSocket extends MockSocket {
     if (this.requestStream) {
       this.requestStream.push(null)
     }
+    this.headers = []
+    this.url = ''
+  }
+
+  private onResponseHeaders(headers: Array<string>, url: string): void {
+    // Once we exceeded headers limit - stop collecting them
+    if (this.responseParser.maxHeadersPair <= 0 || this.headers.length < this.responseParser.maxHeadersPair) {
```
kettanaito, line 612: I'm not adding this handling in the proposed changes because I think this isn't something we should handle. I'm happy to be proven wrong though.

_http_common.d.ts
```diff
@@ -10,13 +11,15 @@ declare var HTTPParser: {
...
+  [HTTPParser.kOnHeaders]: (headers: Array<string>, url: string) => void
   [HTTPParser.kOnHeadersComplete]: ParserType extends 0
     ? RequestHeadersCompleteCallback
     : ResponseHeadersCompleteCallback
   [HTTPParser.kOnBody]: (chunk: Buffer) => void
   [HTTPParser.kOnMessageComplete]: () => void
+  maxHeadersPair: number
 
-  initialize(type: ParserType, asyncResource: object): void
+  initialize(type: ParserType, asyncResource: object, maxHeaderSize?: number, lenient?: number, headersTimeout?: number): void
```
kettanaito, line 22: Could you please point me to Node.js source where you saw this additional arguments? I'm struggling to find them...

YangJonghun: It didn't have any particular meaning. I just aligned with Node.js types.

src/interceptors/ClientRequest/MockHttpSocket.ts
YangJonghun: I agree. It's not really necessary

PR #732, opened 2025-06-26
fix(ClientRequest): support more than 32 headers

mikicho:
TL;DR: Node call `kOnHeaders` if there are more than 32 headers.

https://github.com/nock/nock/discussions/2878

@kettanaito Need to tidy up this PR, but want to make sure you think this is the right direction first.


YangJonghun:
@mikicho
I think it's a duplicate of PR #730

mikicho:
@YangJonghun Epic!
Closing this one in favor of #730

Issue #675, opened 2024-11-07
this._getSession is not a function while use HttpsProxyAgent


WhoisUnknown:
Hi, I encounter a bug when I use proxy for requests

```
TypeError: this._getSession is not a function
    at HttpsProxyAgent.createConnection (node:https:140:26)
    at MockHttpSocket.passthrough (example/node_modules/@mswjs/interceptors/src/interceptors/ClientRequest/MockHttpSocket.ts:160:25)
    at _ClientRequestInterceptor.onRequest (example/node_modules/@mswjs/interceptors/src/interceptors/ClientRequest/index.ts:151:21)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Example repo https://github.com/WhoisUnknown/example-interseptor

I used version 0.36.10

fionawhim:
Seeing this as well, with 0.37.1 and Node v20.12.0. My use case is MSW running in a Vitest test, and trying to access (unmocked) HTTPS endpoints.

I think the code in the Node repo is here: https://github.com/nodejs/node/blob/v20.x/lib/https.js#L163

Possibly a situation in the way that the `createConnection` method is used [here](https://github.com/mswjs/interceptors/blob/main/src/interceptors/ClientRequest/agents.ts#L40)? It gets bound to something without a `_getSession` method, but an `_agentKey` is still being passed into options?

sebws:
https://github.com/mswjs/interceptors/blob/c338da8363660fbd08e794511e35678dae095670/src/interceptors/ClientRequest/agents.ts#L67-L82

Looks like what happens is that `this.customAgent` is there, but it isn't an instance of `https.Agent`, so `createConnection` is taken from `super.createConnection`. Then when we go to bind at L75-L79, it binds `super.createConnection` to `this.customAgent`. No good!

Given that `createConnection` does exist on the `HttpsProxyAgent` (this.customAgent), I think it shouldn't be that `super.createConnection` is now bound to `this` instead of `this.customAgent`, instead that the check isn't if `this.customAgent` is an instance of `https.Agent` but something else? 

I'm not sure of the intent here. Hopefully somebody who knows better what's intended to happen here, might just be @kettanaito

mikicho:
Maybe https://github.com/mswjs/interceptors/pull/701 fixed this.

mikicho:
@sebws do you mean something like:
```js
public createConnection(options: any, callback: any) { 
   const createConnection = 
[...]

Issue #570, opened 2024-04-29
Protocol "https:" not supported. Expected "http:" when using "proxy-agent"


indrek-sixfold:
When using ClientRequestInterceptor with ProxyAgent, it ends with error

	TypeError [ERR_INVALID_PROTOCOL]: Protocol "https:" not supported. Expected "http:"
		at new ClientRequest (node:_http_client:183:11)
		at new _NodeClientRequest (/home/indrek/service/node_modules/@mswjs/interceptors/lib/node/chunk-SXGRMPXP.js:226:5)
		at interceptorsHttpRequest (/home/indrek/service/node_modules/@mswjs/interceptors/lib/node/chunk-SXGRMPXP.js:850:12)
		at /home/indrek/service/node_modules/node-fetch/lib/index.js:1478:15
		at new Promise (<anonymous>)
		at fetch (/home/indrek/service/node_modules/node-fetch/lib/index.js:1447:9)
		at Generator.next (<anonymous>)

Code is here:

	import { ClientRequestInterceptor } from '@mswjs/interceptors/ClientRequest';
	import fetch from 'node-fetch';
	import { ProxyAgent } from 'proxy-agent';

	  async execute() {
		const interceptor = new ClientRequestInterceptor();
		interceptor.apply();
		await fetch('https://google.com', {
		  agent: new ProxyAgent(), // ALL_PROXY=http://127.0.0.1:3000
		});
		interceptor.dispose();
	  }

Using packages:

	node-fetch:2.7.0
	proxy-agent:6.4.0
	@mswjs/interceptors:0.29.1

Is there anything to configure to make it work with proxy?

kettanaito:
Hi, @indrek-sixfold! Thanks for reporting this. 

At the first glance, the issue is happening because the request-making module (`http` or `https`) is bound to the initial request when using Interceptors. I assume during proxying, since you are switching protocols, the interceptor attempts to perform the proxied request using the wrong request module for some reason. 

In theory, that shouldn't happen because the request originates from `proxy-agent`, we aren't constructing them internally (unless the request is bypassed, which I believe is precisely your use case). 

We need an integration test for this.

justemu:
This bug is still there.

PR #741, opened 2025-08-02
feat: intercept requests on the `net` socket level

### LINKED ISSUE #707: Support raw `net.connection`
- https://github.com/mswjs/msw/issues/2165
- https://github.com/mswjs/msw/issues/2409

Since https://github.com/mswjs/interceptors/releases/tag/v0.32.0, we now technically can implement a `net.Socket` interceptor that would allow us to intercept `ClientRequest` but also any agents using sockets directly, like Undici. 

This task would involve:

- Implementing the `net.Socket` interceptor interface.
- Refactoring ClientRequest (and potentially XHR) interceptors to rely on the Socket interceptor. These higher level interceptors would only concern themselves with parsing the socket packets and deciding whether a particular connection should be intercepted, then controlling it.
- Design some sort of manager that would "lock" socket instances behind certain interceptors. This is to avoid issues when a single request can be handled by multiple interceptors (see #378). Once we bring the interception to the socket level, _all_ requests will be covered by that interceptor and subsequent higher interceptors that extend it.

kettanaito:
- Fixes #707 
- Potentially related to #570, #675
- Enables #374 

## Motivation

This finally moves the request interception to the un-opinionated `net` module level, allowing each interceptor to route socket packets through their respective parsers (HTTP, SMTP, etc). 

## Changes

- `SocketInterceptor` patches `net.connect()` and `net.createConnection()` to introduce a proxy socket instance to listen to outgoing packets and mock incoming packets.
- Other interceptors, like the new take on `HttpRequestInterceptor`, can now become truly client-agnostic and use the underlying socket interceptor (should probably be a singleton) to provision the interception.
- The higher-level interceptors determine if the outgoing socket connection is relevant to them by _parsing the first sent packet_, which is usually the application protocol + metadata. 
- 6318aea correctly frees the request/response HTTP parsers, mimicking what Node.js does in its [`freeParser()`](https://github.com/nodejs/node/blob/f3adc11e37b8bfaaa026ea85c1cf22e3a0e29ae9/lib/_http_common.js#L180) internal utility.

## Roadmap

- [x] Try reusing `MockHttpSocket` somehow as it handles the suppression of connection errors and passthrough properly. 
- [x] Do not call `socket.connect()` in `net.connect()`. Instead, delegate that action to the user (i.e. the interceptor). Just in case they want to abort the connection before establishing it.
  - Do clients start writing _after_ `connect`? Double-check this. If so, we _must_ emit a preemptive `connect` but only in the `HttpRequestInterceptor`. 
 - [ ] Implement `HttpRequestInterceptor` 
 - [ ] Remove `ClientRequestInterceptor`
 - [x] ~~Update `XMLHttpRequestInterceptor`~~. Out of scope. New feature.
 - [x] ~~Update `FetchInterceptor`.~~
 - [ ] Check if `.prependOnce()` is implemented via `.prepend()` and if that's the case, ignore recording `.prependOnce()` like we do with `.once()`.
 - [ ] Expose the recorder from the `MockSocket` to stop the recording after initiating a mocked response. At that point, we know we will never need to replay those recordings (they are only for passthrough). Just call something like `socket[kRecorder].free()` in `HttpRequestInterceptor.respondWith()`.
 - [ ] Copy tests from https://github.com/mswjs/interceptors/pull/745#issuecomment-3419534078 to verify that these changes fix the referenced issue (`secureConnect` before `.write()`).
 - [ ] `SocketController.constructor`: finish the `onEntry` callback of the recorder for passthrough requests. **This fails the tests right now**. 
 - [ ] Design: decide how the `SocketInterceptor` will be reused across multiple higher-level interceptors. Should it be used as a singleton? Or just constructed everywhere and rely on the existing symbol-base interceptor deduping? Any pros/cons here?


kettanaito:
## Discovery: Skipped protocol check for HTTPS `ClientRequest`

Turns out, our previous implementation completely disregarded the unpatched behavior of `ClientRequest` when providing it an HTTPS URL. We've skipped the [protocol check](https://github.com/nodejs/node/blob/f3adc11e37b8bfaaa026ea85c1cf22e3a0e29ae9/lib/_http_client.js#L242) that has to be there!

If you want to perform an HTTPS request via `http.ClientRequest` you MUST use the HTTPS agent so Node.js knows what's the `expectedProtocol` (it takes that from the agent). If you don't, you will get this exception and that's a correct behavior we should keep:

```
Protocol "https:" not supported. Expected "http:"
```

> Because as far as Node.js is concerned, you are constructing an HTTP request. Nothing in your request says it has to be HTTPS. 

I've updated `test/modules/http/intercept/http-client-request.test.ts` tests for HTTPS to now reflect this behavior.

kettanaito:
## Note: Internal HTTP parser is not associated with the socket

The HTTP parsers we create for parsing out requests and responses are _not_ associated with the underlying socket, mocked or passthrough, in any way. Those sockets create their own parsers per Node.js and keep using them for their own purposes.

This is important to keep in mind when we free our own parsers. No need to nuke the `socket.parser` association: 

1. It's not set by us;
2. It will be cleared by `freeParser()` in Node.js for us.

kettanaito:
This is mostly prior art at this point. Closing in favor of functional https://github.com/mswjs/interceptors/pull/770.

Issue #738, opened 2025-07-30
Coercion of stream errors to 500 fails because headers have already been sent


paulbrimicombe:
Response bodies can be streams and we have tests that check the application behaviour if the response body stream errors for some reason (e.g. some network error while the body is streaming from the server). Currently this library catches any errors reading from the response stream and tries to coerce the response into a fulfilled 500 response ([in this `catch` block](https://github.com/mswjs/interceptors/blob/90c28f7593aa627228624c253365b431f7de1ae5/src/interceptors/ClientRequest/MockHttpSocket.ts#L396)). For a streaming response body, however, the status code and headers will already have been returned to the client ([see this code block](https://github.com/mswjs/interceptors/blob/90c28f7593aa627228624c253365b431f7de1ae5/src/interceptors/ClientRequest/MockHttpSocket.ts#L370-L374)).

In addition, the `unhandledException` event does not appear to be issued for this kind of error and so users can't opt-out of the response coercion (in our case even if it worked the response coercion is problematic — we want to simulate a network error rather than an incorrect server response).

Here is a minimal example:

```mjs
import { ClientRequestInterceptor } from './lib/node/interceptors/ClientRequest/index.mjs'
import https from 'https'
import webstream from 'stream/web'

const interceptor = new ClientRequestInterceptor()

interceptor.apply()

interceptor.on('request', ({ controller }) => {
  let readOperations = 0
  const readable = new webstream.ReadableStream({
    pull: (controller) => {
      if (readOperations === 3) {
        controller.error(new Error('Interrupted'))
        return
      }
      controller.enqueue(`CHUNK-${readOperations}/`)
      readOperations += 1
    },
  })

  const response = new Response(readable, {})
  controller.respondWith(response)
})

https
  .get('https://www.example.org')
  .on('response', (response) => {
    console.log('Event: response', `${response.statusCode}`)

    const chunks = []
    response
      .on('data', (chunk) => {
        chunks.push(chunk)
      })
      .on('error', (error) => {
        console.log('Event: error', error)
      })
      .on('end', () => {
        console.log('Event: end')
        console.log('BODY:', Buffer.concat(chunks).toString('utf-8'))
      })
  })
  .on('error', () => console.log('Event: error'))
```

The console output looks like this:

```text
Event: response 200
Event: end
BODY: CHUNK-0/CHUNK-1/CHUNK-2/HTTP/1.1 500 Unhandled Exception
Content-Type: application/json

[...]

PR #742, opened 2025-08-04
fix(MockHttpSocket): handle response body streaming errors correctly

paulbrimicombe:
# Problem

- Response body streaming errors are not handled correctly
- See #738 

# Solution

- Add a number of tests around stream destruction (socket / response body / request) including some baseline tests to verify the Node.js behaviour without any interception enabled
- Expand existing response readable stream tests to add more error scenarios
- Update `MockHttpSocket` so it:
  - Emits an `error` event if the response body stream errors after the headers have been returned (rather than attempting to return a 500 status error)
  - Only calls `error` event listeners once on error
  - Forwards response body streaming errors to the `IncomingRequest` that the client is interacting with

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -7,13 +7,14 @@ import {
 } from '_http_common'
 import { STATUS_CODES, IncomingMessage, ServerResponse } from 'node:http'
 import { Readable } from 'node:stream'
+import { pipeline } from 'node:stream/promises'
+import webstream from 'node:stream/web'
 import { invariant } from 'outvariant'
 import { INTERNAL_REQUEST_ID_HEADER_NAME } from '../../Interceptor'
 import { MockSocket } from '../Socket/MockSocket'
 import type { NormalizedSocketWriteArgs } from '../Socket/utils/normalizeSocketWriteArgs'
 import { isPropertyAccessible } from '../../utils/isPropertyAccessible'
 import { baseUrlFromConnectionOptions } from '../Socket/utils/baseUrlFromConnectionOptions'
-import { createServerErrorResponse } from '../../utils/responseUtils'
 import { createRequestId } from '../../createRequestId'
 import { getRawFetchHeaders } from './utils/recordRawHeaders'
 import { FetchResponse } from '../../utils/fetchUtils'
@@ -126,7 +127,8 @@ export class MockHttpSocket extends MockSocket {
     // Response parser.
     this.responseParser = new HTTPParser()
     this.responseParser.initialize(HTTPParser.RESPONSE, {})
-    this.responseParser[HTTPParser.kOnHeaders] = this.onResponseHeaders.bind(this)
+    this.responseParser[HTTPParser.kOnHeaders] =
+      this.onResponseHeaders.bind(this)
     this.responseParser[HTTPParser.kOnHeadersComplete] =
       this.onResponseStart.bind(this)
     this.responseParser[HTTPParser.kOnBody] = this.onResponseBody.bind(this)
@@ -159,17 +161,17 @@ export class MockHttpSocket extends MockSocket {
     return emitEvent()
   }
 
-  public destroy(error?: Error | undefined): this {
+  public _destroy(
+    error?: Error | null,
+    callback?: (error?: Error | null) => void
+  ): void {
     // Destroy the response parser when the socket gets destroyed.
     // Normally, we shoud listen to the "close" event but it
     // can be suppressed by using the "emitClose: false" option.
     this.responseParser.free()
 
-    if (error) {
-      this.emit('error', error)
-    }
-
-    return super.destroy(error)
+    callback?.(error)
+    process.nextTick(() => this.emit('close'))
   }
 
   /**
@@ -201,8 +203,8 @@ export class MockHttpSocket extends MockSocket {
     }
 
     // If the developer destroys the socket, destroy the original connection.
-    this.once('error', (error) => {
-      socket.destroy(error)
+    this.once('error', () => {
+      socket.destroy()
     })
 
     this.address = socket.address.bind(socket)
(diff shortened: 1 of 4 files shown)


src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -160,17 +161,17 @@ export class MockHttpSocket extends MockSocket {
...
     // can be suppressed by using the "emitClose: false" option.
     this.responseParser.free()
 
-    if (error) {
-      this.emit('error', error)
-    }
-
-    return super.destroy(error)
+    callback?.(error)
+    process.nextTick(() => this.emit('close'))
```
paulbrimicombe, line 174: This seems to be required to mimic the Node.js behaviour for the appropriate events to be emitted (see the `http-res-destroy` tests))

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -202,8 +203,8 @@ export class MockHttpSocket extends MockSocket {
     }
 
     // If the developer destroys the socket, destroy the original connection.
-    this.once('error', (error) => {
-      socket.destroy(error)
+    this.once('error', () => {
+      socket.destroy()
     })
```
paulbrimicombe, line 208: Forwarding the error to the `socket` leads to the `error` event handler being called twice.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -380,21 +381,24 @@ export class MockHttpSocket extends MockSocket {
...
-          const { done, value } = await reader.read()
-
-          if (done) {
-            serverResponse.end()
-            break
-          }
-
-          serverResponse.write(value)
-        }
+        serverResponse.socket?.on('error', () => {})
```
paulbrimicombe, line 384: There isn't an `error` handler on the mock socket we're using on the `serverResponse`. When we pipe the response body to the server response and there's an error, the error event is propagated to the underlying socket. If we don't handle that, we get an unhandled event error and the Node.js kills the process.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -380,21 +381,24 @@ export class MockHttpSocket extends MockSocket {
...
+          ),
+          serverResponse
+        )
       } catch (error) {
-        // Coerce response stream errors to 500 responses.
-        this.respondWith(createServerErrorResponse(error))
+        // We don't have direct access to the real IncomingMessage here to
+        // emit the 'error' event so we have to reach to the Socket internal state
+        // @ts-expect-error
+        this._httpMessage?.res?.emit('error', error)
```
paulbrimicombe, line 396: This is pretty annoying but we have to manually forward the error event to the actual `IncomingMessage` so that any `error` event handler the user has attached to it is called correctly.

test/modules/http/compliance/http-res-destroy.test.ts
```diff
@@ -12,57 +14,158 @@ const httpServer = new HttpServer((app) => {
...
   await httpServer.close()
 })
 
-it('emits the "error" event when a bypassed response is destroyed', async () => {
-  const socketErrorListener = vi.fn()
+const scenarios = [
+  {
+    condition: 'when the interceptor is not enabled [baseline test]',
+    setup: () => {},
+  },
```
paulbrimicombe, line 33: The first scenario is for plain Node.js without this library enabled at all. This means that we can check that the assertions we're using are correct based on the behaviour of Node.js.

test/modules/http/compliance/http-res-destroy.test.ts
```diff
@@ -12,57 +14,158 @@ const httpServer = new HttpServer((app) => {
...
+      })
+      .on('response', (response) => {
+        response.on('error', responseErrorListener)
+        response.destroy(responseError)
+      })
+      .on('error', requestErrorListener)
+
+    const { res } = await waitForClientRequest(request)
+
+    await setTimeout(0)
```
paulbrimicombe, line 76: Some of the event handling etc. is performed async so we have to yield the event loop.

test/modules/http/response/http-response-readable-stream.test.ts
```diff
@@ -29,13 +29,16 @@ afterAll(async () => {
...
-        controller.enqueue(encoder.encode(' '))
-        controller.enqueue(encoder.encode('world'))
-        controller.close()
+      pull(controller) {
+        const chunk = chunks.shift()
+        if (chunk) {
+          controller.enqueue(encoder.encode(chunk))
+        } else {
+          controller.close()
+        }
```
paulbrimicombe, line 41: Using `pull` is a more realistic way of creating a streaming response than using `start`

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: Let's store `IncomingMessage` in a variable and emit error on it directly maybe? So we don't have to claw our way to the same message reference here.

src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -160,17 +161,17 @@ export class MockHttpSocket extends MockSocket {
...
     this.responseParser.free()
 
-    if (error) {
-      this.emit('error', error)
-    }
-
-    return super.destroy(error)
+    callback?.(error)
+    process.nextTick(() => this.emit('close'))
   }
```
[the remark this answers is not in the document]
kettanaito: We can also just assign to `this._destroy` and keep the types provided by Node.js instead of re-implementing the `_destroy` method. I'd recommend we did that (similar to what we do with `this._final` right now).

src/interceptors/ClientRequest/MockHttpSocket.ts
kettanaito: Could you please link the Node.js source that we are mimicking here? That would be really helpful.

kettanaito:
Hi, @paulbrimicombe. I've accidentally got tangled in this issue on a different branch and would like to ask your opinion on the fix I'm proposing here: https://github.com/mswjs/interceptors/pull/749/files#diff-94fc27cca818a0da8d402e251763d5dbf450d62c3cde37ceef7ea8494a324dde

What do you think?

test/modules/http/response/http-response-readable-stream.test.ts
```diff
@@ -127,19 +132,69 @@ it('forwards ReadableStream errors to the request', async () => {
...
+  expect(responseErrorListener).toHaveBeenCalledWith(
+    expect.objectContaining({ code: 'ERR_INVALID_ARG_TYPE' })
+  )
 
-  // Response stream errors are not request errors.
-  expect(requestErrorListener).not.toHaveBeenCalled()
-  expect(request.destroyed).toBe(false)
+  expect(requestErrorListener).toHaveBeenCalledOnce()
+  expect(requestErrorListener).toHaveBeenCalledWith(
+    expect.objectContaining({ code: 'ERR_INVALID_ARG_TYPE' })
```
kettanaito: I don't believe this is correct. From what I've seen, once `response` is emitted, request can only be aborted if the response stream errors. This test implies that the request stream receives the same error as the response stream, but those are different streams and they do not share errors. Once the response headers are received, the request has done its work (well, technically, when the request stream is closed but you get the idea).

PR #754, opened 2025-11-29
chore: switch to trusted publishing

kettanaito:


kettanaito:
Weird CI failures due to FormData responses returning different byte length in Noide.js v24. We aren't even running against v24... Odd.

Reported in Node.js: https://github.com/nodejs/undici/issues/4680

PR #755, opened 2025-12-08
fix(ClientRequest): remove passthrough socket listeners on close

Stanzilla:
## Summary

This PR fixes a memory leak that occurs when using passthrough mode in Node.js 20+.

## The Problem

When `MockHttpSocket.passthrough()` is called, event listeners are attached to the original socket (lookup, connect, secureConnect, secure, session, ready, drain, data, error, resume, timeout, prefinish, finish, close, end) but they are never removed when the socket closes.

This causes memory leaks in long-running processes, as reported in mswjs/msw#2537.

## The Fix

Call `socket.removeAllListeners()` in the close event handler before emitting the close event to the MockHttpSocket. This ensures all listeners attached to the original socket are properly cleaned up.

```typescript
.on('close', (hadError) => {
  // Remove all listeners from the original socket to prevent memory leaks.
  // @see https://github.com/mswjs/msw/issues/2537
  socket.removeAllListeners()
  this.emit('close', hadError)
})
```

## Test

Added a regression test that makes many passthrough requests and verifies no `MaxListenersExceededWarning` is emitted.

Fixes mswjs/msw#2537

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -44,7 +44,6 @@ interface MockHttpSocketOptions {
 }
 
 export const kRequestId = Symbol('kRequestId')
-export const kPassthroughSocket = Symbol('kPassthroughSocket')
 
 export class MockHttpSocket extends MockSocket {
   private connectionOptions: HttpConnectionOptions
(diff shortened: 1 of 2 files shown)


kettanaito:
Thanks for working on this, @Stanzilla. This looks great. One issue here is that the test you've added passes even if your fix is reverted. I'm trying to add a direct test on the socket listeners instead, but that proves problematic since request's `socket` is _our_ `MockHttpSocket`. The underlying passthrough socket instance is never exposed.

kettanaito:
## Released: v0.41.0 🎉

This has been released in v0.41.0.

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.41.0)
- 📦 [View on npm](https://www.npmjs.com/package/@mswjs/interceptors/v/0.41.0)

```
```

---

PR #756, opened 2025-12-11
Fix: Correctly pass http version to the ServerResponse in node

RobinVdBroeck:
This PR updates the MockHttpSocket to correctly set the http version passed to ServerResponse. This fixes a bug where mocked responses with the Connection: keep-alive header never completed before. For that, I've added some regression tests.

I found it kinda hard to wrap my head around the architecture of mswjs/interceptors, so please let me know if I did something stupid.

This fixes a bug in nock: https://github.com/nock/nock/issues/2918

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -348,7 +348,7 @@ export class MockHttpSocket extends MockSocket {
     // This is critical for Node.js to correctly allow keep alive.
     // By default, nodejs sets keepalive to false if the HttpVersion is not at least
     // 1.1: https://github.com/nodejs/node/blob/70f6b58ac655234435a99d72b857dd7b316d34bf/lib/_http_server.js#L211C1-L211C34
-    // The version should normally already have been set by the onResponseStart method
+    // The version should normally already have been set by the onRequestStart method
     if(this.httpVersionMajor != null && this.httpVersionMinor != null) {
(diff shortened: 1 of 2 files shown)


src/interceptors/ClientRequest/MockHttpSocket.ts
```diff
@@ -340,7 +342,20 @@ export class MockHttpSocket extends MockSocket {
...
     // Transfer-Encoding, and other things to Node.js internals.
-    const serverResponse = new ServerResponse(new IncomingMessage(this))
+    const incomingMessage = new IncomingMessage(this)
+    
+    // Set the HTTP version on the IncomingMessage.
+    // This is critical for Node.js to correctly allow keep alive.
+    // By default, nodejs sets keepalive to false if the HttpVersion is not at least
+    // 1.1: https://github.com/nodejs/node/blob/70f6b58ac655234435a99d72b857dd7b316d34bf/lib/_http_server.js#L211C1-L211C34
+    // The version should normally already have been set by the onRequestStart method
+    if(this.httpVersionMajor != null && this.httpVersionMinor != null) {
```
kettanaito: Hm, this is interesting! From what I gathered, this `IncomingMessage` was purely internal to help us handle the mock response and shouldn't influence the user-facing behaviors in any way. So you say it does?

This might be an indicator that we should consider migrating from this.

All `ServerResponse` does here is help us parse out the response. User-facing `ClientRequest` constructs its own instances, including `IncomingMessage`, and ours never leaks or affects the user-facing data in any way. At least, that is the intention.

kettanaito:
## Released: v0.42.0 🎉

```
```

---

PR #759, opened 2025-12-19
feat: use `tsdown` for bundling

kettanaito:
- Use `tsdown` for bundling.
- Update `exports`, including at the nested paths, like `/presets`.
- Drop `alias` from tsdown config as it resolves aliases from `tsconfig` instead.
- Add a new `tsconfig.browser.json` to resolve the `internal:brotli-compress` correctly during the build. 
- Removes explicit `types` from export extensions to rely on implicit dts lookup. 

## Motivation

- Trusted publishing (#758) needs the new version of `ossjs/release`.
- The new version of `ossjs/release` uses `publint` for package linting.
- Publint complains on incorrect CJS/MJS type definition files mixing (actual issue).
- tsup doesn't produce different .d.ts for different formats, but tsdown does.

### DIFF
--- a/ClientRequest/package.json
+++ b/ClientRequest/package.json
@@ -2,17 +2,10 @@
   "main": "../lib/node/interceptors/ClientRequest/index.cjs",
   "module": "../lib/node/interceptors/ClientRequest/index.mjs",
   "browser": null,
-  "types": "../lib/node/interceptors/ClientRequest/index.d.cts",
   "exports": {
     ".": {
-      "import": {
-        "types": "./../lib/node/interceptors/ClientRequest/index.d.mts",
-        "default": "./../lib/node/interceptors/ClientRequest/index.mjs"
-      },
-      "default": {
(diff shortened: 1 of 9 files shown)


kettanaito:
## `/presets/node` no longer exports `default`

```ts
export default [
  new FetchInterceptor(),
] as const
```

This code no longer produces a named `default` export as it did with tsup. Rightfully, the way Nock gets the default value breaks:

https://github.com/nock/nock/blob/e7418da29feb4a7bf0aa1612bfb9d32a4051651e/lib/recorder.js#L10-L12

This can be fixed on Nock's side by rewriting the require to:

```ts
const nodeInterceptors = require('@mswjs/interceptors/presets/node') 
```

kettanaito:
## Released: v0.41.0 🎉

This has been released in v0.41.0.

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.41.0)
- 📦 [View on npm](https://www.npmjs.com/package/@mswjs/interceptors/v/0.41.0)

```
```

---

PR #766, opened 2026-02-12
fix(fetch): sanitize header tuples to fix Node.js 24 compatibility

### LINKED ISSUE #762: Node 24: Headers constructor: expected name/value pair to be length 2, found 3.
https://github.com/mswjs/interceptors/blob/9560baaacc7fb98864fc6ae6f474f1ae13dbc466/src/interceptors/ClientRequest/utils/recordRawHeaders.ts#L116-L120

This constructor throws the following error only in Node.js 24:

```
TypeError: Headers constructor: expected name/value pair to be length 2, found 3.
    at Object.webidl.errors.exception (node:internal/deps/undici/undici:4063:14)
    at fill (node:internal/deps/undici/undici:9177:33)
    at new Headers (node:internal/deps/undici/undici:9442:11)
    at Object.construct (file:///mswjs/msw/node_modules/.pnpm/@mswjs+interceptors@0.40.0/node_modules/@mswjs/interceptors/lib/node/chunk-5UGIB6OX.mjs:175:36)
    at normalizeResponseInit (/mswjs/msw/src/core/utils/HttpResponse/decorators.ts:20:19)
    at new HttpResponse (/mswjs/msw/src/core/HttpResponse.ts:48:26)
    at /mswjs/msw/test/node/rest-api/request/body/body-blob.node.test.ts:9:12
    at /mswjs/msw/src/core/handlers/RequestHandler.ts:341:24
    at HttpHandler.run (/mswjs/msw/src/core/handlers/RequestHandler.ts:322:28)
    at executeHandlers (/mswjs/msw/src/core/utils/executeHandlers.ts:42:14)
```

For some reason, recorded headers have an extra _third_ value stored, some boolean:

```ts
[ [ 'content-type', 'application/octet-stream', true ] ]
```

I suspect Node.js 24 started providing an extra parameter to some method call and we're naively recording it instead of trimming the values to a tuple.

KrantiPonala:
Node.js 24 internally passes a third boolean argument to Headers.prototype.set() and .append(). This extra argument was being recorded in raw header tuples, causing "Headers constructor: expected name/value pair to be length 2, found 3" errors when those tuples were later passed to the Headers constructor.

- Only record [name, value] in set/append proxies, ignoring extra args
- Sanitize header tuples when cloning Headers instances

Fixes #762

### DIFF
--- a/src/interceptors/ClientRequest/utils/recordRawHeaders.ts
+++ b/src/interceptors/ClientRequest/utils/recordRawHeaders.ts
@@ -122,7 +122,10 @@ export function recordRawFetchHeaders() {
         ) {
           // Ensure each header tuple has exactly 2 elements (name, value).
           // Node.js 24+ may have stored tuples with extra internal arguments.
-          const rawHeadersFromInit = Reflect.get(headersInit, kRawHeaders) as RawHeaders
+          const rawHeadersFromInit = Reflect.get(
+            headersInit,
+            kRawHeaders
+          ) as RawHeaders
           const sanitizedHeaders = rawHeadersFromInit.map(
             (tuple): HeaderTuple => [tuple[0], tuple[1]]
           )
--- a/src/interceptors/ClientRequest/utils/recordRawHeaders.ts
+++ b/src/interceptors/ClientRequest/utils/recordRawHeaders.ts
@@ -113,9 +113,15 @@ export function recordRawFetchHeaders() {
           headersInit instanceof Headers &&
           Reflect.has(headersInit, kRawHeaders)
         ) {
+          // Ensure each header tuple has exactly 2 elements (name, value).
+          // Node.js 24+ may have stored tuples with extra internal arguments.
+          const rawHeadersFromInit = Reflect.get(headersInit, kRawHeaders) as RawHeaders
+          const sanitizedHeaders = rawHeadersFromInit.map(
+            (tuple): HeaderTuple => [tuple[0], tuple[1]]
+          )
           const headers = Reflect.construct(
             target,
-            [Reflect.get(headersInit, kRawHeaders)],
+            [sanitizedHeaders],
             newTarget
           )
           ensureRawHeadersSymbol(headers, [
@@ -124,7 +130,7 @@ export function recordRawFetchHeaders() {
              * This prevents multiple Headers instances from pointing
              * at the same internal "rawHeaders" array.
              */
-            ...Reflect.get(headersInit, kRawHeaders),
+            ...sanitizedHeaders,
           ])
           return headers
         }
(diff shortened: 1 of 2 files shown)


src/interceptors/ClientRequest/utils/recordRawHeaders.ts
```diff
@@ -149,14 +155,20 @@ export function recordRawFetchHeaders() {
...
     },
   })
 
   Headers.prototype.append = new Proxy(Headers.prototype.append, {
     apply(target, thisArg, args: HeaderTuple) {
-      recordRawHeader(thisArg, args, 'append')
+      // Use only the first two arguments (name, value) to record raw headers.
+      // Node.js 24+ may pass additional internal arguments that should not
+      // be included in the raw headers array.
+      recordRawHeader(thisArg, [args[0], args[1]], 'append')
```
kettanaito: Do you have a reference to Node.js implementation of those additional arguments to `Headers`? What do they affect?

By not recording them, are we risking recreating a `Headers` instance that's not a 1-1 representation of the one recorded?

KrantiPonala:
Could you review the changes @kettanaito. This fix would help us move forward with node24 upgrade in our repo. thanks!

KrantiPonala:
hey @kettanaito , were you able to review this PR. Would need help in merging this. Kindly review at the earliest. So we could use node24 in our repo. This is a blocker for us to move forward.

jcaracciolo:
Looking for some movement on this @kettanaito 

I looked into the "third parameter" behavior that this PR addresses. In Node.js 24+, the public `Headers` constructor (`new Headers(...)`) is strict and only accepts 2-tuples (`[name, value]`). Passing 3-tuples now throws:

> Headers constructor: expected name/value pair to be length 2, found 3

The extra value we were seeing is not part of the Fetch API. It comes from Undici's internal header handling (`HeadersList`). For example, Undici uses an internal flag in redirect logic:

```js
request.headersList.delete("proxy-authorization", true);
```

That second argument (`true`) is an internal signal for proxy-authentication handling. It is not part of the public Fetch `Headers` API. You can see similar vendored Undici logic [here](https://gitea.psi.ch/actions/setup-python/commit/9a7ac94420f42ee15fc60ab88d0dca4be1fd5757.patch#:~:text=https%3A%2F%2Ffetch.spec.whatwg.org%2F%23cors,headers%2C%20which%20undici%20doesn%27t%20implement)

This behavior implements the most likely the Fetch spec's authentication entries rules: https://fetch.spec.whatwg.org/#authentication-entries. Related Undici advisory: https://github.com/nodejs/undici/security/advisories/GHSA-3787-6prv-h9w3

Because the Fetch standard only defines headers as `(name, value)` pairs, normalizing headers (i.e., stripping additional metadata before constructing a `Headers` instance) is correct and expected.
Dropping the third value does not change observable behaviour. The proxy/auth distinction is re-derived internally by Undici based on the header name (e.g. `proxy-authorization`).

jcaracciolo:
@kettanaito Bumping this so it doesn't get closed out

kettanaito (approved):

kettanaito:
Thanks for fixing this, @KrantiPonala. We don't currently have any CI on Node.js 24 and I can see some other things failing on that version. This is a good start.

kettanaito:
## Released: v0.41.7 🎉

This has been released in v0.41.7.

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.41.7)
- 📦 [View on npm](https://www.npmjs.com/package/@mswjs/interceptors/v/0.41.7)

```
```

---

PR #666, opened 2024-10-30
fix(ClientRequest): support `addRequest` from custom http agents

kettanaito:
- Originates from https://github.com/mswjs/msw/issues/2338

PR #621, opened 2024-08-13
feat(wip): support "CONNECT" request

kettanaito:
https://github.com/nodejs/node/blob/29ec7e9331c4944006ffe28e126cc31cc3de271b/lib/_http_client.js#L568

PR #560, opened 2024-04-18
fix(MockHttpSocket): support keepalive (reused) sockets

kettanaito:

PR #770, opened 2026-02-23
feat!: TCP/TLS wrap-based interception

### LINKED ISSUE #481: Proxy URL is not generated correctly and causes Invalid URL error from node:internal
MSW is a great library and I have used it successfully in other "got" calls but in one part of my unit testing I was actually checking that a proxy call would actually try and use the proxy.  Unfortunately I think the way that the path and proxy information is generated results in an invalid URL which bails before you can get a matcher to handle it.  The proxy host is correctly assigned but then the path(the actual url we call) is just directly concatenated as a path.

```

   // From getUrlByRequestOptions.ts:169
   =============
   const url = new URL(`${protocol}//${hostname}${path}`)
   url.username = credentials?.username || ''
   url.password = credentials?.password || ''

   // ERROR
   =============
          input: 'http://fake.proxy:443fake.ping:443',
          code: 'ERR_INVALID_URL',

            at new URL (node:internal/url:783:36)
            at getUrlByRequestOptions (/node_modules/.pnpm/@mswjs+interceptors@0.25.7/node_modules/@mswjs/interceptors/lib/node/chunk-YVNH3GJ5.js:640:15)
            at normalizeClientRequestArgs (/node_modules/.pnpm/@mswjs+interceptors@0.25.7/node_modules/@mswjs/interceptors/lib/node/chunk-YVNH3GJ5.js:766:11)

 
// From the DEBUG:
=============
3:51:32:307 [http normalizeClientRequestArgs] successfully cloned RequestOptions! {"rejectUnauthorized":false,"agent":{"_events":{},"_eventsCount":2,"defaultPort":443,"protocol":"https:","options":{"keepAlive":true,"keepAliveMsecs":1000,"maxSockets":256,"maxFreeSockets":256,"scheduling":"lifo","noDelay":true,"path":null},"requests":{},"sockets":{},"freeSockets":{},"keepAliveMsecs":1000,"keepAlive":true,"maxSockets":256,"maxFreeSockets":256,"scheduling":"lifo","maxTotalSockets":null,"totalSocketCount":0,"maxCachedSessions":100,"_sessionCache":{"map":{},"list":[]},"proxy":"http://proxyUsername:proxyPassword@fake.proxy:443/"},"setHost":true,"method":"GET","headers":{"user-agent":"got (https://github.com/sindresorhus/got)","accept-encoding":"br, gzip, deflate"}}
13:51:32:307 [http normalizeClientRequestArgs] derived request options: {"method":"GET","protocol":"https:","hostname":"fake.ping","host":"fake.ping","path":"/pinging","rejectUnauthorized":false,"agent":{"_events":{},"_eventsCount":2,"defaultPort":443,"protocol":"https:","options":{"keepAlive":true,"keepAliveMsecs":1000,"maxSockets":256,"maxFreeSockets":256,"scheduling":"lifo","noDelay":true,"path":null},"requests":{},"sockets":{},"freeSockets":{},"keepAliveMsecs":1000,"keepAlive":true,"maxSockets":256,"maxFreeSockets":256,"scheduling":"lifo","maxTotalSockets":null,"totalSocketCount":0,"maxCachedSessions":100,"_sessionCache":{"map":{},"list":[]},"proxy":"http://proxyUsername:proxyPassword@fake.proxy:443/"},"setHost":true,"headers":{"user-agent":"got (https://github.com/sindresorhus/got)","accept-encoding":"br, gzip, deflate"}}
13:51:32:307 [http normalizeClientRequestArgs] has no default agent, setting the default agent for "https:"
13:51:32:307 [http normalizeClientRequestArgs] successfully resolved url: https://fake.ping/pinging
13:51:32:307 [http normalizeClientRequestArgs] successfully resolved options: {"method":"GET","protocol":"https:","hostname":"fake.ping","host":"fake.ping","path":"/pinging","rejectUnauthorized":false,"agent":{"_events":{},"_eventsCount":2,"defaultPort":443,"protocol":"https:","options":{"keepAlive":true,"keepAliveMsecs":1000,"maxSockets":256,"maxFreeSockets":256,"scheduling":"lifo","noDelay":true,"path":null},"requests":{},"sockets":{},"freeSockets":{},"keepAliveMsecs":1000,"keepAlive":true,"maxSockets":256,"maxFreeSockets":256,"scheduling":"lifo","maxTotalSockets":null,"totalSocketCount":0,"maxCachedSessions":100,"_sessionCache":{"map":{},"list":[]},"proxy":"http://proxyUsername:proxyPassword@fake.proxy:443/"},"setHost":true,"headers":{"user-agent":"got (https://github.com/sindresorhus/got)","accept-encoding":"br, gzip, deflate"},"_defaultAgent":{"_events":{},"_eventsCount":2,"defaultPort":443,"protocol":"https:","options":{"keepAlive":true,"scheduling":"lifo","timeout":5000,"noDelay":true,"path":null},"requests":{},"sockets":{},"freeSockets":{},"keepAliveMsecs":1000,"keepAlive":true,"maxSockets":null,"maxFreeSockets":256,"scheduling":"lifo","maxTotalSockets":null,"totalSocketCount":0,"maxCachedSessions":100,"_sessionCache":{"map":{},"list":[]}}}
13:51:32:308 [http normalizeClientRequestArgs] successfully resolved callback: undefined
13:51:32:310 [http request] request call (protocol "http"): [{"method":"CONNECT","host":"fake.proxy","port":"443","path":"fake.ping:443","setHost":false,"headers":{"connection":"keep-alive","host":"fake.ping:443","proxy-authorization":"Basic cHJveHlVc2VybmFtZTpwcm94eVBhc3N3b3Jk"},"agent":false}]
13:51:32:310 [http normalizeClientRequestArgs] arguments [{"method":"CONNECT","host":"fake.proxy","port":"443","path":"fake.ping:443","setHost":false,"headers":{"connection":"keep-alive","host":"fake.ping:443","proxy-authorization":"Basic cHJveHlVc2VybmFtZTpwcm94eVBhc3N3b3Jk"},"agent":false}]
13:51:32:310 [http normalizeClientRequestArgs] using default protocol: http:
13:51:32:310 [http normalizeClientRequestArgs] first argument is RequestOptions: {"method":"CONNECT","host":"fake.proxy","port":"443","path":"fake.ping:443","setHost":false,"headers":{"connection":"keep-alive","host":"fake.ping:443","proxy-authorization":"Basic cHJveHlVc2VybmFtZTpwcm94eVBhc3N3b3Jk"},"agent":false}
13:51:32:310 [http normalizeClientRequestArgs] normalized request options: {"method":"CONNECT","host":"fake.proxy","port":"443","path":"fake.ping:443","setHost":false,"headers":{"connection":"keep-alive","host":"fake.ping:443","proxy-authorization":"Basic cHJveHlVc2VybmFtZTpwcm94eVBhc3N3b3Jk"},"agent":false,"protocol":"http:"}
13:51:32:310 [utils getUrlByRequestOptions] request options {"method":"CONNECT","host":"fake.proxy","port":"443","path":"fake.ping:443","setHost":false,"headers":{"connection":"keep-alive","host":"fake.ping:443","proxy-authorization":"Basic cHJveHlVc2VybmFtZTpwcm94eVBhc3N3b3Jk"},"agent":false,"protocol":"http:"}
13:51:32:310 [utils getUrlByRequestOptions] figuring out url from request options...
13:51:32:310 [utils getUrlByRequestOptions] protocol http:
13:51:32:310 [utils getUrlByRequestOptions] host fake.proxy
13:51:32:310 [utils getUrlByRequestOptions] port 443
13:51:32:310 [utils getUrlByRequestOptions] hostname fake.proxy:443
13:51:32:310 [utils getUrlByRequestOptions] path fake.ping:443
13:51:32:310 [utils getUrlByRequestOptions] credentials undefined
13:51:32:310 [utils getUrlByRequestOptions] auth string:
```

Code that pretty much reproduces it.
```
// Pseudo code reproduction
const { HttpsProxyAgent } = require('hpagent');
const host = 'fake.proxy';
const username = "proxyUsername";
const password = "proxyPassword";
const port = '443';

   let proxy =  HttpsProxyAgent({
      keepAlive: true,
      keepAliveMsecs: 1000,
      maxSockets: 256,
      maxFreeSockets: 256,
      scheduling: 'lifo',
      proxy: `http://`${username}:${password}@${host}:${port}`,
    });


  // Then used in a got call like this:
 export const PING_URL  = "https://fake.ping/pinging";
 got.get({url: PING_URL, agent: { https: proxy})

```

### LINKED ISSUE #707: Support raw `net.connection`

### LINKED ISSUE #713: MockHttpSocket does not expose localAddress or remoteAddress
The MockHttpSocket implementation does not expose localAddress or remoteAddress. It can be reproduced below:

```js
const {ClientRequestInterceptor} = require('@mswjs/interceptors/ClientRequest');
const http = require('http');

const printAddresses = ()=>new Promise((resolve, reject)=>{
    const req = http.get('http://example.com');
    req.once('response', res=>{
        console.log('Addresses: ',
            {localAddress: res.socket.localAddress, remoteAddress: res.socket.remoteAddress});
        req.destroy();
        resolve();
    });
    req.once('error', reject);
});

const test = async ()=>{
    console.log('Before interceptor');
    await printAddresses();
    const interceptor = new ClientRequestInterceptor();
    interceptor.apply();
    console.log('After interceptor');
    await printAddresses();
};

test();
```

This outputs the following:

```
Before interceptor
Addresses:  { localAddress: '10.0.2.15', remoteAddress: '96.7.128.175' }
After interceptor
Addresses:  { localAddress: undefined, remoteAddress: undefined }
```

I would expect the mock implementation to expose the same properties as the original socket. For now, the workaround I'm using is accessing it through `socket.originalSocket.localAddress` but it is not great as it requires deeper knowledge of the mock implementation. If it's relevant, I'm experiencing this while using the [nock](https://www.npmjs.com/package/nock) library which uses this library internally.

### LINKED ISSUE #725: Use `llhttp` instead of `HTTPParser` from `_http_common`
The HTTP parser we're using right now is built in into Node.js but it's a part of Node's public API. Relying on it potentially dangerous as it might yield different behaviors across Node.js versions. 

Instead, we can use the WASM build of `llhttp`, which is also the HTTP parser used by Undici:

https://github.com/nodejs/undici/blob/219da8b7b3fea7e38a7644b8bc35fe6fec97d66e/lib/dispatcher/client-h1.js#L139

I've heard that Node's `HTTPParser` uses `llhttp` under the hood so, hopefully, this refactoring isn't that difficult.

### LINKED ISSUE #802: Bug: socket.address() returns an empty object {} on mocked HTTPS requests
When intercepting an HTTPS request using `ClientRequestInterceptor` and providing a mocked response via `controller.respondWith()`, the socket emitted by the request returns an empty object `{}` when calling `socket.address()`.

In a standard Node.js environment, `socket.address()` is expected to return an object containing address details (such as `address`, `family`, and `port`). When the request is mocked, this metadata is missing, which can break downstream libraries or application logic that relies on inspecting the socket's connection details.

```js
import http from 'node:http';
import https from 'node:https';
import zlib from 'node:zlib';
import { ClientRequestInterceptor } from "@mswjs/interceptors/ClientRequest";

interceptor.apply()

interceptor.on('request', async ({ controller }) => {
  controller.respondWith(new Response('mocked response', { status: 200 }))
})
const request = https.get('https://example.com', { rejectUnauthorized: false });

request.once('socket', socket => {
  // Both of these print {} instead of socket address details
  socket.on('secureConnect', (...args) => console.log('secureConnect', socket.address()))
  socket.on('connect', (...args) => console.log('connect', socket.address()))
})
```
### Expected Behavior

Even when a response is mocked, the mocked socket should provide a fallback stub or realistic mock data for `socket.address()` rather than an empty object. For example, it should return something like:
```json
{
  "address": "127.0.0.1",
  "family": "IPv4",
  "port": 443
}
```

### LINKED ISSUE #2569: MSW mock server not handling `secureConnect`
### Prerequisites

- [x] I confirm my issue is not in the [opened issues](https://github.com/mswjs/msw/issues)
- [x] I confirm the [Frequently Asked Questions](https://mswjs.io/docs/faq) didn't contain the answer to my issue

### Environment check

- [x] I'm using the [latest](https://github.com/mswjs/msw/releases/latest) `msw` version
- [x] I'm using Node.js version 20 or higher

### Node.js version

v22.13.1

### Reproduction repository

https://github.com/juicemia/repro-msw-secureconnect-issue

### Reproduction steps

```
pnpm i
pnpm test
```

### Current behavior

The `secureConnect` event is not being correctly handled. HTTP requests made using the `NodeHttpClient` in the reproduction repository will hang indefinitely.

### Expected behavior

I expected HTTP requests made using the `NodeHttpClient` in the reproduction repository to complete successfully.

### LINKED ISSUE #2617: MSW emits 2 error events when ClientRequest is destroyed
### Prerequisites

- [x] I confirm my issue is not in the [opened issues](https://github.com/mswjs/msw/issues)
- [x] I confirm the [Frequently Asked Questions](https://mswjs.io/docs/faq) didn't contain the answer to my issue

### Environment check

- [x] I'm using the [latest](https://github.com/mswjs/msw/releases/latest) `msw` version
- [x] I'm using Node.js version 20 or higher

### Node.js version

v22.21.0

### Reproduction repository

https://github.com/markuslewin/msw-http-error

### Reproduction steps

```sh
npm test
```

### Current behavior

When `ClientRequest.destroy` is called with an error, **2** error events are emitted to the error listener. This results in an uncaught exception for HTTP clients that only expect 1 error, such as the clients in the [Azure SDK for JavaScript](https://github.com/Azure/azure-sdk-for-js/blob/4219296463e6ce7f1e56fab1d8a4769acf7afcd6/sdk/core/ts-http-runtime/src/nodeHttpClient.ts#L240-L244).

If I disable MSW by removing the line `server.listen()`, the tests pass, but when MSW is enabled, I receive an uncaught exception for the second error event:

```sh
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
AbortError: The operation was aborted. Rejecting from abort signal callback while making request.
 ❯ AbortSignal.<anonymous> node_modules/@typespec/ts-http-runtime/dist/esm/nodeHttpClient.js:199:36
 ❯ AbortSignal.[nodejs.internal.kHybridDispatch] node:internal/event_target:827:20
 ❯ AbortSignal.dispatchEvent node:internal/event_target:762:26
 ❯ runAbort node:internal/abort_controller:486:10
 ❯ abortSignal node:internal/abort_controller:457:3
 ❯ AbortController.abort node:internal/abort_controller:505:5
 ❯ Timeout._onTimeout node_modules/@typespec/ts-http-runtime/dist/esm/nodeHttpClient.js:85:33
 ❯ listOnTimeout node:internal/timers:588:17
 ❯ processTimers node:internal/timers:523:7

This error originated in "app.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
The latest test that might've caused the error is "azure credential doesn't throw". It might mean one of the following:
- The error was thrown, while Vitest was running this test.
- If the error occurred after the test had been completed, this was the last documented test before it was thrown.
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
```

### Expected behavior

MSW should emit only 1 error event when `ClientRequest.destroy` is called with an error, mimicking the behavior of the original `http.ClientRequest`.

kettanaito:
- Related to #483, #741 (prior art)
- Closes #707 
- Fixes https://github.com/mswjs/msw/issues/2617
- Fixes https://github.com/mswjs/msw/issues/2569
- Closes #697
- Fixes #745
- Closes #560
- Closes #621 
- Closes #643 
- Closes #666 (irrelevant)
- Fixes #770
- Fixes #481 
- Fixes #713
- Closes #725
- Closes #756 (fixed by the arhictecture)
- Fixes #802

## Changes

- Implements a `SocketInterceptor` that intercepts the outgoing socket packets on the `net.connect()`/`tls.connect()` levels. The interception mostly works via a "server-side socket" instance and patching TCPWrap/TLSWrap JavaScript bindings for C++ network code (`socket._handle`) to pend the connection until the interceptor decides what to do with it.
- Implements a new `HttpRequestInterceptor` to replace `ClientRequestInterceptor`.

## ⚠️ Breaking changes

- **Node.js v22 is the minimal supported version**.
- `getRawRequest` utility has been removed. Use the `initiator` reference directly.
- `/utils/node` and `getClientRequestBodyStream` has been removed. Use the `initiator` reference directly.
- `file://` protocol requests are unsupported with `fetch` the same way they aren't supported in Node.js 22 and 24 (improved compliance).
- HTTP-to-WebSocket upgrade requests throw with `fetch` the same way they do in Node.js. Other upgrade paths are unaffected and function correctly (`http.request`, `WebSocket`, `undici`).

## Features

- Supports Node.js v26.
- Supports `CONNECT` requests (fixes #481, closes #621).
- Supports Undici `fetch` and `request` even when they aren't global.
- Exposes the request `initiator` on the "request" event listener. The `initiator` property points to `ClientRequest`, `XMLHttpRequest`, or `Request` instance that initiated the intercepted request. If no instance can be reliably deduced, points to the raw `net.Socket` behind the connection. 

## Bug fixes

- The library now executes more of the Node.js network that ever before. In fact, it executes the biggest possible amount of Node.js code. We are literally patching before the C++ bindings.
- ClientRequest using sockets is now properly supported.
- Now it doesn't matter how you import `http`/`https` (e.g. `import * as http` broke previously since we cannot patch that).
- You can now end requests in the write callbacks. The documented limitation of Interceptors is gone.
- Clients waiting for the `secureConnect` before writing anything are now properly supported.
- TLS connections now correctly emit the `keylog` event on the socket.
- Fixes an issue where the automatic `Content-Type` header added by Undici was not set on mocked responses.
- `socket.address()`, as well as `localAddress`, `localPort`, `localFamily`, `remoteAddress`, `remotePort`, and `remoteFamily, are correctly set for both mocked and bypassed connections.
- XMLHttpRequest: Fixes an issue where redirect responses weren't being followed.
- XMLHttpRequest: Fixes an issue where the "progress" event for `request.upload` wasn't dispatched correctly.
- XMLHttpRequest: `total` response body length is now computed _only_ from the `content-length` mocked response header, just like it's done natively.
- XMLHttpRequest: The `progress` event is now correctly emitted on the request for chunked responses.
- XMLHttpRequest (JSDOM): Correctly performs the `OPTIONS` preflight request based on the JSDOM implementation (previously, preflight triggered for requests that shouldn't trigger it).
- XMLHttpRequest: Implements mock response handling in accordance to the WHATWG XMLHttpRequest specification to improve browser compliance. 

## Todos

- [x] Migrate all the existing tests to the new interceptor.
- [x] Add a thorough compliance test suite for `SocketInterceptor`. Cover behaviors like socket re-usage, event emission, properties (especially on TLS sockets), etc.
  - [ ] [`_tlsError`](https://github.com/nodejs/node/blob/abddfc921bf2af02a04a6a5d2bca8e2d91d80958/lib/internal/tls/wrap.js#L1054) emission
- [x] TLS: The `session` event shouldn't be emitted for reused sessions. Mocks always emit it now. `options.session` should be respected. Afaik, if provided, _that_ is a pointer to a previously active session.
- [x] Add a test for `CONNECT` requests (see #481 / #621). Might they be supported now?
- [x] **Finish logging**.
- [x] Add tests to ensure that multiple interceptors using `SocketInterceptor` can run at the same time without conflicting.
- [x] Bug: `TypeError: Second argument must be a buffer`. Reproduce: `pnpm run test:node test/modules/http` _multiple_ times in a row. Flaky.
- [x] Add http compliance test that calls `req.end()` in the socket's `connect` event. 
- [x] **Support global `fetch`** (Undici fetch) in the `HttpRequestInterceptor` (currently times out).
- [x] Ensure passthrough HTTPs for global `fetch` works. Our test server has an expired certificate and `fetch` doesn't provide the means to set `rejectUnauthorized: false` so requests fail.
- [x] Consider if `HttpRequestInterceptor` needs to have the response decompression and redirect following logic. It likely doesn't. Just wonder how we can implement `FetchInterceptor` on top of it then. 
  - Decompression—**no**. Undici will do that for us. Confirmed: completely gone from Node.js, left for Fetch web since nothing will decompress it for us there.
  - Redirects—also **no**. Similarly to decompression, everything in Node.js will use its own redirects following logic from now. Kept only for the browser for the same reason.
- [x] Implement the `initiator` property on the `request` event.
  - [x] `http`
  - [x] `XMLHttpRequest`
  - [x] `fetch`
  - [x] Remove `getRawRequest`
- [x] For the browser versions of fetch and XHR interceptors, still provide `initiator`. Refactor what `setRawRequest` did before.
- [x] **Export node and inject versions of XHR and fetch interceptors**. In Node.js, `HttpRequestInterceptor` now intercepts _everything_ and more specialized interceptors only provide the `initiator` via async context. In the browser, however, the `/inject` version of the interceptors has to be used instead. 
- [x] Remove `controller.errorWith()` because it's more semantically valid to use `socket.destroy()`. It provides no additional functionality and is a 1-to-1 replacement.
- [x] `applyPatch`: Use `Symbol.for()` instead of `Symbol()` to account for isolated environments (e.g. browser extensions (see https://github.com/mswjs/interceptors/issues/771). 
- [x] Add export conditions for fetch/XHR interceptors:
  - In the browser, resolve to the patch version.
  - In Node.js, use the `/node.ts` version (more powerful). 
- [x] "finish" is not emitted when the socket writes end. A bug. Causes the related test to fail and also results in "Premature close" from some XHR scenarios since close while writable means premature close.
- [x] Add lazy `response`/`unhandledRequest` listener forwarding by checking `this.emitter.listenerCount(event)` inside the `request` event attached to the nested interceptor singleton. 
- [x] Refactor `interceptorRegistry` from module scope to `globalThis`.
- [x] Modify passthrough HTTP headers only if the request headers were changed in the request listener.
- [x] `BatchInterceptor.symbol` reassignment in constructor affects the entire class, not just this instance.
- [x] ~~Make request processing conditional based on `request` listener count (the same as response).~~ Will be useful for on-demand interceptors in MSW in the future. 
- [x] **Rewrite `fetch` tests to be neutral like XHR**.
- [x] ⚠️ **Migrate to use `llhttp` the same way Undici uses it**. That's the most performant and reliable way of introducing a parser to the stack. (#781)
- [x] Bug: Undici doesn't emit `free` on the socket, it uses an internal pool. `SocketInterceptor` never transitions into a `PENDING` state on the subsequent request (fixed in 65052ba307c88083d9472c2e03c3a9158ad5a8b9).
- [x] Should `HttpInterceptor` even be public? Combining it with any other interceptor seems to fail.  
- [x] Fix the usage of `__dirname` caused by `llhttp` or its surrounds.
- [x] Add tests for the `response` listener of a compressed `fetch` response. Should the response in the listener be compressed? Is it the same for mocked/passthrough responses?  Refine the expectations. 
- [x] Remove `src/utils/canParseUrl.ts`, use `URL.canParse()`. We are at Node.js 22+ now.
- [x] Align failed to fetch errors to be identical to those in Node.js and the browser (they are neither right now). 
- [x] Decide whether to support `file://` requests in Node.js.
- [x] Migrate from `net.connect` patching to `Socket.prototype.connect` patching to circumvent wildcard imports in ESM (cannot mutate those imports).


kettanaito:
## Nock tests

Nock tests will be broken until we migrate it to use `initiator` instead of `getRawRequest`. That's a breaking change so test failures are expected here.

kettanaito:
## Compressed responses in the `response` event

> This is relevant only for _compressed_ responses.

Following the minimal intrusion principle, the library doesn't meddle with the response exposed in the `response` listener. That means:

- In Node.js, the `response` is compressed, just like the server sends it.
- In the browser, the `response` is *de*compressed. While we can skip the decompression from mocked responses, the real responses arrive _from the browser_ (not server) and the browser automatically decompresses them without exposing any means of getting the original server-sent stream. For environment-based consistency, both response instances will be decompressed.

kettanaito:
@coderabbitai please review this.

kettanaito:
## Released: v0.42.0 🎉

```
```

---

PR #777, opened 2026-04-25
fix(fetch): add react-native export condition to skip decompression

### LINKED ISSUE #2367: Following React Native docs does not work and Property 'TransformStream' doesn't exist on React Native (runtime)
### Prerequisites

- [X] I confirm my issue is not in the [opened issues](https://github.com/mswjs/msw/issues)
- [X] I confirm the [Frequently Asked Questions](https://mswjs.io/docs/faq) didn't contain the answer to my issue

### Environment check

- [X] I'm using the [latest](https://github.com/mswjs/msw/releases/latest) `msw` version
- [X] I'm using Node.js version 18 or higher

### Node.js version

v18.15.0

### Reproduction repository

https://github.com/lgibso34/react-native-msw

### Reproduction steps

Similar to #2344 but not Jest related.

1. checkout `main` branch
2. npm start
3. launch on device

TransformStream error

1.` git checkout bug/sync`
2. npm start
3. launch on device

### Current behavior

Bug 1: Seems that React Native does not support the dynamic import syntax and fails to register the component in time.

Bug 2: Property 'TransformStream' doesn't exist on React Native at runtime running the synchronous code. This would also occur if the async version (bug 1) worked

### Expected behavior

Bug 1: I expect the docs to explain how to get this working with additional metro config or a complete example with Metro and RN CLI. I expect the docs to accurately describe how to get msw to work in React Native.

Bug 2: I expect this to work on React Native like the docs suggest.

Async error (main)
not_registered

Sync error (bug/sync)
TransformStream no exist

vladlenskiy:
On React Native with Hermes, importing `@mswjs/interceptors/fetch` throws at module load:

```
ReferenceError: Property 'TransformStream' doesn't exist
```

`decompression.ts` declares `class PipelineStream extends TransformStream` at the top level. Hermes has no Web Streams, so evaluating the module fails before any compressed response is involved. This follows the direction suggested in mswjs/msw#2367.

Changes:
- add a `lib/native/` build using `tsconfig.react-native.json`
- alias `internal:decompression` to a native no-op that returns `null`
- import decompression through `internal:decompression` in the fetch interceptor
- add the `react-native` export condition for `.` and `./fetch`, before `browser`

MSW handlers do not set `Content-Encoding`, so returning `null` keeps the existing body unchanged.

Checks:
- `pnpm test:unit`: 205 tests pass
- `pnpm build` creates `lib/node/`, `lib/browser/`, and `lib/native/`
- built native JS has no `TransformStream`, `DecompressionStream`, or `BrotliDecompressionStream` references, excluding sourcemaps
- `node --conditions=react-native -e "delete globalThis.TransformStream; delete globalThis.DecompressionStream; require('@mswjs/interceptors/fetch')"` exits 0

Only `fetch` is included in the native build because that is what `msw/native` consumes today. The other Hermes issues in mswjs/msw#2367 can be handled separately.

Closes mswjs/msw#2367


package.json
```diff
@@ -8,6 +8,10 @@
   "types": "./lib/node/index.d.cts",
   "exports": {
     ".": {
+      "react-native": {
```
kettanaito, line 11: Is there anything in the root that breaks React Native also? I'm wondering if a different export for the root is even needed.

src/interceptors/fetch/index.ts
```diff
@@ -9,7 +9,7 @@ import { canParseUrl } from '../../utils/canParseUrl'
 import { createRequestId } from '../../createRequestId'
 import { createNetworkError } from './utils/createNetworkError'
 import { followFetchRedirect } from './utils/followRedirect'
-import { decompressResponse } from './utils/decompression'
+import { decompressResponse } from 'internal:decompression'
```
kettanaito, line 12: Let's keep the branching behavior down to `brotli-decompress` module only. It's already there. Let's add another module `brotli-decompress-noop.ts` that will act as a passthrough for the stream without relying on `TransformStream`. 

Tapping on the `./decompression.ts` level is too aggressive.

kettanaito:
I think the actual solution to this is to include the web streams polyfill in your test environment, alongside other polyfills for the things React Native doesn't implement, like the `URL` class.

kettanaito:
Hi, @vladlenskiy. Thanks for opening this!

I believe this is a wrong way to fix the problem. Interceptors mustn't be aware of React Native in any way. Instead, I propose developers use the web streams polyfill for React Native tests if they want to use MSW there (see https://github.com/mswjs/msw/issues/2367#issuecomment-4358830184).

PR #781, opened 2026-05-04
feat: use `llhttp` parser

mikicho:


kettanaito:
Alright, I think this branch is ready! Let's double-check with the CI before merging. Thanks, @mikicho 👏

PR #782, opened 2026-05-07
fix: support responses exceeding high watermark

mikicho:
1. Fix interceptor reapply
2. Forward drain event to the server response socket
3. Fixed socket.address family string case


test/modules/http/response/http-https.test.ts
```diff
@@ -146,3 +146,17 @@ it('bypasses any request after the interceptor was restored', async () => {
   expect(response.statusText).toBe('OK')
   await expect(response.text()).resolves.toBe('original-response')
 })
+
+it('big response', async () => {
```
kettanaito: Can you tell us more what was the problem big responses were exhibiting? Was there some error being thrown? Let's give this test case a descriptive name.

test/modules/http/response/http-https.test.ts
```diff
@@ -146,3 +146,17 @@ it('bypasses any request after the interceptor was restored', async () => {
...
+  const responseBody = new Array(1024 * 1024 + 1).join('.')
```
kettanaito: Wouldn't it be better to create a `Uint8Array` of the same size instead?

src/interceptor.ts
```diff
@@ -54,10 +54,6 @@ export abstract class Interceptor<
   protected abstract setup(): void
 
   public apply(): void {
-    if (this.readyState === InterceptorReadyState.DISPOSED) {
```
kettanaito, line 57: But we still need a guard to prevent calling `apply()` when the interceptor is already applied. I believe the fix here is to change:

```ts
if (this.readyState === InterceptorReadyState.ACTIVE) {
```

src/Interceptor.test.ts
```diff
@@ -192,3 +192,20 @@ it('removes all listeners when the interceptor is disposed', () => {
   interceptor['emitter'].emit(new TypedEvent('test'))
   expect(listener).not.toHaveBeenCalled()
 })
+
+it('can apply after disposal', () => {
```
kettanaito:
```suggestion
it('applies the interceptor after disposal', () => {
```

src/Interceptor.test.ts
```diff
@@ -192,3 +192,20 @@ it('removes all listeners when the interceptor is disposed', () => {
...
+  class MyInterceptor extends Interceptor<{ test: TypedEvent }> {
```
kettanaito:
```suggestion
  class MyInterceptor extends Interceptor<{}> {
```

Or `any`, if it accepts it.

src/Interceptor.test.ts
```diff
@@ -192,3 +192,20 @@ it('removes all listeners when the interceptor is disposed', () => {
...
+    protected predicate(): boolean {
+      return true
+    }
+
+    protected setup(): void {}
+  }
+  const interceptor = new MyInterceptor()
+
+  interceptor.apply()
```
kettanaito: Do we also need `readyState` assertions after individual actions? That way, we can guard against false positive/negatives by making sure every state transitions indeed produces the correct state (`readyState`).

kettanaito (changes requested):

test/modules/http/response/http-https.test.ts
mikicho: At the consumer level, it checks whether we support 1MB responses. On a technical level, it verifies that we properly forward drain events for streams larger than the high water mark.

kettanaito (approved):
Looks good to me!

I polished a few things, fixed the `interceptors.test.ts` file name casing.

PR #784, opened 2026-05-09
fix(fetch): detach abort listener from request.signal after settle

### LINKED ISSUE #783: Per-call memory retention regression in FetchInterceptor between 0.36.10 and 0.40.0 (still present in 0.41.8)
Upfront I want to say thank you for creating MSW. I use it in multiple Projects, such pleasure!

> [!NOTE]
> Due to time constraints the following analysis was drafted with the help of an AI assistant. The reproduction, numbers, and version comparisons have been triple-checked against the linked repo and against my own production heap snapshots before filing.

## Summary

`FetchInterceptor` retains substantially more state per intercepted `fetch()` call starting in **0.40.0**, and the regression is still present in **0.41.8**. The retention persists across explicit `global.gc()` and is in addition to whatever undici keeps in flight. In production this surfaces as a slow but unbounded heap/RSS climb on a Nuxt 4 SSR server using `@artmizu/nuxt-prometheus`, which applies a `BatchInterceptor` to every server-side `fetch()`. Downgrading to `0.36.10` (or disabling the interceptor entirely) restores flat memory.

## Environment

- Node `v24.15.0` on macOS (also reproduces on our prod containers — Linux x64)
- `@mswjs/interceptors` `0.36.10`, `0.40.0`, `0.41.8`
- No mock handlers attached. The interceptor is applied with all three sub-interceptors (Fetch + ClientRequest + XMLHttpRequest), no `request`/`response` listeners.

## Reproduction

Self-contained repro repo: **https://github.com/PatrikBird/mswjs-interceptors-leak-repro**

```bash
git clone https://github.com/PatrikBird/mswjs-interceptors-leak-repro.git
cd mswjs-interceptors-leak-repro

pnpm add @mswjs/interceptors@0.36.10 && node --expose-gc repro.mjs v0.36.10
pnpm add @mswjs/interceptors@0.40.0  && node --expose-gc repro.mjs v0.40.0
pnpm add @mswjs/interceptors@latest  && node --expose-gc repro.mjs v0.41.8
node --expose-gc repro.mjs baseline 8000 0    # interceptor disabled, control
```

The harness spins up a small local HTTP server that returns `302` → unique URL (mimicking a redirect-follow flow), runs `N` concurrent `fetch()` calls with `redirect: 'follow'`, forces GC, and reports `process.memoryUsage()` deltas + a V8 heap snapshot. See the README for full details and a snapshot scanner (`count.mjs`).

## Results (8,000 calls, redirect: follow, concurrency 20, post-GC)

| Version             | heapUsed Δ | RSS Δ  | Δ over baseline |
| ------------------- | ---------- | ------ | --------------- |
| baseline (no apply) | 2.46 MB    | 205 MB | —               |
| **0.36.10**         | 22.9 MB    | 255 MB | +20.4 MB        |
| **0.40.0**          | 30.9 MB    | 263 MB | +28.4 MB        |
| **0.41.8** (latest) | 30.9 MB    | 267 MB | +28.5 MB        |

Retained instance counts from a `v8.writeHeapSnapshot()` after the same workload (post-GC):

|                            | baseline | 0.36.10 | 0.40.0 | 0.41.8 |
| -------------------------- | -------: | ------: | -----: | -----: |
| `URLContext`               |        5 |   3,614 |  5,743 |  5,781 |
| `URL`                      |        6 |   3,615 |  5,744 |  5,782 |
| `_HeadersList`             |        6 |   3,616 |  5,745 |  5,783 |
| `Request` / `FetchRequest` |        3 |   3,612 |  2,872 |  2,889 |
| `RequestController`        |        0 |   3,610 |    0\* |  2,889 |
| `DeferredPromise`          |        0 |  14,440 | 11,480 | 11,556 |

\* 0.40.0 retains the controller via closures rather than as a directly counted instance — the rise in `URL` / `_HeadersList` shows the same retention is present.

The shape that changes between 0.36.10 and 0.40.x:

- **0.36.10**: `URL ≈ Request ≈ RequestController` (~one set per logical call)
- **0.40.x+**: `URL ≈ _HeadersList ≈ 2× Request` (one URL/Headers per leg of the redirect chain — both the original and the followed redirect are retained)

Both regimes retain after GC; the regression is that 0.40.x retains roughly **40% more** per call.

## Suspected cause (hypothesis only — not single-stepped)

Comparing the two `FetchInterceptor` patches:

- **0.36.10** does `pureFetch(request).then(...)` directly and clones the response only when a `response` listener is attached.
- **0.40.0** introduces a `RequestController` constructed with three closure callbacks (`passthrough`, `respondWith`, `errorWith`) and unconditionally calls `request.clone()` inside `passthrough` before dispatching to `pureFetch` — even when no `response` listener is registered. For 3xx with `redirect: 'follow'`, the wrapper's own `followFetchRedirect()` re-enters the wrapper on the redirect target, doubling per-transaction allocations.
- **0.41.8** keeps the controller/closure layout (with `FetchRequest` replacing the direct `Request` wrap) and the retention numbers are unchanged.

The candidate fix shape would be: skip the upfront `request.clone()` when no listener will consume it, and/or detach the controller's closure graph from `request`/`requestId` once `responsePromise` resolves so the FinalizationRegistry doesn't have to walk back through the closure to release undici state.

## Real-world impact

We see this on a Nuxt 4 SSR server (`@artmizu/nuxt-prometheus@3.0.1`, which pulls `@mswjs/interceptors@^0.40.0`). After `nuxt-prometheus` bumped from `2.8.0` (which pinned `^0.36.10`) to `3.0.1`, RSS started climbing on otherwise unchanged code. Production heap snapshots show the same retention shape: thousands of unique URL strings, undici `Pool`/`Client`/`TLSSocket`/`Gunzip` instances pinned per-request, with the MSW request-ID fingerprint on the wrapper. Setting `prometheus.disableRequestInterceptor: true` (which skips `interceptor.apply()`) restores flat RSS.

## What would help

- Confirmation of the regression and target version range to fix in.
- If there's a known way to use `FetchInterceptor` such that no per-call retention occurs when no listeners are attached (i.e. an "observe but don't mock" mode), I'm happy to switch our consumer to that — it would remove the need for the upstream fix from our side.

I can try to provide a PR.

PatrikBird:
> [!NOTE]
> This PR was drafted with the help of an AI assistant. The fix, tests, and numbers have been verified locally before opening.

Closes #783.

## Problem

`FetchInterceptor` (and any interceptor sharing `handleRequest`) leaks **native** memory under sustained traffic. JS heap snapshots stay clean, but RSS climbs hundreds of MB because undici-side state (sockets, zlib contexts, response buffers) is transitively pinned. The leak is amplified by `redirect: 'follow'` traffic and was the cause of unbounded RSS growth on a Nuxt 4 SSR server using `@artmizu/nuxt-prometheus`.

## Root cause

`handleRequest` attaches an `'abort'` listener to `request.signal` with `{ once: true }`, so the listener self-removes only when an abort fires. For requests that complete normally — the overwhelming majority — the listener stays attached forever.

Undici keeps internal references to a `Request` after `fetch` returns (keepalive socket metadata, etc.), keeping its `AbortSignal` reachable. The still-attached listener pins its closure (`requestAbortPromise` and the surrounding scope), which transitively pins per-request undici state.

The same listener pattern exists in `0.36.10`, which is why it also leaks (just less amplified — the closure captured a smaller scope before the `0.40.0` rewrite to `RequestController`-with-callbacks).

## Fix

Wrap the body of `handleRequest` in `try { ... } finally { ... }` and detach the abort listener in `finally`. Also `await controller.handled` before the trailing return so the `finally` block runs after the request is fully resolved (rather than synchronously, which would detach the listener while the fetch was still in flight).

## Numbers

Reproduction at 50,000 requests, `redirect: 'follow'`, concurrency 20, post-pumped-GC, on Node 24:

| Run                          | RSS Δ      | over baseline |
| ---------------------------- | ---------- | ------------- |
| baseline (no interceptor)    | 227 MB     | —             |
| 0.36.10                      | 408 MB     | +181 MB       |
| **0.41.8 unpatched**         | **684 MB** | **+457 MB**   |
| **0.41.8 patched (this PR)** | **239 MB** | **+12 MB**    |

\~97 % reduction; patched 0.41.8 is also cleaner than 0.36.10.

Repro: <https://github.com/PatrikBird/mswjs-interceptors-leak-repro>

## Tests

`test/modules/fetch/fetch-memory.test.ts` adds two tests:

1. **Behavioral, deterministic** — spies on `AbortSignal.prototype.{add,remove}EventListener` and verifies that the abort listener attached during a fetch is also detached before the fetch resolves.
2. **Worker-isolated heap snapshot** — spawns a worker that runs 5,000 fetches with no listeners, forces pumped GC, writes a heap snapshot, and asserts retained `Request` / `FetchRequest` instances < 100. Mirrors the pattern from the existing (separate) `MockHttpSocket` memory test.

All 76 fetch integration tests pass. Broader integration suite passes apart from `test/modules/http/compliance/http-errors.test.ts > forwards EHOSTUNREACH error for a bypassed request`, which was already failing on `main` before this PR (network-dependent, unrelated).

## Compatibility

No public API change. Behavior is identical except that the abort listener is detached when the request settles instead of remaining attached until GC. Aborts that arrive after the request settles were already no-ops because `requestAbortPromise` is no longer awaited at that point.

### DIFF
--- a/test/modules/fetch/fetch-memory.test.ts
+++ b/test/modules/fetch/fetch-memory.test.ts
@@ -1,0 +1,124 @@
+// @vitest-environment node
+import fs from 'node:fs'
+import path from 'node:path'
+import { Worker } from 'node:worker_threads'
+import { HttpServer } from '@open-draft/test-server/http'
+import { it, expect, beforeAll, afterAll, afterEach } from 'vitest'
+import { DeferredPromise } from '@open-draft/deferred-promise'
+import { FetchInterceptor } from '../../../src/interceptors/fetch'
+
+const server = new HttpServer((app) => {
+  app.get('/', (_req, res) => res.status(200).end())
+})
+
+const interceptor = new FetchInterceptor()
+
+beforeAll(async () => {
+  await server.listen()
+  interceptor.apply()
+})
+
+afterEach(() => {
+  interceptor.removeAllListeners()
+})
+
+afterAll(async () => {
+  interceptor.dispose()
+  await server.close()
+})
+
+it('detaches the abort listener from request.signal after the request resolves', async () => {
+  // Under sustained traffic, undici-side state can transitively keep the
+  // request's AbortSignal reachable after fetch returns. If the abort
+  // listener is left attached, its closure pins per-request scope and
+  // eventually compounds into hundreds of MB of native memory growth.
+  // See https://github.com/PatrikBird/mswjs-interceptors-leak-repro
+  // Capture the abort listener that handleRequest attaches and verify it is
+  // detached again before the request resolves. We can't assert "all abort
+  // listeners on the signal are removed" because undici attaches its own
+  // listeners that it manages independently.
+  const addedAbortListeners = new Set<EventListenerOrEventListenerObject>()
+  const removedAbortListeners = new Set<EventListenerOrEventListenerObject>()
+  const originalAdd = AbortSignal.prototype.addEventListener
+  const originalRemove = AbortSignal.prototype.removeEventListener
+  AbortSignal.prototype.addEventListener = function (type, listener, opts) {
+    if (type === 'abort' && listener) addedAbortListeners.add(listener)
+    return originalAdd.call(this, type, listener, opts)
+  }
+  AbortSignal.prototype.removeEventListener = function (type, listener, opts) {
+    if (type === 'abort' && listener) removedAbortListeners.add(listener)
+    return originalRemove.call(this, type, listener, opts)
+  }
+
+  try {
+    const response = await fetch(server.http.url('/'))
+    await response.text()
+
+    expect(addedAbortListeners.size).toBeGreaterThan(0)
+    // At least one of the abort listeners that was added during the request
+    // must also have been removed — the interceptor's own listener.
+    const detached = [...addedAbortListeners].filter((l) =>
+      removedAbortListeners.has(l)
+    )
+    expect(detached.length).toBeGreaterThan(0)
+  } finally {
+    AbortSignal.prototype.addEventListener = originalAdd
+    AbortSignal.prototype.removeEventListener = originalRemove
+  }
+})
+
+it(
+  'does not retain Request instances after passthrough without listeners',
+  async () => {
+    const snapshotPath = path.resolve(__dirname, 'fetch-memory.heapsnapshot')
+
+    const worker = new Worker(
+      path.resolve(__dirname, './fetch-memory-worker.js'),
+      {
+        workerData: {
+          requestCount: 5_000,
+          serverUrl: server.http.url('/'),
+          snapshotPath,
+        },
+        stderr: true,
+        stdout: true,
+      }
+    )
+
+    const completePromise = new DeferredPromise<void>()
+    worker.once('message', () => completePromise.resolve())
+    worker.on('error', (error) => completePromise.reject(error))
+
+    await completePromise
+
+    // Scan the heap snapshot for retained Request / FetchRequest instances.
+    const snap = JSON.parse(fs.readFileSync(snapshotPath, 'utf8'))
+    const types = snap.snapshot.meta.node_types[0]
+    const fields = snap.snapshot.meta.node_fields
+    const FC = fields.length
+    const TYPE_IDX = fields.indexOf('type')
+    const NAME_IDX = fields.indexOf('name')
+    const nodes = snap.nodes
+    const strings = snap.strings
+
+    let requestLikeCount = 0
+    for (let n = 0; n < nodes.length / FC; n++) {
+      const type = types[nodes[n * FC + TYPE_IDX]]
+      const name = strings[nodes[n * FC + NAME_IDX]] ?? ''
+      if (
+        type === 'object' &&
+        (name === 'Request' || name === 'FetchRequest')
+      ) {
+        requestLikeCount++
+      }
+    }
+
+    fs.rmSync(snapshotPath, { force: true })
+
+    // After 5,000 fetches with no listeners, only a small number of in-flight
+    // / FinalizationRegistry-pending Request instances should remain. The
+    // regression retains thousands.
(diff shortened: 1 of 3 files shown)


test/modules/fetch/fetch-memory.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    AbortSignal.prototype.removeEventListener = originalRemove
+  }
+})
+
+it(
+  'does not retain Request instances after passthrough without listeners',
+  async () => {
+    const snapshotPath = path.resolve(__dirname, 'fetch-memory.heapsnapshot')
+
+    const worker = new Worker(
```
kettanaito, line 75: Is the worker really required to reproduce this behavior? Or is it here just so we could measure the heap snapshot without Vitest affecting it?

PatrikBird: Thanks for your reply.
The worker isolates the heap snapshot from Vitest's own retained state, which would otherwise pollute the Request count and make the <100 threshold unreliable. The leak itself reproduces fine in the main process, so the worker is purely for clean measurement.

We could also swap this for a simpler WeakRef-based check in the main process (issue N fetches, hold WeakRefs, force GC, assert they all deref to undefined) if you'd prefer to avoid the worker + heap-snapshot overhead.

kettanaito:
Hi @PatrikBird. Thanks for opening this!

I tried porting your changes to #785 with changed tests and a smaller fix (no need for try/catch at all, that's what `until` does). Could you give it a review?

kettanaito:
## Released: v0.41.9 🎉

This has been released in v0.41.9.

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.41.9)
- 📦 [View on npm](https://www.npmjs.com/package/@mswjs/interceptors/v/0.41.9)

```
```

---

PR #794, opened 2026-05-11
Fix abort errors on large passthrough responses

MartinKolarik:
In passthrough mode, `MockHttpSocket` forwards data from the real socket into the mock socket that Node’s HTTP client observes. For large responses, the real socket can emit `close` before the mock socket has surfaced the pushed EOF as `end` to Node. Forwarding `close` immediately violates the ordering Node expects: the response readable side must finish before the socket is considered closed. Node then reports the otherwise valid response as `Error: aborted`

This PR tracks when the passthrough readable side has ended and defers forwarding close until the mock socket emits end. Error closes and closes before the upstream readable side ends are still forwarded immediately, so legitimate abort/error semantics are preserved.

### DIFF
--- a/src/interceptors/ClientRequest/MockHttpSocket.ts
+++ b/src/interceptors/ClientRequest/MockHttpSocket.ts
@@ -286,6 +286,17 @@ export class MockHttpSocket extends MockSocket {
       })
     }
 
+    let passthroughReadableEnded = false
+    let passthroughCloseEmitted = false
+    const emitPassthroughCloseOnce = (hadError: boolean) => {
+      if (passthroughCloseEmitted) {
+        return
+      }
+
+      passthroughCloseEmitted = true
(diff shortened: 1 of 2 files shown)


MartinKolarik:
Hey @kettanaito can you please clarify why you closed this? I'm happy to discuss the details if you think this wasn't the right fix, but the issue currently breaks many of our tests that rely on `nock` where this package is used.

kettanaito:
Hi @MartinKolarik. Thanks for opening this but I will close external pull requests until https://github.com/mswjs/interceptors/pull/770 is merged. It likely already fixed the issues you're tackling with your changes. It's unfeasible for me to maintain two dramatically different and growing versions of the architecture.

MartinKolarik:
I see, thanks. I'll subscribe there and recheck afterward.

PR #796, opened 2026-05-17
fix(HttpRequestInterceptor): prevent double `connection` listeners

mikicho:
## Why

The FetchInterceptor -> HttpRequestInterceptor chain resulted in _two_ `connection` event listeners being added to the underlying SocketInterceptor singleton. Because of that, handling the same request resulted in this error:

```
Invariant Violation: CANNOT HANDLE ALREADY HANDLED REQUEST
```


kettanaito:
Thanks for reporting this! I've pushed the fix into the net-v2 branch.

PR #797, opened 2026-05-27
refactor: make tests neutral (platform-agnostic)

mikicho:
- migrates all integration tests to neutral.
- drops `playwright` tests entirely in favor of vitest browser mode.
- update a bit of dependencies.
- update pnpm version on ci.


kettanaito:
Looking into this now...

PR #800, opened 2026-07-13
fix(net): prevent reusing unpatched sockets

mikicho:


kettanaito:
This is tricky to handle because despite my initial assumptions, `connection` on the socket interceptor _will not_ be emitted since the socket that does the second request was never patched, in the first place. My initial proposal to use a symbol to detect such scenarios also fails since it relied on the `connection` event emitting anyway and then us checking the symbol.

I have two thoughts:
- Consider the concept of "pausing" the interception. Only a partial fix since it still matters when you _enable_ it.
- Consider patching `Agent` so it could do the checks without doing anything else. Somewhat problematic due to default/custom agents and all of that hassle I'd rather not manage.

PR #801, opened 2026-07-15
fix(ClientRequest): improve IPv6 handling in connection options and tests

mikicho:
Fixes incorrect IPv6 detection/address resolution when a request is made with the `family: 6` option (rather than an explicit IPv6 host).

**Changes:**
- **`address-info.ts`**: simplified `getAddressInfoByConnectionOptions` to derive the mocked socket address purely from `isIPv6` (`family === 6 || net.isIPv6(host)`), removing the flawed fallback that used any parsable IP host as the address.
- **`connection-options-to-url.ts`**: reverted URL construction to rely only on `net.isIPv6(host)` (not `family`), since `family` should affect the socket's address but not the constructed request URL.

### DIFF
--- a/src/interceptors/net/utils/address-info.ts
+++ b/src/interceptors/net/utils/address-info.ts
@@ -9,14 +9,9 @@ export function getAddressInfoByConnectionOptions(
   }
 
   const isIPv6 = options.family === 6 || net.isIPv6(options.host || '')
-  const ipAddress = options.host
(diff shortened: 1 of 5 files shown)


kettanaito:
@mikicho can you please provide a bit of context in the description as to why this was needed/what it improves? Thanks.

mikicho:
@kettanaito Updated description

PR #812, opened 2026-07-23
fix(socket): move last resort passthrough to the socket level

### LINKED ISSUE #808: Raw non-HTTP `net` socket never connects while `HttpRequestInterceptor` is active
When only `HttpRequestInterceptor` is applied, a plain `net.connect()` socket carrying non-HTTP traffic to a real server deadlocks: it never passes through,
  never receives data, and never errors. It stays pending forever.

  ### Reproduction

  ```ts
  import net from 'node:net'
  import { it, expect } from 'vitest'
  import { HttpRequestInterceptor } from '@mswjs/interceptors/http'

  const interceptor = new HttpRequestInterceptor()
  interceptor.apply()

  it('passes a non-http socket through', async () => {
    const server = net.createServer((connection) => {
      connection.on('data', () => {
        connection.write('PONG')
      })
    })

    await new Promise<void>((resolve) => {
      server.listen(0, () => resolve())
    })
    const address = server.address() as net.AddressInfo

    const response = await new Promise<string>((resolve, reject) => {
      const socket = net.connect(address.port, '127.0.0.1')
      socket.on('connect', () => {
        socket.write('PING')
      })
      socket.on('data', (chunk) => {
        resolve(chunk.toString())
        socket.destroy()
      })
      socket.on('error', reject)
      setTimeout(() => reject(new Error('Timed out: never passed through')), 3000)
    })

    expect(response).toBe('PONG')
  })
  ```

  **Expected:** the socket is not an HTTP connection, so it passes through to the real server and receives `PONG`.

  **Actual:** the client never receives anything. The test times out.

  ### Root cause

  The socket-layer interceptors each own their "unhandled connection" fallback, and the fallbacks don't compose:

  1. `SocketInterceptor` (`src/interceptors/net/index.ts`) calls `controller.passthrough()` only when there are **zero** `connection` listeners on its emitter.
  2. `HttpRequestInterceptor` (`src/interceptors/http/source.ts`) subscribes a `connection` listener to sniff every socket, which suppresses that fallback.
  3. When the sniffer sees a non-HTTP first packet, it sets `isHttpConnection = false` and returns — it never claims the socket, but it never passes it through
  either.

  Nobody is responsible for the socket after every subscriber has declined it, so the `SocketController` stays in `PENDING` indefinitely.

  The same structural gap affects any combination of socket-layer subscribers (WebSocket and SMTP ride the same layer): each interceptor's presence suppresses the
  default passthrough for all traffic, not just the traffic it handles.

  ### Proposed direction

  Move the last-resort passthrough into the `SocketController` itself, fired only after every subscriber has had its chance to claim the socket, and remove the
  per-interceptor fallbacks. That makes "unclaimed after everyone declined" a single owned state instead of something each interceptor has to reimplement.

kettanaito:
Fixes #808

### DIFF
--- a/src/interceptors/http/source.ts
+++ b/src/interceptors/http/source.ts
@@ -164,6 +164,13 @@ export class NodeHttpRequestSource extends Interceptor<HttpRequestEventMap> {
 
               socket.on('drain', () => targetSocket.resume())
               socket.on('data', (data) => targetSocket.write(data))
+              /**
+               * @note "finish" on the client socket is the client
+               * half-closing the tunnel. Forward it to the target as
+               * the FIN it stands for ("end" cannot signal it: that is
+               * the readable-side EOF, echoing the target's own end).
+               */
+              socket.on('finish', () => targetSocket.end())
               socket.on('close', () => targetSocket.destroy())
 
               targetSocket.write(toBuffer(chunk))
--- a/src/interceptors/http/source.ts
+++ b/src/interceptors/http/source.ts
@@ -21,7 +21,6 @@ import { handleRequest, HandleRequestOptions } from '../../utils/handle-request'
 import { isResponseError, kErrorResponse } from '../../utils/response-utils'
 import { createLogger } from '../../utils/logger'
 import {
-  kPatched,
   kRawSocket,
   SocketController,
   type FlushPendingDataFunction,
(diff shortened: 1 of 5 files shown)


kettanaito:
## Released: v0.42.2 🎉

This has been released in v0.42.2.

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.42.2)
- 📦 [View on npm](https://www.npmjs.com/package/@mswjs/interceptors/v/0.42.2)

```
```

---

PR #814, opened 2026-07-24
fix(socket): properly handle first bytes when switching socket types

### LINKED ISSUE #813: Kept-alive socket leaks the next request's bytes to the real server before the interceptor's verdict
When a kept-alive connection's previous exchange was passed through, the bytes of the **next** request on that socket are flushed to the real server before the interceptor decides how to handle that request. Mocking the second exchange does not prevent the real server from receiving it.

### Reproduction

```ts
// 1. Exchange A on a kept-alive connection: unhandled → passthrough.
//    The controller transitions to PASSTHROUGH; the real handle is swapped in.
// 2. Exchange B on the same socket: the interceptor mocks it.
const received: Array<string> = []
const server = http.createServer((request, response) => {
  received.push(request.url!)
  response.end('real')
})

interceptor.on('request', ({ request, controller }) => {
  if (new URL(request.url).pathname === '/b') {
    controller.respondWith(new Response('mock'))
  }
})

// Reuse one connection (e.g. Undici, or a raw socket writing
// two HTTP/1.1 messages back to back).
await client.get('/a') // passthrough
await client.get('/b') // mocked

expect(received).toEqual(['/a']) // ❌ actual: ['/a', '/b']
```

**Expected:** the real server never observes `/b` — it was mocked.

**Actual:** the real server receives `/b`'s request line and headers (and any body written before the verdict). Its response to those stray bytes is forwarded to the client and can interleave with the mocked response.

### Root cause

`TcpSocketController#handleWrite` dispatches on `readyState`, which is still `PASSTHROUGH` from the previous exchange when the next request's first bytes are written. The controller only returns to `PENDING` at the HTTP message boundary — either on the Agent `free` event or inside `onRequest` *after* the new request's headers are parsed. By then the write has already reached the real socket (`#realWriteGeneric` post-swap, or a direct passthrough write pre-swap).

Clients that emit `free` (Node's `http.Agent`) are mostly unaffected: `free` resets the controller before the next write. Clients that reuse sockets silently (Undici, raw `net` sockets carrying multiple messages) hit the lazy `onRequest`-time reset, which is one write too late.

A concrete downstream casualty: a mocked `CONNECT` following a passthrough exchange — the `CONNECT` line leaks to the origin server, which typically answers `400`, corrupting the freshly mocked tunnel.

### Proposed direction

Reset the controller to `PENDING` at the *end* of the previous exchange rather than at the start of the next one — the request parser knows the message boundary (`onMessageComplete`), so post-boundary writes buffer in the pending state until the next verdict, the same way a connection's very first write does. The open question is passthrough exchanges whose boundary is defined by the *response* completing on the real socket; that may require tracking the response boundary (the machinery `HttpResponseParser` already provides when `response` listeners exist).

kettanaito:
Fixes #813

### DIFF
--- a/src/interceptors/http/http-parser.ts
+++ b/src/interceptors/http/http-parser.ts
@@ -9,14 +9,17 @@ interface HttpRequestParserOptions {
     url: URL
   }
   onRequest: (request: Request, abortController: AbortController) => void
+  onMessageComplete?: () => void
 }
 
 export class HttpRequestParser extends HttpParser<1> {
   #requestBodyStream?: Readable
+  #upgrade = false
 
   constructor(options: HttpRequestParserOptions) {
     super(1, {
-      onHeadersComplete: ({ rawHeaders, method, url: path }) => {
+      onHeadersComplete: ({ rawHeaders, method, url: path, upgrade }) => {
+        this.#upgrade = upgrade
         /**
          * @note When the socket is reused, "connectionOptions" will point
          * to the "net.connect()" call options that established the connection,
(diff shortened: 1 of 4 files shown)


kettanaito:
## Released: v0.42.3 🎉

This has been released in v0.42.3.

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.42.3)
- 📦 [View on npm](https://www.npmjs.com/package/@mswjs/interceptors/v/0.42.3)

```
```

---

PR #816, opened 2026-07-25
fix(http): passthrough non-http sockets writing on server data

mikicho:


mikicho:
@kettanaito Failing test for now. with `SocketInterceptor` it works, but with `HttpRequestInterceptor` it doesn't.
I'm not sure how we can fix this. We need to wait for the client's request body to determine whether the request should be mocked or passed through, while the client waits for the server's greeting.

mikicho:
@kettanaito I think I've found the culprit.

`NodeHttpRequestSource` [subscribes to the `connection` event](https://github.com/mswjs/interceptors/blob/2a087221c461602486758b1ca14ab1995c581e66/src/interceptors/http/source.ts#L64), which prevents the underlying `SocketInterceptor` from [transitioning to the passthrough state](https://github.com/mswjs/interceptors/blob/d2020820e491fcada48327263a64a8487acead69/src/interceptors/net/index.ts#L256). At the same time, it doesn't `decline` the connection because [it waits for the client to start writing the request body before determining whether it's an HTTP request](https://github.com/mswjs/interceptors/blob/2a087221c461602486758b1ca14ab1995c581e66/src/interceptors/http/source.ts#L142).

kettanaito:
Thanks for writing the test, @mikicho. 

This seems to be hitting the infamous first-write issue. We can fix your use case but it will break this:

```ts
interceptor.on('request', ({ controller }) => {
  controller.respondWith(new Response('mocked'))
})

// No fetch/http.request context identifies this as HTTP.
const socket = net.connect(80, 'example.com')

socket.on('connect', () => {
  socket.write(
    'GET / HTTP/1.1\r\nHost: example.com\r\nConnection: close\r\n\r\n'
  )
})
```

If writes are deferred to `connect`, we can never know what gets written to the socket. This contradicts the mock-first approach in Interceptors. It's a catch 22.

What do you think we should do?

kettanaito:
The scenario is: socket writes once it receives data from the server. Nothing to do with connect (we already emit that in the interceptor). 

As such, this isn't a behavior we should/can support but a blocker we should unblock. This kind of socket scenarios are outside of Interceptors. 

Here's raw thoughts:

- In HttpRequestInterceptor
- after we emitted `connect`
- and <UNCLEAR_CONDITION> is true
- and the socket hasn't written anything
- and the socket isn't destroyed
- passthrough automatically, this is not HTTP

To think of it, `UNCLEAR_CONDITION` is probably nextTick + reasonable timeout. Here's why.

In regular HTTP, clients write the message before connect. Some clients write after connect, and we support that, too, so the condition above will be false.

It's only non-HTTP clients that will prove the above condition true. 

_Update:_ Use `setImmediate`. It runs after the event loop clears. This will reliably support delays + nested nextTick/queueMicrotask from the userland between connect and write. 

```ts
// socket-interceptor
socket.once('connect', () => {
  setImmediate(() => {
    if (!socket.destroyed && !socket.bytesWritten) {
      controller.passthrough()
      // and prevent any other actions on the socket in the root closure.
    }
  })
})
```

kettanaito:
## Update

The originally reported issue is now fixed. `socket.bytesWritten` was replaced by `isHttpConnection` as that value already exists and gets set on the first written bytes, which substitutes the `bytesWritten` check. 

This still won't be bulletproof if some clients decide to schedule writes on `setImmediate`, which I believe Undici does in some scenarios.

kettanaito:
## Released: v0.44.0 🎉

This has been released in v0.44.0.

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.44.0)
- 📦 [View on npm](https://www.npmjs.com/package/@mswjs/interceptors/v/0.44.0)

```
```

---

PR #824, opened 2026-09-08
fix(http): proper HTTP/2 negotiation for mocked and passthrough

### LINKED ISSUE #819: ClientRequestInterceptor passthrough breaks fetch() to any HTTP/2 (ALPN h2) host
### Description

`ClientRequestInterceptor`'s passthrough path breaks any `fetch()` call to a server that negotiates HTTP/2 via ALPN — which is virtually every modern HTTPS server.

Reproduces on `0.42.0` through `0.42.3` (the current latest). Does not reproduce on `0.41.9`, the previous architecture. Looks like it was introduced by the TCP/TLS wrap-based interception rewrite in `0.42.0` (#770).

### Reproduction

https://github.com/melkir/mswjs-interceptors-h2-repro

```js

interceptor.on('request', () => {
  // Unhandled -> passthrough to the real network.
})
interceptor.apply()

const res = await fetch('https://api.github.com/zen')
```

Fails:

```
TypeError: fetch failed
  cause: SocketError: other side closed
```

(Sometimes instead: `HTTPParserError: Response does not match the HTTP/1.1 protocol`, with a raw HTTP/2 SETTINGS frame in the response bytes — both look like symptoms of the same underlying mismatch, not two different bugs.)

Comment out `interceptor.apply()`, or point the request at a plain `http://` host instead, and it succeeds.

### Root cause

The passthrough connection is a real socket the interceptor opens itself (`createConnection()`), using TLS options — including the ALPN protocol list — copied from the connection the caller was originally about to make. If the remote server negotiates `h2` from that list, the interceptor still writes the request as plain HTTP/1.1 text and parses the response with its own HTTP/1.1 parser. The remote server, receiving non-HTTP/2-framed bytes over an HTTP/2 connection, closes it — or the client receives real HTTP/2 frames back that its HTTP/1.1 parser can't read.

Confirmed by patching the library locally to log the negotiated ALPN protocol and the raw bytes written to the passthrough socket: negotiated protocol was `h2`, bytes written were `GET /zen HTTP/1.1\r\n...`.

### Where this was found

Via Next.js's `experimental.testProxy`, which vendors this library: https://github.com/vercel/next.js/issues/96521. Any server-side `fetch()`/request made during rendering, outside a recognized test context, to an HTTP/2-capable host fails. But the bug reproduces standalone with no Next.js involved, as above — it's entirely within this library's passthrough handling.

### Environment

```
@mswjs/interceptors: 0.42.0 – 0.42.3
node: v26.7.0
```

kettanaito:
- Fixes #819

### DIFF
--- a/src/interceptors/net/index.ts
+++ b/src/interceptors/net/index.ts
@@ -51,10 +51,7 @@ declare module 'node:http' {
      * @note An undocumented method backing every agent-driven request
      * (see "#stopReusingUnpatchedSockets").
      */
-    addRequest?: (
-      request: http.ClientRequest,
-      ...args: Array<unknown>
-    ) => void
+    addRequest?: (request: http.ClientRequest, ...args: Array<unknown>) => void
   }
 }
 
@@ -223,7 +220,7 @@ export class SocketInterceptor extends Interceptor<SocketEventMap> {
                 () => {
                   /**
                    * @note Create the passthrough connection via the original
-                   * "tls.connect()" with the original connection options
+                   * "tls.connect()" with the effective connection options
                    * (the real DNS lookup and the caller's certificate
                    * validation included). The latch exempts the transport
                    * connect of that connection from interception.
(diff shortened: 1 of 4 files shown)


kettanaito:
## Released: v0.42.4 🎉

```
```

---

PR #825, opened 2026-09-08
fix(http): stop observation on parser errors instead of crashing

### LINKED ISSUE #817: Uncaught parser error on an observed connection crashes the Node.js process ("Data after `Connection: close`")
## Environment

- `@mswjs/interceptors`: 0.42.3
- Node.js: 24.18.0 (also reproduced on 24.14.0)
- Interceptor: `HttpRequestInterceptor` (via `BatchInterceptor`), passive observation (a `request` and a `response` listener, no mocking)

## What happens

When observing a passthrough request, the http interceptor feeds every chunk of the socket through its llhttp parser. If the server sends stray bytes after a response that declared `Connection: close`, llhttp reports `HPE_CLOSED_CONNECTION` and the parser wrapper throws `Error: Data after \`Connection: close\`` inside the socket's `data` listener:

https://github.com/mswjs/interceptors/blob/75e259c426e080fe6847862af7ae49cdc17e2804/src/interceptors/http/source.ts#L358

An uncaught exception inside a socket event handler terminates the Node.js process.

The real HTTP client parses the same bytes with its own error handling and recovers gracefully; only the interceptor's observer copy of the parser is fatal. The two request-parser feed sites (lines 133 and 415 of the same file) have the same exposure.

## Real-world impact

We hit this on AWS Lambda (Node 24): a kept-alive HTTPS connection to an upstream service delivered trailing bytes after a `Connection: close` response, and the entire runtime died mid-request with `Runtime.ExitError`:

```
Error: Data after `Connection: close`
    at <parser>.execute (bundled source.ts:358)
    at TLSSocket.<listener> ...
Node.js v24.18.0
```

The interceptor is used purely for observability (request/response logging), so a protocol hiccup on one observed connection took down the whole application.

## Reproduction

A raw TCP server that writes a complete `Connection: close` response plus trailing junk in a single frame, fetched through the interceptor with a `response` listener attached:

```ts
import net from 'node:net'
import { fetch } from 'undici'

  name: 'repro',
  interceptors: [new ClientRequestInterceptor(), new FetchInterceptor()],
})
interceptor.apply()
interceptor.on('response', () => {})

const server = new net.Server((connection) => {
  connection.on('data', () => {
    connection.write(
      'HTTP/1.1 200 OK\r\n' +
        'Content-Length: 2\r\n' +
        'Connection: close\r\n' +
        '\r\n' +
        'ok' +
        'stray bytes after close'
    )
    connection.end()
  })
})

server.listen(0, '127.0.0.1', async () => {
  const { port } = server.address() as net.AddressInfo
  const response = await fetch(`http://127.0.0.1:${port}/`)
  console.log(await response.text()) // never reached — process crashes
})
```

```
Error: Data after `Connection: close`
    at Parser.execute (src/interceptors/http/http-parser/index.ts)
    at Socket.<anonymous> (src/interceptors/http/source.ts:358)
Node.js v24.18.0
```

## Expected behavior

Since interception here is observational, a parser error on an observed connection should end the observation of that connection — the consumer owns the socket and surfaces its own protocol errors — and must never crash the host process.

I have a fix with a regression test ready; PR follows shortly.

kettanaito:
Fixes #817.

### DIFF
--- a/src/interceptors/http/http-parser.ts
+++ b/src/interceptors/http/http-parser.ts
@@ -4,6 +4,7 @@ import { FetchRequest, FetchResponse } from '../../utils/fetch-utils'
 import { HttpParser } from './http-parser/index'
 
 interface HttpRequestParserOptions {
+  onError: (error: Error) => void
   connectionOptions: {
     method?: string
     url: URL
@@ -18,6 +19,7 @@ export class HttpRequestParser extends HttpParser<1> {
 
   constructor(options: HttpRequestParserOptions) {
     super(1, {
+      onError: options.onError,
       onHeadersComplete: ({ rawHeaders, method, url: path, upgrade }) => {
         this.#upgrade = upgrade
         /**
@@ -82,6 +84,7 @@ export class HttpRequestParser extends HttpParser<1> {
       },
       onMessageComplete: () => {
         this.#requestBodyStream?.push(null)
+        this.#requestBodyStream = undefined
 
         /**
          * @note An upgraded exchange (e.g. "CONNECT", WebSocket) has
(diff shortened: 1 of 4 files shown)


kettanaito:
## Released: v0.42.4 🎉

```
```

---

PR #826, opened 2026-09-08
fix(http): handle server-closed keep-alive sockets and detach completed parsers

### LINKED ISSUE #822: Observed kept-alive connection closed by the server leaves a dead client socket in the http.Agent pool; the next request hangs forever
## Summary

When an observed (passed-through, not mocked) HTTP connection is kept alive by `http.Agent` and the **server** closes it while it sits idle in the agent's free pool, the interceptor forwards the real socket's `close` as a bare event without destroying the client socket. The client socket stays `writable`, so `http.Agent` keeps it in `freeSockets` (`removeSocket` only evicts unwritable sockets, `addRequest` only skips destroyed ones) and the next request written to it **hangs forever**. Without the interceptor Node evicts the socket and the request succeeds on a new connection.

Two related problems on the same code path:

1. A `close` that arrives while the socket reads are corked (waiting for the `"response"` listeners of a request whose response never comes, e.g. the server resets the reused connection) is buffered with the corked reads and never replayed. The request hangs instead of failing with `socket hang up` / `ECONNRESET`.
2. The response parser of a passed-through exchange is never detached from the kept-alive socket, so it keeps parsing the responses of the following exchanges on that connection and emits them as `"response"` events of the original request (duplicate events, with the wrong `requestId`).

The same hang follows unsolicited bytes on an idle pooled socket: since Node 24 (`installFreeSocketDataGuard`) the agent puts a read guard on the socket handle, which destroys the *real* socket and takes us into the first scenario.

## Environment

- `@mswjs/interceptors` 0.42.3 (`ClientRequestInterceptor` / `BatchInterceptor` with `FetchInterceptor`), passive observation only (`request` and `response` listeners, no `respondWith`)
- Node.js 24.20.0 (also seen on 24.18.0), macOS and AWS Lambda (Linux arm64)
- Client: `node:http`/`node:https` with `new Agent({ keepAlive: true })` — in our case the AWS SDK v3 (`@smithy/node-http-handler`) talking to DynamoDB, whose idle connections get closed server-side

## Reproduction

```js
import http from 'node:http'

const scenario = process.argv[2] // 'idle-close' | 'reset-on-reuse' | 'stray-bytes'

  name: 'observer',
  interceptors: [new ClientRequestInterceptor(), new FetchInterceptor()],
})
interceptor.apply()
interceptor.on('response', () => {})

  res.end('ok')
  if (req.url !== '/first') return
  if (scenario === 'idle-close') setTimeout(() => req.socket.destroy(), 100)
  if (scenario === 'reset-on-reuse') req.socket.prependListener('data', () => req.socket.destroy())
  if (scenario === 'stray-bytes') setTimeout(() => req.socket.write('\r\n'), 20)
})
await new Promise((r) => server.listen(0, r))
const { port } = server.address()
const agent = new http.Agent({ keepAlive: true })

const get = (path) =>
  new Promise((resolve, reject) => {
    const req = http.get({ host: '127.0.0.1', port, path, agent }, (res) => {
      res.resume().on('end', () => resolve({ status: res.statusCode, reusedSocket: req.reusedSocket }))
    })
    req.on('error', reject)
    setTimeout(() => reject(new Error('second request HUNG')), 4000).unref()
  })

console.log(await get('/first'))
await new Promise((r) => setTimeout(r, 400))
console.log(await get('/second').catch((e) => `error: ${e.code ?? e.message}`))
agent.destroy(); server.close()
```

| scenario | without interceptor | with interceptor 0.42.3 |
|---|---|---|
| `idle-close` | `{ status: 200, reusedSocket: false }` | **second request hangs** |
| `reset-on-reuse` | `error: ECONNRESET` | **second request hangs** |
| `stray-bytes` | `{ status: 200, reusedSocket: false }` | **second request hangs** |

Same results over `https`.

## Root cause

- `TcpSocketController#onRealSocketClose` → `#emitClientClose` ends with `this.socket.emit('close', hadError)` while the client socket is still alive. Node always destroys a socket before emitting `close` (`destroyed === true`, `writable === false` inside the listeners); `http.Agent` relies on that to drop the socket from its free pool.
- When reads are corked, `#onRealSocketClose` pushes the close into `#corkedReads`; the only `uncorkReads()` callers are the `"response"` path and the response parser error path, so a close with no response never uncorks.
- In `NodeHttpRequestSource`, the `passthrough()` branch registers `realSocket.on('data', onResponseData)` per exchange and never removes it after the response message completes.

## Fix

Branch `fix/keep-alive-dead-socket-eviction` on my fork, commit https://github.com/mdesousa/interceptors/commit/592a39f99a11eb2e9bfebb01d1b8f18ffa709397 (stacked on the parser-crash fix from #817):

- `socket-controller.ts`: destroy the client socket when the real connection closes; emit `close` manually only where the swapped handle suppresses it.
- `source.ts`: track whether a response was parsed; on real socket `close` before any response, `uncorkReads()` so the buffered end-of-stream and close reach the client (`socket hang up`), matching Node.
- `source.ts` / `http-parser.ts`: detach and free the response parser once the final response message completes (kept across 1xx informational responses) or the connection closes.
- Regression test `test/modules/http/regressions/http-keep-alive-server-closed-socket.test.ts` covering the three scenarios. The full suite passes (923 tests).

I cannot open a pull request against this repository (PR creation is limited to collaborators), so the branch above is the proposed change — happy to adjust anything.

## Impact

For observability-only users this is severe: every request that reuses a pooled connection the server has dropped stalls until the client's own timeout. On AWS Lambda that meant DynamoDB calls hanging for the full 300 s function timeout with no error, several times a day.

kettanaito:
Fixes #822

### DIFF
--- a/src/interceptors/http/http-parser.ts
+++ b/src/interceptors/http/http-parser.ts
@@ -107,10 +107,12 @@ export class HttpRequestParser extends HttpParser<1> {
 
 export class HttpResponseParser extends HttpParser<2> {
   #responseBodyStream?: Readable | null
+  #status = 0
 
   constructor(options: {
     onResponse: (response: Response) => void
     onError: (error: Error) => void
+    onMessageComplete?: (status: number) => void
   }) {
     super(2, {
       onError: options.onError,
@@ -119,6 +121,7 @@ export class HttpResponseParser extends HttpParser<2> {
         statusCode: status,
         statusMessage: statusText,
       }) => {
+        this.#status = status
         const headers = FetchResponse.parseRawHeaders([...rawHeaders])
 
         const response = new FetchResponse(
@@ -147,6 +150,7 @@ export class HttpResponseParser extends HttpParser<2> {
       onMessageComplete: () => {
         this.#responseBodyStream?.push(null)
         this.#responseBodyStream = null
+        options.onMessageComplete?.(this.#status)
       },
     })
   }
--- a/src/interceptors/http/source.ts
+++ b/src/interceptors/http/source.ts
@@ -330,15 +330,20 @@ export class NodeHttpRequestSource extends Interceptor<HttpRequestEventMap> {
                        */
                       socketController.corkReads()
 
-                      let responseParserFailed = false
+                      let responseParserDisposed = false
+                      let responseComplete = false
+                      let hasFinalResponse = false
                       const responseParser = new HttpResponseParser({
                         onError: (error) => {
-                          responseParserFailed = true
-                          realSocket.removeListener('data', onResponseData)
-                          responseParser.free(error)
+                          disposeResponseParser(error)
                           socketController.uncorkReads()
                         },
+                        onMessageComplete: (status) => {
+                          responseComplete = status >= 200 || status === 101
+                        },
                         onResponse: async (response) => {
+                          hasFinalResponse =
+                            response.status >= 200 || response.status === 101
                           httpLogger.verbose(
                             'HTTP response parser parsed: %d %s',
                             response.status,
@@ -378,7 +383,7 @@ export class NodeHttpRequestSource extends Interceptor<HttpRequestEventMap> {
                              * final response on the "response" event listeners.
                              */
                             if (
-                              !responseParserFailed &&
+                              !responseParserDisposed &&
                               response.status < 200 &&
                               response.status !== 101
                             ) {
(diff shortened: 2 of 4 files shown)


kettanaito:
## Released: v0.42.4 🎉

```
```

---

PR #833, opened 2026-09-10
fix(fetch): prevent hanging on passthrough response stream cancellation

### LINKED ISSUE #712: Unconsumed cloned responses cause hangs in certain cases
This was a crazy rabbit-hole.

## TL;DR:

The responses cloned here:

https://github.com/mswjs/interceptors/blob/77141e2f51b85830dca632d5026177639627987c/src/interceptors/fetch/index.ts#L161-L169

Create issues in the event that the `body` of the original `response` is cancelled with `response.body.cancel()`, if those clones are also not consumed. In such cases, the `Promise` returned by `response.body.cancel()` never settles (neither resolves nor rejects).

## Further details

As best I can tell, this is due to the underlying `.tee()` call that happens on the `ReadableStream` that is the Response's body.

Particularly this bit from [the docs](https://developer.mozilla.org/en-US/docs/Web/API/Response/clone) made me wonder if the `.clone()` was the cause of [our memory leak issues]() (doesn't look like it)

> If only one cloned branch is consumed, then the entire body will be buffered in memory.

The behavior here though is really odd, and seems to indicate an actual bug in the implementation of `ReadableStream` (I _think_) in `undici`. What happens is that the `Promise` returned by `response.body.cancel()` _never resolves or rejects_ if that response has previously been cloned, and the body of the clone has not yet been consumed.

> [!NOTE]
> This behavior does not occur when tested in the browser. The `.cancel()` call resolves fairly quickly, and I assume the cloned response becomes disconnected from its source and eventually garbage collection.

A simple reproduction can be seen here:

```js
import { setupServer } from 'msw/node';

if (process.env.ENABLE_MSW === 'true') {
  const server = setupServer();
  server.listen({
    onUnhandledRequest: 'bypass',
  });
}

async function run() {
  const response = await fetch('https://dummyjson.com/posts/1');
  await response.body.cancel();
  console.log('response body canceled');
}

run()
  .then(() => console.log('done'))
  .catch((error) => console.error('fail', error));
```

This program will produce no output at all with `ENABLE_MSW=true`, as the `await response.body.cancel();` never settles. Incidentally it seems as though the underlying async operation is unref-ed as well, since it doesn't keep the event loop from terminating.

In our application, we got around this by patching the following in right after the `emitAsync` call:

```diff
const responseClone = response.clone()

await emitAsync(this.emitter, 'response', {
  response: responseClone,
  isMockedResponse: false,
  request: requestCloneForResponseEvent,
  requestId,
})
+ 
+ if (!responseClone.bodyUsed) {
+   await responseClone.bytes(); // Consume the body if it hasn't been used
+ }
```

This is a fairly brute-force approach and I'm not sure it would hold up in every scenario. But I thought I'd put it on our radar since in our case it caused server requests to never complete, creating `504 Gateway` timeout errors in an application that simply called `setupServer()`, without registering any handlers.

### LINKED ISSUE #799: Passthrough response body deadlocks after `reader.cancel()` + retry
## Description

When a passthrough response body is cancelled (via `reader.cancel()`) and a subsequent request is made through the same interceptor, reading the new response body (via `.text()`) deadlocks and it never resolves.

This reared it head for us while using the OpenAI SDK, which cancels the error response body stream before retrying, triggering the deadlock on the retry's response.

## Reproduction

This is as minimal a reproduction I could figure out, with the OpenAI retry logic inlined in the test itself.

```ts
import { createServer, type Server } from 'node:http';
import type { AddressInfo } from 'node:net';
import { http, passthrough } from 'msw';
import { type SetupServer, setupServer } from 'msw/node';
import { afterAll, beforeAll, describe, expect, it } from 'vitest';

const TEST_TIMEOUT_MS = 3_000;

let httpServer: Server;
let mswServer: SetupServer;
let url: string;

beforeAll(async () => {
	// Local HTTP server that always returns 429.
	httpServer = createServer((_req, res) => {
		res.writeHead(429, { 'content-type': 'application/json' });
		res.end(JSON.stringify({
			error: {
				message: 'Rate limit exceeded',
				type: 'requests',
				param: null,
				code: 'rate_limit_exceeded',
			},
		}));
	});

	await new Promise<void>((resolve) => httpServer.listen(0, '127.0.0.1', resolve));
	const { port } = httpServer.address() as AddressInfo;
	url = `http://127.0.0.1:${port}/v1/chat/completions`;

	// MSW passthrough handler — the interceptor will clone the response (tee the body).
	mswServer = setupServer(http.post(url, () => passthrough()));
	mswServer.listen();
});

afterAll(async () => {
	mswServer.close();
	await new Promise<void>((resolve) => httpServer.close(() => resolve()));
});

/** Cancel a ReadableStream the same way the OpenAI SDK does before retrying. */
async function cancelReadableStream(body: ReadableStream | null) {
	if (!body) {
		return;
	}

	const reader = body.getReader();
	await reader.cancel();
	reader.releaseLock();
}

async function fetchWithRetry(
	customFetch: (input: string, init: RequestInit) => Promise<Response>,
	maxRetries: number,
): Promise<Response> {
	for (let attempt = 0; attempt <= maxRetries; attempt++) {
		const response = await customFetch(url, {
			method: 'POST',
			headers: { 'content-type': 'application/json' },
			body: JSON.stringify({ model: 'test', messages: [{ role: 'user', content: 'hi' }] }),
		});

		if (response.ok) {
			return response;
		}

		if (attempt < maxRetries) {
			// Retryable: cancel the body stream and retry.
			await cancelReadableStream(response.body);
			continue;
		}

		// Final attempt: read the error body.
		const errText = await response.text();
		throw new Error(`${response.status}: ${errText.slice(0, 200)}`);
	}

	throw new Error('unreachable');
}

describe('MSW passthrough body stream deadlock', () => {
	it('BUG: .text() deadlocks on retry after body cancellation', {
		timeout: TEST_TIMEOUT_MS,
	}, async () => {
		const error = await fetchWithRetry(
			(input, init) => globalThis.fetch(input, init), 2,
		).catch((e) => e);

		// When the bug is present, this is never reached — the test
		// times out because .text() on the final attempt deadlocks.
		expect(error).toBeInstanceOf(Error);
		expect(error.message).toContain('429');
	});

	it('FIX: identity TransformStream detaches from the tee', {
		timeout: TEST_TIMEOUT_MS,
	}, async () => {
		const error = await fetchWithRetry(async (input, init) => {
			const res = await globalThis.fetch(input, init);
			if (!res.ok && res.body) {
				return new Response(res.body.pipeThrough(new TransformStream()), {
					status: res.status,
					statusText: res.statusText,
					headers: res.headers,
				});
			}
			return res;
		}, 2).catch((e) => e);

		expect(error).toBeInstanceOf(Error);
		expect(error.message).toContain('429');
	});

	it('FIX: eager .text() consumption and re-wrap', {
		timeout: TEST_TIMEOUT_MS,
	}, async () => {
		const error = await fetchWithRetry(async (input, init) => {
			const res = await globalThis.fetch(input, init);
			if (!res.ok) {
				const text = await res.text();
				return new Response(text, {
					status: res.status,
					statusText: res.statusText,
					headers: res.headers,
				});
			}
			return res;
		}, 2).catch((e) => e);

		expect(error).toBeInstanceOf(Error);
		expect(error.message).toContain('429');
	});
});
```

`package.json`:

```json
{
	"dependencies": {
		"msw": "^2.14.6",
		"vitest": "^4.1.8"
	},
	"devDependencies": {
		"@types/node": "^26.0.0"
	}
}
```

### Expected behaviour

All three tests pass. The `fetchWithRetry` function should be able to cancel a response body, retry, and read the new response body.

### Actual behaviour

The first test (BUG) hangs - `.text()` on the retry response never resolves, and the test times out. The two FIX tests pass because they detach the response body from the interceptor's internal stream before consuming it - I added these as (probably bad) examples of how this might be patched in MSW itself.

## Analysis (by AI - might be inaccurate!)

The fetch interceptor's passthrough path clones the response for `response` event listeners:

```ts
// src/interceptors/fetch/index.ts, ~line 88
const responseClone = FetchResponse.clone(originalResponse)
await emitAsync(this.emitter, 'response', {
  response: responseClone,
  isMockedResponse: false,
  request: requestCloneForResponseEvent,
  requestId,
})
```

`FetchResponse.clone()` calls `response.clone()`, which internally calls `ReadableStream.tee()` on the body. The clone's body is never consumed or cancelled after `emitAsync` returns. This leaves an orphaned tee branch.

When the caller cancels the original response body (`reader.cancel()`) and makes a new request, the next response's `.text()` call deadlocks. The exact mechanism inside the interceptor that causes the second response to be affected by the first response's tee state isn't fully clear, but the deadlock is reliable and consistent.

### Workarounds

Both workarounds detach the response body from the interceptor's internal stream:

1. **Pipe through an identity `TransformStream`** (preferred — preserves streaming):
   ```ts
   const res = await fetch(url, init);
   if (!res.ok && res.body) {
     return new Response(res.body.pipeThrough(new TransformStream()), {
       status: res.status,
       statusText: res.statusText,
       headers: res.headers,
     });
   }
   ```

2. **Eagerly consume the body and re-wrap**:
   ```ts
   const res = await fetch(url, init);
   if (!res.ok) {
     const text = await res.text();
     return new Response(text, { status: res.status, statusText: res.statusText, headers: res.headers });
   }
   ```

### Possible fix

Cancel the clone's body after `emitAsync` returns, so the orphaned tee branch doesn't accumulate:

```ts
const responseClone = FetchResponse.clone(originalResponse)
await emitAsync(this.emitter, 'response', {
  response: responseClone,
  // ...
})
// Drain the clone's body to prevent tee back-pressure on the original.
await responseClone.body?.cancel()
```

## Environment

- `msw`: 2.14.6
- `@mswjs/interceptors`: 0.41.9
- `vitest`: 4.1.8
- Node.js: v22.16.0
- OS: Linux

## Maybe Related

- #768

kettanaito:
- Fixes #712
- Fixes #799


kettanaito:
## Released: v0.42.5 🎉

This has been released in v0.42.5.

- 📄 [**Release notes**](https://github.com/mswjs/interceptors/releases/tag/v0.42.5)
- 📦 [View on npm](https://www.npmjs.com/package/@mswjs/interceptors/v/0.42.5)

```
```

---

